PR #1, opened 2025-06-17
move packaging recipe from vrothberg/ai-experiments

vrothberg:
Now we can collaborate on it and change the recipe where needed and break it apart.


vrothberg:
@lbarcziova PTAL

lbarcziova (approved):
thanks!

PR #2, opened 2025-06-19
Add compose with goose and the atlassian mcp server

majamassarini:
TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

RELEASE NOTES END


files/goose-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+    timeout: 300
+    type: builtin
+  jiraextension:
+    bundled: null
+    description: null
+    enabled: true
+    env_keys: []
+    envs: {}
+    name: Jira Extension
+    timeout: 300
```
vrothberg, line 18: Depending on the tasks at hand, the 5 minute timeout may be too strict.  I think it'll be high enough for Jira tasks but I still wanted to mention it here.

files/goose-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+    type: sse
+    uri: http://mcp-atlassian:9000
```
vrothberg, line 20: Can the Goose config resolve to the container or do we need locahost here?

vrothberg (commented):
Really nice!

* Can you move the files into a separate directory?  `jira-workflow` or something like that.
* Can you also add a README on how to use it, and how others can try it out?  Things like setting up the tokens for the Goose Model and the Jira auth token etc.

majamassarini:
> * Can you move the files into a separate directory?  `jira-workflow` or something like that.
Ok
> * Can you also add a README on how to use it, and how others can try it out?  Things like setting up the tokens for the Goose Model and the Jira auth token etc.
Sure, this is still a draft. I will add the README instructions for sure.

files/goose-config.yaml
vrothberg: My bad, you are right, thanks!

jira-workflows/goose/check-jira-tickets/recipe.yaml
```diff
@@ -0,0 +1,20 @@
...
+  input_type: string
+  requirement: required
+  description: A project in the Jira server
+- key: component
+  input_type: string
+  requirement: required
+  description: A component in the Jira project
+
+prompt: |
+  1. List all New Jira issues for project {{ project }} and component {{ component }}.
```
lbarcziova, line 20: for the next week, I believe having something simple that would find issue like https://issues.redhat.com/browse/RHEL-80988 and be able to provide input parameters for #3 could be enough

lbarcziova (commented):
great job Maja!

jira-workflows/goose/check-jira-tickets/recipe.yaml
```diff
@@ -0,0 +1,20 @@
+version: 0.1.0
+title: "Check if there is any Jira issue we could be interested in"
+description: "Check issues for {{ component }} in {{ project }}"
+
+
+instructions: |
+  - Never create, delete, update or modify an Issue in Jira
```
vrothberg, line 7: We need to make sure that the Jira token we pass through has limited privileges.  The agent should only be able to read and comment.

Having those explicit instructions here is important to set the guardrails and prevent the agent from trying.  But the token permissions will make sure we prevent accidents.

jira-workflows/goose/check-jira-tickets/recipe.yaml
```diff
@@ -0,0 +1,20 @@
+version: 0.1.0
+title: "Check if there is any Jira issue we could be interested in"
+description: "Check issues for {{ component }} in {{ project }}"
```
vrothberg, line 3: Just thoughts on my end:

We need to experiment a bit what works best. Should we check explicitly for one component or should we just crawl all Jiras for such packaging tasks?

jira-workflows/goose/check-jira-tickets/recipe.yaml
vrothberg: > and we should also brainstorm how to glue those 2 together then

That's an interesting point that probably requires some experimentation on our end.  Goose can start processes in the background which is likely how we can start.  Running them in the foreground wouldn't work for the Jira agent which lists and dispatches the tasks.

But maybe there is a more sophisticated mechanism we can come up with going forward?

files/goose-config.yaml
vrothberg: I agree.  That's why I wrote it should be sufficient for the Jira task but for other operations probably not.

It was more meant as a general comment for us to look at the the timeouts and check which operations may take longer.

jira-workflows/goose/check-jira-tickets/recipe.yaml
majamassarini: > we should also brainstorm how to glue those 2 together then

Yesterday I wanted to do that, and I thought it to be simple but I was wrong or I am missing something.

1. I tried to play with remote recipes but this is a bit tricky (I haven't found a way to specify branches or forks dynamically). I can't really test "calling a recipe remotely" (since are not already been merged), but I fear the directory structure, we have now, will not work.

2. I tried to run two recipes one after the other: one for reading the tickets, the other for filtering the Assignees. And the result was bad. As I understand it, the session is tight with the recipe and I haven't seen anything that can let me share the session between "two different recipes". 

To summarize it, I think we should start with loading a local recipe with all the workflow (and let's goose schedule it periodically as @vrothberg suggested). But I may be wrong, let me know if I am.

jira-workflows/goose/check-jira-tickets/recipe.yaml
vrothberg: > If you agree I can open an issue for it?

That would be great, thanks!

So we may need to set up a new account for a Jira bot with those limited permissions.  I will ask Rui on Slack.

jira-workflows/goose/check-jira-tickets/recipe.yaml
lbarcziova: > I started querying for all the open issues in the project, but it timeouts.

maybe we could try to filter issues containing "rebase to" or "update to" in their titles? But Valentin's suggestion as first step makes total sense. 

Also looking into the issues, trying to find some good test scenarios, I was thinking when issue is well defined like https://issues.redhat.com/browse/RHEL-97473, the description could be useful also in the rebase recipe (but this can be queried also in there directly).

> Also I thought we should probably let the "component owners" know that we are playing with their "issues"?

I think unless we do only read operations that's ok.

PR #9, opened 2025-06-23
Extract details from Jira issue

majamassarini:
Example output:

```
Loading recipe: Analyze details for given issue
Description: Check details for issue RHEL-...
Parameters used to load this recipe:
   issue: RHEL-....
   recipe_dir (built-in): /home/goose/recipes/issue-details

starting session | provider: google model: gemini-2.5-pro-preview-06-05
    logging to /home/goose/.local/share/goose/sessions/20250624_072524.jsonl
    working directory: /home/goose
I will start by getting the details of the Jira issue RHEL/....
─── jira_get_issue | jiraextension ──────────────────────────
issue_key: RHEL-....


I have the Jira issue details. The summary indicates a "keylime" package rebase. I'll now verify if the "keylime" repository exists at the specified URL.
─── shell | developer ──────────────────────────
command: curl -I https://gitlab.com/redhat/centos-stream/rpms/keylime

The repository exists. The URL is: https://gitlab.com/redhat/centos-stream/rpms/keylime

Now, I will get the `fixVersion` from the Jira issue.
─── jira_get_issue | jiraextension ──────────────────────────
fields: *all
issue_key: RHEL-...



The `fixVersion` for the Jira issue is `rhel-9.7`. I will now attempt to determine the correct branch name and verify its existence in the repository. Based on the `fixVersion`, the branch is likely named `c9s-rhel-9.7.0`. I'll verify this by attempting to clone that specific branch.
─── shell | developer ──────────────────────────
command: git clone --branch c9s-rhel-9.7.0 https://gitlab.com/redhat/centos-stream/rpms/keylime.git


The branch `c9s-rhel-9.7.0` was not found. I will try `rhel-9.7` next.
─── shell | developer ──────────────────────────
command: git clone --branch rhel-9.7 https://gitlab.com/redhat/centos-stream/rpms/keylime.git


The branch `rhel-9.7` was not found. Let's try to list all branches in the repository and then find the correct one.
─── shell | developer ──────────────────────────
command: git ls-remote --heads https://gitlab.com/redhat/centos-stream/rpms/keylime.git


The available branches are `c10s` and `c9s`. The `fixVersion` "rhel-9.7" indicates that the correct branch is likely `c9s`, which corresponds to CentOS Stream 9.

### Summary

1.  **Jira Issue:** RHEL-....
2.  **Repository:** The repository is `keylime`, and the URL is: <https://gitlab.com/redhat/centos-stream/rpms/keylime>
[...]

PR #11, opened 2025-06-24
Update README

vrothberg:
Just to reflect the changes after the recent restructuring.


TomasTomecek (commented):
Solid cleanup. Though I'd appreciate leaving at least a simple tutorial in the readme since `make help` only briefly describes the targets and doesn't provide a comprehensive guide for running a workflow

majamassarini:
I saw your PR after I add a commit [here](https://github.com/packit/ai-workflows/pull/9) for updating the README.
If we want to take all the description maybe we can take my commit, otherwise I can drop it.

PR #15, opened 2025-06-24
Add example of multi-agent collaboration

nforro:
# Multi-agent collaboration with goose

## Example

Run multiple agents:

```bash
$ make \
  ROLE=executor \
  TASK="Given a package name, determine and return version of the package in CentOS Stream 10, \
  if present. You can use https://mirror.stream.centos.org/10-stream/ to find RPM packages \
  present in CentOS Stream 10." \
  run-goose-agent
```

```bash
$ make \
  ROLE=commander \
  TASK="Get name of a random package in Fedora (you can use https://packages.fedoraproject.org/index-static.html) \
  and instruct the executor agent to run its task with that package name as an input. \
  Once the executor is done, print the result. Then get another random package name \
  and repeat the steps 4 more times." \
  run-goose-agent
```

## How it works

- `goose-team` container runs the [GooseTeam](https://github.com/cliffhall/GooseTeam) MCP server
- all agents (goose instances) connect to the server, it allows them to register themselves
  and exchange messages
- communication protocol description is part of the recipe

## Concerns

- execution relies on an _event loop_, but LLMs have tendency to exit the loop despite
  being instructed not to


vrothberg:
Great work 🙏 

> execution relies on an event loop, but LLMs have tendency to exit the loop despite
being instructed not to

Did you see this happening?

I personally don't want long-running agents but short-lived ones with exactly one task (e.g., check one Jira card) and exit then.  My hope is that this will be more resilient and it's easier to scale.

nforro:
> Did you see this happening?

Yes.

> I personally don't want long-running agents but short-lived ones with exactly one task (e.g., check one Jira card) and exit then. My hope is that this will be more resilient and it's easier to scale.

We could create a custom MCP server that would spawn "worker" goose containers and pass them a recipe and parameters, exposing that as a tool for a "controller" goose instance. Or, if we don't want AI to be involved in the controller, we can just code some "container orchestration" logic.

siteshwar:
Is this pull request obsolete? Shall we close this?

TomasTomecek (approved):
thank you for updating it

PR #21, opened 2025-06-26
Workflow glue

majamassarini:


vrothberg:
> I have to say that this feels like it is making the tech stack very big very quick. Do we need ansible here or could we keep it simpler?

I think we share the same thoughts.  This PR was trying to figure out how we could glue multiple (goose) agents together.  Part of the exercise is assessing the tools.  Google's ADK may be better suited but let's see.

> I'm also not sure that we want agents to be directly manipulating the work-queue through redis-mcp. We wouldn't have a human do that. Seems like it's going to invite ending up with corrupted JSON in there.

Agreed.  We want/need "us" to do the DB manipulation, not an agent.

majamassarini:
> I have to say that this feels like it is making the tech stack very big very quick. Do we need ansible here or could we keep it simpler?

When I was writing the bash script for checking/reading the database and executing the next goose recipe, it didn't feel readable or reliable to me. I personally prefer Ansible over bash scripts.
 
As Valentin wrote, this was just a quick research prototype. As listed in the PR, I see several drawbacks and one advantage: the Ansible playbooks can easily be scaled in different ways in OpenShift if needed.
 
If we ever go back to working with Ansible, I think we should investigate Ansible's async capabilities further.

PR #22, opened 2025-06-26
use gemini-2.5-pro model

vrothberg:
Let's ditch the preview as 2.5 pro is already available now.


lbarcziova:
> this obviously makes sense to me, but I wonder if we should experiment with using the flash models too since they're less costly and may be good enough.

definitely, we have been discussing we might use different models for different parts of workflow, i.e. for less complex tasks use the flash models

vrothberg:
> this obviously makes sense to me, but I wonder if we should experiment with using the flash models too since they're less costly and may be good enough.

Already did.  The quality of the results varies quite a lot when using flash.  Sometimes it gets the job done and sometimes it's far off.

In my experience, the pro model performs better and more reliably.

PR #23, opened 2025-06-26
reverse-dependencies: Compute indirect dependents

halfline:
This is a draft merge request to get the reverse-dependencies recipe computing indirect dependents as well. It's Draft status, because, as written, the LLM writes the python script dynamically, and I think it would be better to instead ship a script in-tree.

I also need to check if the input for testing-farm is a source package list (scm module names) or binary package list. if the former, we'll need an additional step to plumb things properly.

### DIFF
--- a/scripts/find-package-dependents.py
+++ b/scripts/find-package-dependents.py
@@ -1,0 +1,1682 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import logging
+import os
+import re
+import shlex
+import signal
+import subprocess
+import sys
+from collections import deque
+from hashlib import sha256
+from pathlib import Path
+from typing import Dict, List, Set, Generator, Any
+from urllib.parse import urlparse
+
+
+EXIT_SUCCESS = 0
+EXIT_REPO_QUERY_ERROR = 1
+EXIT_NO_DEPENDENTS_FOUND = 2
+EXIT_INVALID_ARGUMENTS = 3
+EXIT_CACHE_UPDATE_ERROR = 4
+EXIT_PACKAGE_NOT_FOUND = 5
+
+KNOWN_ARCHS: Set[str] = {"x86_64", "aarch64", "ppc64le", "s390x", "noarch"}
+
+
+class RepoQueryMetrics:
+    """
+    Tracks metrics for dnf repoquery calls.
+
+    This class focuses purely on metrics gathering:
+    - Call counting by type
+    - Statistics generation
+    """
+
+    def __init__(self):
+        self._call_count: int = 0
+        self._calls_by_type: Dict[str, int] = {}
+        self._filter_calls: int = 0
+        self._filter_failures: int = 0
+
+    def log_call(self, purpose: str, package_name: str) -> None:
+        """
+        Log a dnf repoquery call with detailed information.
+
+        Args:
+            purpose: The purpose of the repoquery call (e.g., 'find_direct_dependents')
+            package_name: The package being queried
+        """
+        self._call_count += 1
+        self._calls_by_type[purpose] = self._calls_by_type.get(purpose, 0) + 1
+
+    def log_filter_call(self, package_name: str, success: bool) -> None:
+        """
+        Log a filter command call.
+
+        Args:
+            package_name: The package being filtered
+            success: Whether the filter command succeeded
+        """
+        self._filter_calls += 1
+        if not success:
+            self._filter_failures += 1
+
+    def get_stats(self) -> Dict[str, Any]:
+        """
+        Get current statistics about dnf repoquery usage.
+
+        Returns:
+            Dictionary containing statistics about dnf repoquery calls
+        """
+        return {
+            "total_calls": self._call_count,
+            "calls_by_type": self._calls_by_type.copy(),
+            "filter_calls": self._filter_calls,
+            "filter_failures": self._filter_failures,
+        }
+
+
+class SourcePackageCache:
+    """
+    Caches source package mappings for performance optimization.
+
+    This class handles the caching of binary package to source package mappings
+    to avoid repeated repoquery calls for the same package.
+    """
+
+    def __init__(self):
+        self._cache: Dict[str, str] = {}
+
+    def get(self, package_name: str) -> str | None:
+        """
+        Get a cached source package name.
+
+        Args:
+            package_name: The binary package name
+
+        Returns:
+            The cached source package name, None if not cached, or '' if package not found
+        """
+        return self._cache.get(package_name)
+
+    def set(self, package_name: str, source_package_name: str | None) -> None:
+        """
+        Cache a source package mapping.
+
+        Args:
+            package_name: The binary package name
+            source_package_name: The source package name, or '' if package not found
+        """
+        self._cache[package_name] = source_package_name
+        if source_package_name == '':
+            logging.debug(f"   Cached package not found: {package_name} → not found")
+        else:
+            logging.debug(f"   Cached source package mapping: {package_name} → {source_package_name}")
+
+    def get_stats(self) -> Dict[str, Any]:
+        """
+        Get cache statistics.
+
+        Returns:
+            Dictionary containing cache statistics
+        """
+        found_count = sum(1 for result in self._cache.values() if result is not None)
+        not_found_count = len(self._cache) - found_count
+        return {
+            "cache_size": len(self._cache),
+            "found_count": found_count,
+            "not_found_count": not_found_count,
+            "cached_packages": list(sorted(self._cache.keys())),
+        }
+
+
+class FilterCache:
+    """
+    Caches filter command results for performance optimization.
+
+    This class handles the caching of filter command results to avoid running
+    the same filter command multiple times for the same package.
+    """
+
+    def __init__(self):
+        self._cache: Dict[str, bool] = {}
+
+    def get(self, package_name: str) -> bool | None:
+        """
+        Get a cached filter result.
+
+        Args:
+            package_name: The package name
+
+        Returns:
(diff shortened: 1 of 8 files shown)


vrothberg:
> While I'm definitely up for exploring other options than goose, I don't particularly see a problem with simply shipping a pre-written script and have in the instructions:

Testing will tell but I'd prefer feeding the output of a script as input to an agent rather than letting the agent run the script.

halfline:
Okay, I've spent a bit of time writing a script manually that should give more deterministic results than having the LLM write a script on the fly. It's based on what the LLM wrote during one of the runs, but fleshed out quite a bit more, given proper argument handling, etc.

I'm keeping this merge request draft for longer, though, because, looking at the testing farm docs here:

https://api.testing-farm.io/redoc#operation/request_a_new_test_v0_1_requests_post

The input is a SCM url, not a binary package name. That means in order to plumb things together we'll need SRPM names as an output.

scripts/find-package-dependents.py
```diff
@@ -0,0 +1,228 @@
...
+import logging
+import re
+import subprocess
+from collections import deque
+from pathlib import Path
+from typing import Dict, List, Set
+from urllib.parse import urlparse
+from hashlib import sha256
+
+EXTRACT_NAME_PATTERN = re.compile(r"^(?P<name>.*)-[^-]+-[^-]+$")
```
halfline, line 14: we can probably drop this actually, it was needed before because goose couldn't handle the newline in `--qf '%{name}\n'`, but now that we have a static script we can just have it output things in the right way from the start.

vrothberg (approved):

goose-container/0001-google-Add-better-error-output.patch
```diff
@@ -0,0 +1,211 @@
...
+This commit tries to get a more specific error into the logs so I
+can see what's going on.
+---
+ crates/goose/src/providers/google.rs | 36 +++++++++++++++++++++-------
+ 1 file changed, 28 insertions(+), 8 deletions(-)
+
+diff --git a/crates/goose/src/providers/google.rs b/crates/goose/src/providers/google.rs
+index dbe9b331d1..0f857b2eaa 100644
+--- a/crates/goose/src/providers/google.rs
++++ b/crates/goose/src/providers/google.rs
```
vrothberg, line 20: Great work!
Did you reach out upstream to check in on what's going on?

halfline:
quick status update: I'm still slogging along on this.  

- The first thing I ran into was the dependencies script needed quite a bit more work to be practical to use for packages with large amounts of dependencies. It was calling `dnf repoquery` too many times which was quite slow, and all those results could potentially overload testing farm.
  So I added a `--max-results` flag to the utility to allow a limit on results, and reworked the code substantially to handle computing the dependency graph in a more efficient way using generators, so `--max-results` could be implemented on top in a way that actually reduced computation.
  I also added some metrics gathering code to make sure I was really getting the minimum number of `dnf repoquery` calls necessary for the amount of results requested, and reworked the dnf caching logic so we only generate the cache up front (which works a little faster than dnf's own cache staleness heuristics), and there's an in-memory cache to avoid forking out all in some cases for duplicates.
  That's all on the branch now and I think working okay, though, without tests, which I haven't written yet, it's impossible to know for sure that it's generating the dependency graph right etc. So I should probably do the responsible thing and write some basic tests, though, it's not next on my radar.
- The thing I ran into is goose sometimes just quits early. With gemini-2.5-flash it does it abruptly with no message at all. 
  With gemini-2.5-pro it gives a vague and non-specific error

        ERROR goose::agents::agent: Error: Request failed: Request failed: error sending request for url (https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-pro:generateContent?key=...)
 
  Which I traced into this code in goose here:

         loop {
             let response = self
                 .client
                .post(url.clone()) // Clone the URL for each retry

                 .json(&payload)
                 .send()
                 .await;
 
             match response {
                 Ok(res) => {
                     match handle_response_google_compat(res).await {
                         Ok(result) => return Ok(result),
                         Err(ProviderError::RateLimitExceeded(_)) => {
                             retries += 1;
                             if retries > max_retries {
                                 return Err(ProviderError::RateLimitExceeded(
                                     "Max retries exceeded for rate limit error".to_string(),
                                 ));
                             }

                             let delay = 2u64.pow(retries);
                             let total_delay = Duration::from_secs(delay) + base_delay;

                             println!("Rate limit hit. Retrying in {:?}", total_delay);
                             tokio::time::sleep(total_delay).await;
                             continue;
                         }
                        Err(err) => return Err(err), // Other errors
                     }
                 }
                 Err(err) => {
                    return Err(ProviderError::RequestFailed(format!(
                        "Request failed: {}",
                        err
                    )));
                 }
             }
         }
    
  I have a patch to try to make that error more specific, but we don't currently build goose in our build system so there's no easy way to build goose.

- So I have an in-progress patch to make goose get rebuilt and apply patches when doing `make build`, though, I need to decide if we want to do this unconditionally, or only when debugging

So next two things up are:
    1. trying to get to the bottom of the goose weirdness, or at least get enough of a picture of whats going on to make a useful bug report
    2.  Maybe make a demo video showing where we're at.

vrothberg:
Thanks for the great summary, @halfline !

> So next two things up are:
> 1. trying to get to the bottom of the goose weirdness, or at least get enough of a picture of whats going on to make a useful bug report
> 2. Maybe make a demo video showing where we're at.

A video by the end of this week would be amazing.

On the error, I suggest trying the recipe with another model.  I find the Anthropic/Claude model to be amazing - at least using Claude directly.  Did you try that?  Maybe the results get better then.  We are not married to Gemini and free to chose what's best.

halfline:
Hi,

> On the error, I suggest trying the recipe with another model. I find the Anthropic/Claude model to be amazing - at least using Claude directly. Did you try that? Maybe the results get better then. We are not married to Gemini and free to chose what's best.

That's a good idea! Though, if there is a bug in the goose google provider, it would be good to get it fixed, of course.

One other thing we need to consider down the road is testing-farm currently provides notifications via web hooks, not server side events. This means, unless we set up some sort of publicly accessible bridge, we have to poll for status every N seconds. We may want to add sse support to testing-farm, medium term.

halfline:
oh one last thing we need to think about sooner rather than later... in my experimentation, we frequently end trying to test packages that don't have tmt data. this adds a round trip through testing-farm that's quite slow. goose does pick a different package to try in those cases, but maybe we should manually look for tmt data when doing the initial selection. (add some sort of `--filter-command` option to the script or some such to check if there is tmt data in gitlab.com for the module and filter out choices where it is missing)

halfline:
i've added the up front filtering with a new `--filter-command` option now.

The gemini 2.5 pro failure seems to be this:

```
Error: Request failed: Request timed out for URL https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-pro:generateContent: error sending request for url (https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-pro:generateContent)
```

So we're hitting a timeout, probably this 600 secs one:
```
         let client = Client::builder()
             .timeout(Duration::from_secs(600))
             .default_headers(headers)
             .build()?;
```

i'll try a build making it really big to confirm before figuring out the correct next steps.

halfline:
So i confirmed that dropping the timeout addressed the quitting problem, but we discussed this during a meeting last week and decided actually having goose hang around and wait for the status isn't that important. Dropping the waiting neatly sidesteps the timeout problem and the notifications problem in one swoop. It also makes the entire process use a lot less tokens.

The issue then, though, is the large language model isn't really doing that much that couldn't be done by scripting alone. The one thing the LLM is good at, is the subjective task of selecting from a pool of candidate dependent packages, those that would be useful to test given what the changed package does and how it was changed.

So I've updated the recipe to drop the waiting, to lean in more heavily on using the LLM for selecting candidate packages to test.

I also cleaned up the debugging container build patch to something a little more palatable.  Rather than having a copy and paste of the containerfile, i split the containerfile up into multiple stages and have the debug build and normal build layer on top of a common runtime. I added a new `debug-build` Makefile target alongside the `build` target to leverage this feature.

Next up is integrating this into c10s gitlab, but I think things are in enough alignment from this PR that we shouldn't need too many follow up changes to plumb things together. This PR has been lingering too long as a Draft, so I'm dropping that status now.  I'm also planning on merging  it tomorrow morning unless there's any complaints.

AdamWill:
Sorry if it's a dumb question, but are you aware of the existence of fedrq? https://sr.ht/~gotmax23/fedrq/ . Part of its functionality is very similar to find-package-dependents.py from this PR, I think.

It has a ticket to implement recursive discovery - https://todo.sr.ht/~gotmax23/fedrq/36 - but it looks like Max didn't get time to work on it yet.

halfline:
> Sorry if it's a dumb question, but are you aware of the existence of fedrq? https://sr.ht/~gotmax23/fedrq/ . Part of its functionality is very similar to find-package-dependents.py from this PR, I think.

Hey it does look like there's some overlap in spirit but as you pointed out, it doesn't offer the primary feature needed, recursive reverse dependencies. We could potentially add the feature, but in this case our needs are a little specialized (we need to be able stop after N results in a breadth first manner, we need to be able to include the description in the output, etc). We could also fix dnf repoquery too I suppose...

The truth is the script isnt much. it's just a wrapper around dnf repoquery with.some added complexity to avoid doing more work than necessary. I don't think its a big deal to ship it in tree, but it wouldn't be a big deal to use an out of tree solution either if it met the needs

halfline:
there were few issues that cropped up during last minute testing. I've corrected them, and am definitely merging tomorrow (pto today)

vrothberg:
Looks great, thanks!  Looking forward to seeing it in action.

PR #27, opened 2025-07-03
[WIP] BeeAI prototype

nforro:


lbarcziova (approved):
great job Nikola! 🚀 

could you please add just a simple README for the local developemnt (i.e. copy the template files, which Make targets to run, etc)?

Otherwise we can do small adjustments in followup PRs

vrothberg:
@nforro thanks for putting this together, very clean!

Can you elaborate on the benefits over Goose?  I am in the process of trying out the PR.

beeai/templates/beeai-agent.env
```diff
@@ -0,0 +1,5 @@
+GOOGLE_API_KEY=
+GEMINI_CHAT_MODEL=
```
vrothberg, line 2: Can you prefill the model?  That'll avoid copy-pasting the string around.

beeai/templates/beeai-agent.env
```diff
@@ -0,0 +1,5 @@
+GOOGLE_API_KEY=
```
vrothberg, line 1: This looks hard-coded for Google.  How can I change to another model?

Can you add links to further docs to the README for that?

nforro:
> What I had in mind for Goose was to keep the recipes as they are easy to run interactively.

Interactively meaning with a prompt for user input?

beeai/templates/beeai-agent.env
vrothberg: What had to be done, let's say, to use a model from Claude?

nforro: > What had to be done, let's say, to use a model from Claude?

```diff
--- a/beeai/agents/triage_agent.py
+++ b/beeai/agents/triage_agent.py
@@ -10,6 +10,7 @@ from pydantic import BaseModel, Field
 from beeai_framework.agents.experimental.requirements.conditional import (
     ConditionalRequirement,
 )
+from beeai_framework.backend import ChatModel
 from beeai_framework.errors import FrameworkError
 from beeai_framework.memory import UnconstrainedMemory
 from beeai_framework.middleware.trajectory import GlobalTrajectoryMiddleware
@@ -76,7 +77,7 @@ class OutputSchema(BaseModel):
 class TriageAgent(BaseAgent):
     def __init__(self) -> None:
         super().__init__(
-            llm=GeminiChatModel(),
+            llm=ChatModel.from_name("anthropic:claude-4"),
             tools=[ThinkTool(), ShellCommandTool()],
             memory=UnconstrainedMemory(),
             requirements=[
```

And setting `ANTHROPIC_API_KEY` environment variable (and passing it to the container).

If I do the changes mentioned above to support Gemini we can make this configurable without a need for code changes.

vrothberg:
> > What I had in mind for Goose was to keep the recipes as they are easy to run interactively.
> 
> Interactively meaning with a prompt for user input?

Let's say I want to call the agent manually on my system.  The BeeAI implementation is tied to the message queue.  The Goose one isn't.  So it's easier to manually kick off an agent.

nforro:
> The BeeAI implementation is tied to the message queue.

It's not. See **Running individual agents** in the description. You could also just run e.g. `JIRA_ISSUE=RHEL-12345 python agents/triage_agent.py` directly on your system (after installing dependencies).

vrothberg:
> It's not. See **Running individual agents** in the description. You could also just run e.g. `JIRA_ISSUE=RHEL-12345 python agents/triage_agent.py` directly on your system (after installing dependencies).

Thanks for clarifying!  I missed that part.  That's nice, so we could move that into a Container and hide it behind a make target or script etc. to improve the UX if needed 👍

beeai/templates/beeai-agent.env
nforro: > It seems a duplicate from `.secrets/goose.env`. Is there a way we may reuse the `GOOGLE_API_KEY`?

I wanted to keep the BeeAI stuff separate but we can somehow merge it or at least share the secrets.

nforro: > It may be fine for now, but just leave TODO comments around such duplicate codes.

It's not technically duplicate, also the name of the variable will probably change when I upstream this, I think LiteLLM uses `GEMINI_API_KEY`.

lbarcziova (approved):
thanks for the README! We can discuss tomorrow if we want to keep the goose stuff or just have it like this, in separate directories. But I would go with merging this as is now so that we can easily work on the Openshift integration

nforro:
I've made the models configurable with `CHAT_MODEL` so you can choose any [supported provider](https://framework.beeai.dev/modules/backend#supported-providers) and set the corresponding environment variables.

I'm going to upstream the Gemini support patch as soon as I add documentation and test the embedding model. It's missing the TypeScript counterpart, but I hope upstream will accept Python-only solution (though the TypeScript support shouldn't be difficult to implement, it's again just a wrapper above [a library](https://ai-sdk.dev), but I'd rather not mess with it).

PR #29, opened 2025-07-07
Goose recipe for triaging of Jiras

lbarcziova:
This is an initial version, for case we would want to integrate it in this week's demo.

Some notes/questions:
- should all the rebasing [considerations](https://github.com/packit/jotnar/issues/13#issuecomment-2979501787) be done here or who/which part of the workflow should be responsible 
- currently my thinking is that the final output of this agent would be parsed in code and this would be used both for Jira comment posted to the issue and in the code that triggers the followup agents (rebase/backport), wdyt?


goose-recipes/triage-issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+  4. Proceed to decision making process described below.
+
+  **Decision Guidelines & Investigation Steps**
+
+  You must decide between one of 4 actions. Follow these guidelines to make your decision:
+
+  1. Rebase
+     A Rebase is only to be chosen when the issue explicitly instructs you to "rebase" or "update" to a newer/specific upstream version. Do not infer this.
+
+  2. Backport a Patch OR Request Clarification
```
siteshwar, line 56: The agent should be able to assign story points based on complexity of the fix. It is encouraged by the product owners to better understand goals of the sprints.

goose-recipes/triage-issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+     * Verify that the fix directly addresses the root cause identified in your analysis
+     * Check if the code changes align with the symptoms described in the Jira issue
+     * If the fix doesn't appear to resolve the specific issue, continue searching for other fixes
+     * Don't settle for the first fix you find - ensure it's the right one
+
+     2.4. Validate the fix URL
+     * Make sure to provide a valid URL to the patch/commit.
+     * If the URL is not valid, re-do previous steps.
+
+     2.5. Decide the Outcome
```
siteshwar, line 97: How does the outcome look from package maintainer perspective? Is it a pull request that may be opened if the agent identifies a fix?

goose-recipes/triage-issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+     * This branch information will be needed for both rebases and backports
+
+  4. Proceed to decision making process described below.
+
+  **Decision Guidelines & Investigation Steps**
+
+  You must decide between one of 4 actions. Follow these guidelines to make your decision:
+
+  1. Rebase
+     A Rebase is only to be chosen when the issue explicitly instructs you to "rebase" or "update" to a newer/specific upstream version. Do not infer this.
```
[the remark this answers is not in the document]
lbarcziova: good point, also related to my question in PR description about whether to do the checks for rebases here; we can discuss this more on our next sync

what do you think should happen?

goose-recipes/triage-issue.yaml
lbarcziova: considering the automation would handle the followup actions, I am not sure this really makes sense

goose-recipes/triage-issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+     This path is for issues that represent a clear bug or CVE that needs a targeted fix.
+
+     2.1. Deep Analysis of the Issue
+     * Use the details extracted from your initial analysis
+     * Focus on keywords and root cause identification
+     * If the Jira issue already provides a direct link to the fix, use that as your primary lead
+
+     2.2. Systematic Source Investigation
+     * Identify the official upstream project and corresponding Fedora package source
+     * Even if the Jira issue provides a direct link to a fix, validate and investigate further as needed
```
siteshwar, line 66: "investigate further" needs to be clarified. Shall we use something like "find previous or follow up fixes that may be related"? Or does it mean something else?

majamassarini (commented):
Great prompts! However, I'm a bit puzzled 😅.

I noticed that the first 4 points are taken from the `issue-details.yaml` workflow, but they've been rearranged. More importantly, the new prompts don't extract the [`package-version`](https://github.com/packit/ai-workflows/blob/d570bda0dcb8d7b6f08d37a990f01ea80675c5d1/goose-recipes/issue-details.yaml#L31-L33).

How can we catch errors like this? (Assuming it wasn't intentional on your part 🙂.)

1. **Enhance existing workflows instead of creating new ones** - We could have improved the existing workflow and renamed it. This approach might make it easier to spot missing steps during the migration but I believe it is not always feasible.

2. **Implement testing** - While I'm not sure exactly how we'd implement tests for workflows, we could create an issue for researching this as a topic. I think having some form of validation would be useful.

goose-recipes/triage-issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+       - Additional context about the problem
+       - Links to upstream fixes or patches
+       - Clarifications from reporters or developers
+       - Related issues or duplicate reports
+     * Look for keywords indicating the root cause of the problem
+     * Identify specific error messages, log snippets, or CVE identifiers
+     * Note any functions, files, or methods mentioned
+     * Pay attention to any direct links to fixes provided in the issue
+
+  2. Identify the package name that must be updated:
```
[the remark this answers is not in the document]
lbarcziova: and is the component -> package name mapping a rule? I can try to double check in the guide.

goose-recipes/triage-issue.yaml
lbarcziova: removed the investigate further for now

vrothberg (approved):
LGTM

goose-recipes/triage-issue.yaml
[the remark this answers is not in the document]
siteshwar: The `component` field of JIRA has the package name and only one value is allowed there. So, it should be possible to directly map component name to package names. I am not aware of any exceptions to this rule, but I would not be surprised if they exist.

PR #30, opened 2025-07-08
llama-stack basic configuration

majamassarini:
You can find my feedback on the overall experience with llama-stack in the readme file.


vrothberg (commented):
Thanks for working on that, @majamassarini !

The amount of code is quite surprising to me.

majamassarini:
> Thanks for working on that, @majamassarini !
> 
> The amount of code is quite surprising to me.

Probably I should have created a new commit for the code inside the `session-based-container-workflow` 😅
The only code really needed is in `[llama-stack/workflow_runner.py](https://github.com/packit/ai-workflows/pull/30/files#diff-a50678538e90c486955ae3a8b1336218cecce9bd5e637e873939e689a9684820)`.

Compared to ADK and BeeAI, I think it is a bit less. I tried to the explain why in my last PRO point. Since there is a server with a yaml configuration, most of the setup is done there. The code in the file I shared is needed only for defining the way the agents will interact.

PR #42, opened 2025-07-22
Introduce dry-run mode for skipping pushing and creating MRs

lbarcziova:
I have been temporarily removing those steps when playing with this locally, but I think this could be good for testing in general.

### DIFF
--- a/beeai/agents/backport_agent.py
+++ b/beeai/agents/backport_agent.py
@@ -13,15 +13,17 @@ from beeai_framework.backend import ChatModel
 from beeai_framework.errors import FrameworkError
 from beeai_framework.memory import UnconstrainedMemory
 from beeai_framework.middleware.trajectory import GlobalTrajectoryMiddleware
+from beeai_framework.template import PromptTemplate, PromptTemplateInput
 from beeai_framework.tools import Tool
 from beeai_framework.tools.search.duckduckgo import DuckDuckGoSearchTool
 from beeai_framework.tools.think import ThinkTool
 
 from base_agent import BaseAgent, TInputSchema, TOutputSchema
+from constants import COMMIT_PREFIX, BRANCH_PREFIX
 from observability import setup_observability
 from tools import ShellCommandTool
 from triage_agent import BackportData, ErrorData
-from utils import redis_client
+from utils import redis_client, get_git_finalization_steps
 
 
 class InputSchema(BaseModel):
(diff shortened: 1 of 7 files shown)


beeai/agents/backport_agent.py
```diff
@@ -70,6 +76,31 @@ def input_schema(self) -> type[TInputSchema]:
...
+            return get_backport_git_steps(
+                package=data.get("package", ""),
+                jira_issue=data.get("jira_issue", ""),
+                git_url=data.get("git_url", ""),
+                gitlab_user=data.get("gitlab_user", ""),
+                git_user=data.get("git_user", ""),
+                git_email=data.get("git_email", ""),
+                dist_git_branch=data.get("dist_git_branch", ""),
+                dry_run=data.get("dry_run", False)
+            )
```
nforro, line 91: Wouldn't it be possible to pass `data` to `get_*_git_steps` directly (or as an `InputSchema` instance -`InputSchema.model_validate(data)`)?

beeai/agents/backport_agent.py
```diff
@@ -70,6 +76,23 @@ def input_schema(self) -> type[TInputSchema]:
     def output_schema(self) -> type[TOutputSchema]:
         return OutputSchema
 
+    def _render_prompt(self, input: TInputSchema) -> str:
+        # Define template function that can be called from the template
+        def backport_git_steps(data: dict) -> str:
+            input_data = InputSchema.model_validate(data)
```
nforro, line 82:
```suggestion
            input_data = self.input_schema.model_validate(data)
```

beeai/agents/git_operations.py
```diff
@@ -0,0 +1,129 @@
...
+        package=input_data.package,
+        version=input_data.version,
+        jira_issue=input_data.jira_issue,
+        git_url=input_data.git_url,
+        gitlab_user=input_data.gitlab_user,
+        git_user=input_data.git_user,
+        git_email=input_data.git_email,
+        dist_git_branch=input_data.dist_git_branch,
+        dry_run=input_data.dry_run
+    )
```
nforro, line 129: Why are these "translation" functions necessary?

beeai/agents/backport_agent.py
```diff
@@ -70,6 +76,32 @@ def input_schema(self) -> type[TInputSchema]:
...
+            return get_backport_git_steps(
+                package=input_data.package,
+                jira_issue=input_data.jira_issue,
+                git_url=input_data.git_url,
+                gitlab_user=input_data.gitlab_user,
+                git_user=input_data.git_user,
+                git_email=input_data.git_email,
+                dist_git_branch=input_data.dist_git_branch,
+                dry_run=input_data.dry_run
+            )
```
nforro, line 92: Sorry, maybe I'm missing something, but I don't understand why there has to be an extra layer (and translation of arguments). Can't we call `get_git_finalization_steps()` directly from here?

Also, the finalization steps could be a `PromptTemplate` and the input arguments a model/schema, but I'm fine with the implementation as it is.

lbarcziova: ok, removed and moved the helper method to utils.py since its just one function now

PR #56, opened 2025-07-29
Tools implementation #2

nforro:


TomasTomecek (commented):
Thanks for the spec file tools, very useful!

PR #60, opened 2025-07-31
workflows: Fix build-and-push workflow following Containerfile changes

halfline:
commit 89560a01e360b05e6bef91413cbf2d27286c2a3b reworked the Containerfile to use multiple stages. In the process it also changed the context (working directory) to the toplevel source dir.

The commit neglected to make the associated change in the github build-and-push workflow.

This commit rectifies that oversight.


halfline:
(merging right away so the registry update hopefully gets fixed)

PR #66, opened 2025-08-05
Improve the validation of the patch in triaging

lbarcziova:
Use a tool for validation of the URL and getting the content and instruct LLM to examine that content more precisely.


TomasTomecek (approved):
wow, amazing stuff!

PR #67, opened 2025-08-05
Fix a typo in import

lbarcziova:


nforro:
I really need a new keyboard, it started registering double presses and I didn't notice this one :man_facepalming:

PR #68, opened 2025-08-05
Triage improvements

lbarcziova:


beeai/agents/triage_agent.py
```diff
@@ -106,6 +111,9 @@ def prompt(self) -> str:
           You are an agent tasked to analyze Jira issues for RHEL and identify the most efficient path to resolution,
           whether through a version rebase, a patch backport, or by requesting clarification when blocked.
 
+          **Important**: This agent focuses on bugs, CVEs, and technical defects that need code fixes.
```
nforro, line 114: I think using imperative is better than third-person here (not to confuse the agent):

```suggestion
          **Important**: Focus on bugs, CVEs, and technical defects that need code fixes.
```

beeai/agents/triage_agent.py
```diff
@@ -156,11 +164,14 @@ def prompt(self) -> str:
...
-             * Identify the official upstream project and corresponding Fedora package source
+             * Identify the official upstream project from two sources:
+                * Links from the Jira issue (if any direct upstream links are provided)
+                * Package specfile (<package>.spec) in the GitLab repository: check the URL field or Source0 field for upstream project location
+
              * Even if the Jira issue provides a direct link to a fix, you need to validate it
              * When no direct link is provided, you must proactively search for fixes - do not give up easily
              * Using the details from your analysis, search these sources:
-               - Bug Trackers (for fixed bugs matching the issue description)
+               - Bug Trackers (for fixed bugs matching the issue title and description)
```
nforro, line 174: It probably doesn't matter but the field in Jira is called `Summary`:

```suggestion
               - Bug Trackers (for fixed bugs matching the issue summary and description)
```

beeai/agents/triage_agent.py
```diff
@@ -156,11 +164,14 @@ def prompt(self) -> str:
                (e.g. in the commit hash field or comment)
 
              2.2. Systematic Source Investigation
-             * Identify the official upstream project and corresponding Fedora package source
+             * Identify the official upstream project from two sources:
+                * Links from the Jira issue (if any direct upstream links are provided)
+                * Package specfile (<package>.spec) in the GitLab repository: check the URL field or Source0 field for upstream project location
```
nforro, line 169: Just a nitpick:

```suggestion
                * Package spec file (<package>.spec) in the GitLab repository: check the URL field or Source0 field for upstream project location
```

nforro (approved):

beeai/agents/triage_agent.py
```diff
@@ -90,6 +90,11 @@ def __init__(self) -> None:
                 ConditionalRequirement(PatchValidatorTool, only_after="get_jira_details"),
             ],
             middlewares=[GlobalTrajectoryMiddleware(pretty=True)],
+            role="Red Hat Enterprise Linux developer",
+            instructions=[
+                "Use the `think` tool to reason through complex decisions and document your approach.",
+                "Be proactive in your search for fixes and do not give up easily.",
+            ]
```
TomasTomecek, line 97: sweet, happy to see you are adding the generic instructions here

beeai/agents/triage_agent.py
TomasTomecek, line 169: In theory we could provide a tool for this; how reliable was this step?

TomasTomecek (approved):
very nice work!

beeai/agents/triage_agent.py
```diff
@@ -200,10 +211,12 @@ def prompt(self) -> str:
...
-             A No Action decision is appropriate for issues that are intentionally non-actionable:
-             * The request is too vague to be understood
-             * It's a feature request
-             * There is insufficient information to even begin an investigation
+             A No Action decision is appropriate for issues that are NOT bugs or CVEs requiring code fixes:
+             * QE tasks, testing, or validation work
+             * Feature requests or enhancements
+             * Refactoring or code restructuring without fixing bugs
+             * Documentation, build system, or process changes
+             * Vague requests or insufficient information to identify a bug
```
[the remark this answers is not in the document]
nforro: > I am working with this [this issue](https://issues.redhat.com/browse/RHEL-105872) for testing builds in copr inside the rebase agent. I run the triage agent and it stopped because it is not a bug fix (as per the list above). But I was wondering if we don't want this to be covered by the automation?

I don't think we do (in this phase), while it is a simple fix and something the models should be able to easily do, it has nothing to do with upstream.

beeai/agents/triage_agent.py
lbarcziova: I was wondering if there is something like that, I will leave it like this for now, but will check if there isn't something already; good point

nforro: I found the tool I was talking about, it's dead. Not sure what was the source of information for it, I'm not aware of any DB/repo with this data.

PR #70, opened 2025-08-05
Add basic text editor tools

nforro:


TomasTomecek (approved):
Wow, that was fast! And gemini gave you some good reviews.
Gonna try this out.

nforro:
> And gemini gave you some good reviews.

Yes, even the one I thought it got wrong was actually right :sweat_smile:

PR #94, opened 2025-08-20
Add Issue fetcher

majamassarini:
I went with a simple script to run in OpenShift as a **CronJob**. Listening to Jira events required installing a java broker (as per the [guide I found](https://spaces.redhat.com/plugins/servlet/remotepageview?pageId=398889351#Plugin:JIRAMessagingPlugin-AdministrationGuide) ) 

The `QUERY` can be easily changed without redeploying the code (if we need, as an example, update the `fixVersion` values).

I run it with `podman-compose`. It found 25 tickets; the triaging agent moved 9 to the rebase queue, 1 to backport, 3 to clarification needed, 3 to error and 12 to no action.

```
COMPOSE=podman-compose make run-jira-issue-fetcher
podman-compose up -d triage-agent
```

This is still in progress because:
- I want to add some small unit tests for the script
- We need to decide:
  - how to track completed issues to prevent them from reappearing in the queue. I could query all queues and collect already-processed issues, but this data will grow quickly. Should we implement a purging mechanism (perhaps by storing the `fixVersion` for each issue)?
  - how to force an issue to be re-inserted into the queue (should I query for a specific label?)


beeai/issue_fetcher/jira_issue_fetcher.py
```diff
@@ -0,0 +1,315 @@
...
...-nihtest, python-systemd, python-pyproject-metadata, scipy, python-meson-python, pybind11, mod_wsgi, python-cups, python-typogrify, python-markdown, dbus-python, python-smartypants, python-pyqt5-sip, python3-mallard-ducktype, pycairo, pyxdg, pygobject3, fonttools, python-fs, python-cffsubr, python-appdirs, python-pycotap, python-qt5, python-pyqt6, PyQt-builder, python-scour, pyatspi, python-libevdev, python-lit, python-jsonpointer, python-sphinx_rtd_theme, python-markupsafe, python-flit-core, python-psutil, python-hatchling, python-six, python-urllib3, python-charset-normalizer, python3.12, python-wheel, python-pytest-timeout, python-iniconfig, python-pygments, python-jinja2, python-testpath, python-rpm-macros, marshalparser, babel, python-idna, python-setuptools, python-pathspec, python-packaging, python-requests, python-imagesize, PyYAML, pytest, python-sphinx-theme-alabaster, python-pyproject-hooks, python-pluggy, python-pip, pyproject-rpm-macros, python-lxml, python-docutils, py3c, python-installer, python-trove-classifiers, python-sphinx, micropipenv, python-setuptools_scm, python-sphinxcontrib-jquery, plotutils, mksh, git-lfs, "less", diffstat, groff, lftp, ftp, mc, ed, ksh)'
```
nforro, line 57: Please make the list of packages a separate file with one package per line. This is unmaintainable.

nforro:
Couldn't we reuse existing code (functions and context managers from `utils`, already defined pydantic models, Jira MCP tools)?

nforro:
For manual triggering, what I think could work nicely would be a Jira label added by maintainers/users and automatically removed by the agent once the Jira would be picked up for processing.

majamassarini:
> Couldn't we reuse existing code (functions and context managers from `utils`, already defined pydantic models, Jira MCP tools)?

I thought we don't want and agent for this.

nforro:
> > Couldn't we reuse existing code (functions and context managers from `utils`, already defined pydantic models, Jira MCP tools)?
> 
> I thought we don't want and agent for this.

I didn't say anything about an agent. But I think we will benefit from sharing (testable) pieces of code rather than reinventing the wheel multiple times.

majamassarini:
> > > Couldn't we reuse existing code (functions and context managers from `utils`, already defined pydantic models, Jira MCP tools)?
> > 
> > 
> > I thought we don't want and agent for this.
> 
> I didn't say anything about an agent. But I think we will benefit from sharing (testable) pieces of code rather than reinventing the wheel multiple times.

sorry, probably I am not following. I don't see why using mcp for making a rest call? Unless we want it later called by agents. I am not against it, if we prefer that way (I simply thought this had to be something unrelated)

lbarcziova:
> how to track completed issues to prevent them from reappearing in the queue. I could query all queues and collect already-processed issues, but this data will grow quickly. Should we implement a purging mechanism (perhaps by storing the fixVersion for each issue)?

I think the labels could help here as well (added this for tomorrows arch as well), e.g. adding label for each type of resolution at the end of triaging and then consider this in the filtering rule (don't collect issues having specific labels). We might also consider statuses, the agents but currently not changing them yet, but once they do, we can filter only NEW (maybe with combination with labels, e.g. for no-action, where the status probably won't be automatically changed).

beeai/issue_fetcher/jira_issue_fetcher.py
TomasTomecek: 100% agree to reuse that filter

owtaylor:
> @owtaylor I re-read your PR, and now I see that you aren't using an mcp call (like me) for fetching the jira issues.
> We should probably use the same code. So we need to agree on a module name and create an easily reusable method. I think. 

100% agree that we shouldn't have two implementations of doing a JIRA search result and iterating over the results. If we keep having two separate components that query jira, we should put that somewhere central.

< And I am still wondering if, instead of both of us doing the same query, shouldn't the supervisor collect the tickets from ours redis queues? I don't know your workflow, I am just wondering if it makes more sense or not.

I think I haven't fully figured out how the queue system works - so say this script decides to add an incoming issue to the triage queue - When the triage agent starts, isn't the item removed from the queue? So what keeps this script from running again at that point and adding it back?

I'll write down in some more detail about how the approaches compare (at least as I understand the queue system) and might relate.

owtaylor:
I spent some time trying to think through how the supervisor framework would work on the pre-merge side, and came up with the following pseudo-code (Compare https://github.com/owtaylor/ai-workflows/blob/testing-supervisor/beeai/supervisor/issue_workflow.py for the post-merge workflow)

This handles filing issues for triage, but also additionally other things not currently handled:

 - Flagging issues that need RM review for human attention, but only after the ROG tests complete
 - Flagging issues that get lost in the workflow for human attention

This is written down to show how supervisor works in a concrete way on a familiar example. (Though I don't think it would be a ton of work to make it real.)

```python
async def run_pre_merge_issue_workflow(issue: Issue) -> WorkflowResult:
    """
    Runs the pre-merge workflow for a single issue. This workflow should
    be run on the set of issues for the following JQL:

        filter = Jotnar_1000_packages
        AND (labels is EMPTY OR labels != "jotnar_attention_needed")
        AND status in ('New', 'In Progress')
        AND 'Fixed in Build' is EMPTY
    """
    logger.info("Running pre-merge workflow for issue %s", issue.url)

    # First check that the issue is still relevant to us (things might
    # have changed since the issue was added to the queue) 
    if "jotnar_attention_needed" in issue.labels:   
        return resolve_remove_task(issue, "Issue has jotnar_attention_needed label")

    if issue.status not in (IssueStatus.NEW, IssueStatus.IN_PROGRESS):
        return resolve_remove_task(issue, "Issue is not in New or In Progress status)

    if issue.fixed_in_build is not None:
        return resolve_remove_task(issue, "Issue has fixed_in_build")

    # Now check if the triage/rebase/backport process is in progress

    # Before a triage task finishes, it will set jotnar_triaged and either:
    #  - set jotnar_attention_needed
    #  - Add the issue to the rebase or backport queues
    if issue_in_queue("triage_queue"):
        return resolve_wait(issue, "Issue is being triaged")

    # Before a rebase or backport task finishes, it will either:
    #  - set jotnar_attention_needed
    #  - file a Merge Request
    if issue_in_queue("rebase_queue"):
        return resolve_wait(issue, "Rebase is being prepared")

    if issue_in_queue("backport_queue"):
        return resolve_wait(issue, "Backport is being prepared")

    # Now look for merge requests that have been filed
    merge_requests = await get_merge_requests(issue.key)
    for merge_request in merged_requests:
        if merge_request.state == MergeRequestState.OPEN:
            if merge_request.rog_state == RogState.FAILED:
                # Eventually, this could requeue the backport/rebase agent to try again
                return resolve_flag_attention(issue, "Merge request %s has failed ROG checks", merge_request.url)
            elif merge_request.rog_state == RogState.SUCCESS:
                return resolve_flag_attention(issue, "Merge request %s needs review", merge_request.url)
            else:
                return resolve_wait(issue, "Waiting for ROG tests to complete")
        elif merge_request.state == MergeRequestState.CLOSED:
            # Ignore closed merge requests
            continue
        elif merge_request_state == MergeRequestState.MERGED:
            # Once a merge request is merged, we expect fixed_in_build to be set by automation,
            # but it might take a bit
            if merge_request.merge_data < datetime.now() + timedelta(hours=1):
                return resolve_wait(issue, "Waiting for merged to be processed")
            else:
                return resolve_flag_attention("Automation failed to set fixed_in_build")

    # Now we either have an untriaged issue or we're in an unexpected state
    if "jotnar_triaged" not in issue.labels:
        await add_issue_to_triage_queue(issue.key)
        return resolve_wait(issue, "Issue has been added to triage queue")

    # At this point, we have no idea what's going on. We might want to have an AI
    # analyze what's going on with the issue and suggest a course of action.
    return resolve_flag_attention("Issue is stuck")
```

beeai/issue_fetcher/jira_issue_fetcher.py
```diff
@@ -0,0 +1,344 @@
...
+        time_since_last_request = current_time - self.last_request_time
+        
+        if time_since_last_request < self.RATE_LIMIT_DELAY:
+            sleep_time = self.RATE_LIMIT_DELAY - time_since_last_request
+            logger.debug(f"Rate limiting: sleeping for {sleep_time:.3f} seconds")
+            await asyncio.sleep(sleep_time)
+        
+        self.last_request_time = time.time()
+    
+    def _make_request_with_retries(self, url: str, json_data: Dict[str, Any]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
majamassarini: I was thinking of using a context manager we are already use in copr, packit and logdetective: https://github.com/fedora-copr/logdetective/blob/538d81fd66bf6c8632220dfc91a79190a92668ef/pyproject.toml#L72

Probably I will add it here too.

beeai/issue_fetcher/jira_issue_fetcher.py
```diff
@@ -0,0 +1,344 @@
...
+            await client.ping()
+            logger.info("Connected to Redis")
+            yield client
+        finally:
+            await client.aclose()
+            logger.info("Disconnected from Redis")
+    
+    async def _get_existing_issue_keys(self, redis_conn) -> set:
+        """
+        Get all existing issue keys from all Redis queues to avoid duplicates
```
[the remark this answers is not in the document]
majamassarini: We have discussed it this morning (how to prevent tasks from re-entering the queue).
We [will use labels](https://gist.github.com/majamassarini/dd7c7c7135a8fcc97536d304089adf7e) and this is a draft.
I will try to create some sort of documentation that you can check later and see if it is somehow useful in your workflows.

It should avoid the big race condition thanks to the `in_progess` labels.

beeai/issue_fetcher/jira_issue_fetcher.py
```diff
@@ -0,0 +1,344 @@
...
+                                queue_count += 1
+                        except (json.JSONDecodeError, KeyError) as e:
+                            logger.warning(f"Failed to parse item from {queue_name}: {e}")
+                            continue
+                    
+                    if queue_count > 0:
+                        logger.info(f"Found {queue_count} existing issues in {queue_name}")
+                        
+                except Exception as e:
+                    logger.warning(f"Error checking {queue_name}: {e}")
```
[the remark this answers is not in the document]
majamassarini: I see your point and this is a really good catch.
However I have already seen a task (out of 25) fail here. I don't think our code is safe enough, at the moment, for blocking everything at this point.

What I can probably do easily is making an extra call to jira (if something fails here). Collecting all jiras that have a jotnar_label, and putting them in the already processed queue. WDYT?

beeai/openshift/Makefile
```diff
@@ -1,2 +1,26 @@
...
+	@echo "Recent Jira fetcher jobs:"
+	@oc get jobs -l app=jira-issue-fetcher --sort-by=.metadata.creationTimestamp
+	@echo ""
+	@echo "Logs from latest job:"
+	@oc logs job/$$(oc get jobs -l app=jira-issue-fetcher --sort-by=.metadata.creationTimestamp -o name | tail -1 | cut -d'/' -f2) || echo "No jobs found"
+
+clean-jira-fetcher:
+	oc delete job -l app=jira-issue-fetcher
+	oc delete cronjob jira-issue-fetcher --ignore-not-found=true
+	oc delete configmap jira-issue-fetcher --ignore-not-found=true
```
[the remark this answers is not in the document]
majamassarini: I didn't test the openshift stuff yet. I also don't know if let these makefile's targets here. I think it could be handy if we have a target for running the script manually. But I don't really know if for that purpose we need this clean target. I will try. 
I will add a comment to explain why the "new targets" in the makefile, and about the `imagestream`, my purpose is only to be able to re-run the script.

beeai/openshift/configmap-x-jira-issue-fetcher.yml
```diff
@@ -0,0 +1,11 @@
...
...-nihtest, python-systemd, python-pyproject-metadata, scipy, python-meson-python, pybind11, mod_wsgi, python-cups, python-typogrify, python-markdown, dbus-python, python-smartypants, python-pyqt5-sip, python3-mallard-ducktype, pycairo, pyxdg, pygobject3, fonttools, python-fs, python-cffsubr, python-appdirs, python-pycotap, python-qt5, python-pyqt6, PyQt-builder, python-scour, pyatspi, python-libevdev, python-lit, python-jsonpointer, python-sphinx_rtd_theme, python-markupsafe, python-flit-core, python-psutil, python-hatchling, python-six, python-urllib3, python-charset-normalizer, python3.12, python-wheel, python-pytest-timeout, python-iniconfig, python-pygments, python-jinja2, python-testpath, python-rpm-macros, marshalparser, babel, python-idna, python-setuptools, python-pathspec, python-packaging, python-requests, python-imagesize, PyYAML, pytest, python-sphinx-theme-alabaster, python-pyproject-hooks, python-pluggy, python-pip, pyproject-rpm-macros, python-lxml, python-docutils, py3c, python-installer, python-trove-classifiers, python-sphinx, micropipenv, python-setuptools_scm, python-sphinxcontrib-jquery, plotutils, mksh, git-lfs, "less", diffstat, groff, lftp, ftp, mc, ed, ksh)'
```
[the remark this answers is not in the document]
majamassarini: The reason why I put the QUERY here is that I expect us to need to adjust it "quite often", and I would like to be easily editable in openshift. However I like how you used the `filter` field and I will copy your approach. Making the name of the filter editable.

beeai/agents/backport_agent.py
```diff
@@ -7,6 +7,7 @@
 
 import aiohttp
 from pydantic import BaseModel, Field
+from agents.models import Task, BackportInputSchema as InputSchema, BackportOutputSchema as OutputSchema
```
nforro, line 10: Just wondering if it wouldn't make sense to move the model definitions one level higher, to already existing [common.models](https://github.com/packit/ai-workflows/blob/67073b91373228cfc19805512f7c12e1e17f55d7/beeai/common/models.py), so we have them all in one place?

beeai/issue_fetcher/jira_issue_fetcher.py
```diff
@@ -0,0 +1,381 @@
...
+    @asynccontextmanager
+    async def redis_client(self) -> AsyncGenerator[redis.Redis, None]:
+        client = redis.Redis.from_url(self.redis_url)
+        try:
+            await client.ping()
+            logger.info("Connected to Redis")
+            yield client
+        finally:
+            await client.aclose()
+            logger.info("Disconnected from Redis")
```
nforro, line 183: This is already in `agents.utils` and could be also moved to `common`:

https://github.com/packit/ai-workflows/blob/67073b91373228cfc19805512f7c12e1e17f55d7/beeai/agents/utils.py#L89-L96

beeai/issue_fetcher/jira_issue_fetcher.py
```diff
@@ -0,0 +1,381 @@
...
+            queue_names = [
+                "triage_queue",
+                "rebase_queue", 
+                "backport_queue",
+                "clarification_needed_queue",
+                "error_list",
+                "no_action_list",
+                "completed_rebase_list",
+                "completed_backport_list"
+            ]
```
lbarcziova, line 200: shouldn't we be checking here only the first 3? For the other ones, we want to be able to retry those, or am I missing something?

beeai/issue_fetcher/jira_issue_fetcher.py
```diff
@@ -0,0 +1,381 @@
...
+            # Extend existing_keys with issues that have jötnar labels (except jotnar_retry_needed)
+            for issue in issues:
+                issue_key = issue.get("key")
+                if issue_key:
+                    fields = issue.get("fields", {})
+                    labels = fields.get("labels", [])
+                    jotnar_labels = [label for label in labels if label.startswith('jotnar_')]
+                    
+                    # If issue has jötnar labels other than jotnar_retry_needed, mark as existing
+                    if jotnar_labels and jotnar_labels != ['jotnar_retry_needed']:
```
lbarcziova, line 299: from the workflow diagram, I thought if there is `jotnar_retry_needed`, despite what the rest of the labels are set, the task will be retried, am I wrong?

lbarcziova (commented):
looks good! This is still missing the openshift cronjob definition, right?

beeai/issue_fetcher/jira_issue_fetcher.py
lbarcziova: but we are checking only the Jira issue key here, while there can be e.g. a new comment or a link, for that case, we would still skip the reentering based on the key, or?

beeai/agents/__pycache__/triage_agent.cpython-313.pyc
nforro, line 1: Please no bytecode (we should probably update `.gitignore`) :slightly_smiling_face:

beeai/issue_fetcher/jira_issue_fetcher.py
majamassarini: When I answered your question I was thinking at the labels, now I see we are talking about the redis queues. Probably I shouldn't rely at all on the redis queues but only on the labels. However at the moment the labels are not implemented, so this is the only way I have right now.

And I really need to add some more unit tests. I am working on them I hope I will finish them soon.

majamassarini: > but we are checking only the Jira issue key here, while there can be e.g. a new comment or a link, for that case, we would still skip the reentering based on the key, or?

Now the `retry_needed` label should work, it removed the already added key from those skipped.

But as I said, I would like to test all this logic further with unit tests.

beeai/openshift/Makefile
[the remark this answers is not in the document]
majamassarini: I removed this target.

beeai/.gitignore
```diff
@@ -1 +1,2 @@
 uv.lock
+*__pycache__*
```
nforro, line 2: Perhaps we can add what we have in packit?

```suggestion
*.py[cod]
.tox
noarch/
*.egg-info/
*.eggs
build
*.idea
*.vscode

__pycache__/
*$py.class
dist/
eggs/
.eggs/
sdist/
.installed.cfg

.env
.venv
env/
venv/
ENV/

*.bak
*~

*.rpm
*.tar.gz

.coverage*
```

beeai/Makefile
```diff
@@ -48,6 +48,26 @@ run-backport-agent-standalone:
...
+	@echo "Running Jira Issue Fetcher..."
+	@if [ ! -f .secrets/jira-issue-fetcher.env ]; then \
+		echo "Error: .secrets/jira-issue-fetcher.env not found"; \
+		echo "Copy the template: cp templates/jira-issue-fetcher.env .secrets/jira-issue-fetcher.env"; \
+		echo "Then edit it with your credentials"; \
+		exit 1; \
+	fi
+	@if [ ! -f .secrets/keytab ]; then \
+		echo "Warning: .secrets/keytab not found - Kerberos authentication will be skipped"; \
+	fi
```
nforro, line 62: Should this still be here?

beeai/common/models.py
```diff
@@ -30,3 +35,129 @@ class CVEEligibilityResult(BaseModel):
...
         description="Error message if the issue cannot be processed"
     )
+
+
+class TriageInputSchema(BaseModel):
+    """Input schema for the triage agent - metadata for a JIRA issue task."""
+    issue: str = Field(description="JIRA issue key (e.g., RHEL-12345)")
+    
+    def to_dict(self) -> Dict[str, Any]:
+        """Convert to dictionary for compatibility with existing code."""
```
nforro, line 45: What existing code? This is just a rename of `model_dump()` to `to_dict()` (same as `to_json()` below), do we really need it?

beeai/agents/utils.py
```diff
@@ -87,15 +86,6 @@ async def run_tool(
...
 
-@asynccontextmanager
-async def redis_client(redis_url: str) -> AsyncGenerator[redis.Redis, None]:
-    client = redis.Redis.from_url(redis_url)
-    await client.ping()
-    try:
-        yield client
-    finally:
-        await client.aclose()
-
```
nforro, line 98: I believe the `fix_await()` below is closely related to `redis_client()` and should be thus moved together with it (and used to wrap the redis methods, in the same way as in the agents).

lbarcziova (approved):
thanks Maja, there are also some good structural improvements here! :rocket: 

I'm ok with merging and let's adjust as needed in followups and once we have the retries

I have just one remark related to the retries now in the previous thread

beeai/issue_fetcher/jira_issue_fetcher.py
majamassarini: I would say that once we have the labels implemented for every jotnar workflow action I can remove checking the redis queues.
And the `jotnar_retry_needed` label should be requeued, thanks to [this](https://github.com/packit/ai-workflows/blob/36e884ce8de3bb510f3088fc0ad4709c8094921d/beeai/issue_fetcher/jira_issue_fetcher.py#L295-L297)

PR #107, opened 2025-08-28
Build verification

nforro:
There will be a follow-up to add the build agent to the backport workflow as well.

### DIFF
--- a/beeai/agents/rebase_agent.py
+++ b/beeai/agents/rebase_agent.py
@@ -94,6 +94,7 @@ async def main() -> None:
 
     async with mcp_tools(os.environ["MCP_GATEWAY_URL"]) as gateway_tools:
         rebase_agent = RequirementAgent(
+            name="Rebase",
             llm=ChatModel.from_name(os.environ["CHAT_MODEL"]),
             tools=[
                 ThinkTool(),
--- a/beeai/agents/rebase_agent.py
+++ b/beeai/agents/rebase_agent.py
@@ -50,15 +50,15 @@ def render_prompt(input: InputSchema) -> str:
 
       Follow exactly these steps:
 
-      1. You will find the cloned dist-git repository of the {{ package }} package in {{ local_clone }}.
+      1. You will find the cloned dist-git repository of the {{package}} package in {{local_clone}}.
          It is your current working directory, do not `cd` anywhere else.
 
-      2. Check if the {{ package }} was not already updated to version {{ version }}. That means comparing
+      2. Check if the {{package}} was not already updated to version {{version}}. That means comparing
          the current version with the provided version.
           * The current version of the package can be found in the 'Version' field of the spec file.
           * If there is nothing to update, print a message and exit. Otherwise follow the instructions below.
 
-      3. Update the {{ package }} to the newer version:
+      3. Update the {{package}} to the newer version:
           * Update the local package by:
             * Updating the 'Version' and 'Release' fields (or corresponding macros) in the spec file as needed,
               following packaging documentation.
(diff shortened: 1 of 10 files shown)


TomasTomecek (approved):
Very nice work!! Love your improvements to the rebase agent. We should unify the backport and rebase agents at some points with the shared steps.

Also the telemetry is amazing.

beeai/mcp_server/tests/unit/test_copr.py
```diff
@@ -91,3 +94,36 @@ async def sleep(*_):
...
+            return content
+        yield flexmock(status=404 if "broken" in url else 200, reason="Because", read=read)
+
+    flexmock(aiohttp.ClientSession).should_receive("get").replace_with(get)
+    flexmock(gzip).should_receive("decompress").and_return(content).times(
+        1 if url.endswith(".log.gz") and not "broken" in url else 0
+    )
+    result = await download_artifacts(artifacts_urls=artifacts_urls, target_path=target_path)
+    assert result.startswith("Failed" if "broken" in url else "Successfully")
+    path = target_path / url.rsplit("/", 1)[-1].removesuffix(".gz")
```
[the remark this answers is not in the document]
nforro: True, but the test URLs are simple on purpose, also for this reason.

beeai/agents/rebase_agent.py
```diff
@@ -39,70 +41,110 @@ class InputSchema(BaseModel):
...
+
+      1. Check if the current version is older than <VERSION>. To get the current version,
+         you can use `rpmspec -q --queryformat "%{VERSION}\n" --srpm <PACKAGE>.spec`.
+         To compare versions, use `rpmdev-vercmp`. If the current version is not older than <VERSION>,
+         rebasing doesn't make sense, so end the process with an error.
+
+      2. Try to find past rebases in git history to see how this particular package does rebases.
+         Keep in mind what parts of the spec file are usually changed. At the minimum a rebase should
+         change `Version` and `Release` tags (or corresponding macros) and add a new changelog entry,
+         but sometimes other things are changed - if that's the case, try to understand the logic behind it.
```
lbarcziova, line 68: I'm curious if this could potentially help with cases like LLVM rebase, but I expect there need to be more explicit instructions for that

beeai/agents/rebase_agent.py
```diff
@@ -177,9 +244,26 @@ async def run_rebase_agent(state):
...
+                expected_output=BuildOutputSchema,
+                execution=get_agent_execution_config(),
+            )
+            build_result = BuildOutputSchema.model_validate_json(response.answer.text)
+            if build_result.success:
                 return "commit_push_and_open_mr"
-            else:
-                return "comment_in_jira"
+            state.build_error = build_result.error
+            return "run_rebase_agent"
```
lbarcziova, line 266: neat 🚀 

how do we prevent here the agent to infinite loop?

beeai/agents/rebase_agent.py
```diff
@@ -39,70 +41,110 @@ class InputSchema(BaseModel):
...
+      Your working directory is {{local_clone}}, a clone of dist-git repository of package {{package}}.
+      {{ dist_git_branch }} dist-git branch has been checked out.
+      {{^build_error}}
+      Rebase the package to version {{version}}. Use "- resolves: {{jira_issue}}" as changelog entry.
+      {{/build_error}}
+      {{#build_error}}
+      This is a repeated rebase, after the previous attempt the generated SRPM failed to build:
+
+      {{.}}
+
```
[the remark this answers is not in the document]
nforro: It definitely needs more testing, but I would say it works quite well. One time the build agent reported noting more than `Build NNN failed` (I forgot to import `aiohttp` when I implemented the `download_artifacts` tool) and the agent deduced that if the build failed it must be because of a removed patch and decided to try adding them back one by one, running build after each change. And that actually worked (because the build would fail on account of one removed patch). But the patch handling is another story, it will have to be implemented as a tool.

beeai/agents/rebase_agent.py
nforro: That's a good question :sweat_smile: I suppose we have to add some counter and give up after a number of failed attempts.

lbarcziova: sounds good

beeai/agents/rebase_agent.py
```diff
@@ -224,6 +308,7 @@ async def comment_in_jira(state):
 
         workflow.add_step("fork_and_prepare_dist_git", fork_and_prepare_dist_git)
         workflow.add_step("run_rebase_agent", run_rebase_agent)
+        workflow.add_step("run_build_agent", run_build_agent)
```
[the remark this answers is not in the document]
nforro: > I am just curious if you have tried something like in the example [here](https://framework.beeai.dev/modules/workflows#multi-agent-workflows), where the agents are just added to the workflow but not directly called. The prompts are triggering the agents as far as I understand it.

When you use the `AgentWorkflow` and `add_agent()`, it runs them in sequence, feeding them prompts from the `inputs` list. They share memory, so they know the prompts and conversation of the previously-executed agents, but that's all, there is no triggering-by-agent going on. Quite honestly I don't know what practical use cases this has, except fo the one in the example.

PR #124, opened 2025-09-01
Verify author of Jira issue for rebase resolutions

lbarcziova:
Fixes packit/jotnar#147

### DIFF
--- a/beeai/agents/triage_agent.py
+++ b/beeai/agents/triage_agent.py
@@ -350,7 +350,9 @@ async def main() -> None:
                 )
                 state.triage_result = OutputSchema.model_validate_json(response.answer.text)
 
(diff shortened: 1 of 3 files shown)


siteshwar (approved):

lbarcziova:
@nforro I think for general bug reports, the ticket can also be filed by a customer or someone not in the RH associate group, while still being relevant for Jotnar, but we can definitely discuss extending more. 

But I wanted to introduce this for rebase more strictly, as there we are doing lookaside cache uploads and I want to also play for the Fedora usecase with consuming some additional instructions from Jira description directly when doing rebase (before we have https://github.com/packit/jotnar/issues/146) , so I wanted to make it more secure.

PR #133, opened 2025-09-02
Issue fetcher tests

majamassarini:


beeai/agents/tests/unit/test_tools.py
```diff
@@ -328,6 +333,11 @@ async def test_git_patch_creation_tool_success(git_repo, tmp_path):
     new_file = git_repo / "file.txt"
     new_file.write_text("Line 1\nLine 3\n")
     subprocess.run(["git", "add", "file.txt"], cwd=git_repo, check=True)
+
+    # Ensure git user identity is configured for this commit
+    subprocess.run(["git", "config", "user.name", "Test User"], cwd=git_repo, check=True)
+    subprocess.run(["git", "config", "user.email", "test@example.com"], cwd=git_repo, check=True)
```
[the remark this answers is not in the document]
majamassarini: If I don't set this the tests will fail inside the container

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
```diff
@@ -0,0 +1,393 @@
...
+    # Create mock Redis object
+    mock_redis = flexmock()
+
+    # Create a proper async context manager
+    mock_context_manager = MockAsyncContextManager(mock_redis)
+
+    # Mock the redis_client function in the jira_issue_fetcher module where it's imported
+    flexmock(sys.modules['jira_issue_fetcher']).should_receive('redis_client').and_return(mock_context_manager)
+
+    return mock_redis, mock_context_manager
```
nforro, line 57: This seems overly compilcated, wouldn't something like this work?

```suggestion
@pytest.fixture
def mock_redis_context():
    """Create a mock Redis context manager for testing."""
    mock_redis = flexmock()
    
    @asynccontextmanager
    async def mock_context_manager(*_, **__):
        yield mock_redis
    
    flexmock(jira_issue_fetcher).should_receive("redis_client").and_return(mock_context_manager)

    return mock_redis, mock_context_manager
```

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
```diff
@@ -0,0 +1,393 @@
...
+
+def create_async_mock_return_value(value):
+    """Create a mock awaitable that returns the given value when awaited."""
+    async def async_return():
+        return value
+    return async_return()
```
nforro, line 63: This could be made common and used in other tests.

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
majamassarini: I tried your suggestion and it doesn't seem to work.
I agree this is complex, it took me a while to create this mock objects. But I think they are all needed. I need to mock the result of the `__aenter__`  method (for some of the tests but not all). This should be why we need a more complex mock object and the `asynccontextmanager` you suggested is not working.

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
majamassarini: Agreed. However at the moment we don't have a place to share code between the tests. I think we should first address [this](https://github.com/packit/jotnar/issues/148) and after it this should be simpler. WDYT?

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
nforro: `__aenter__` returns the context variable, which is what an `@asynccontextmanager`-decorated generator does when yielding. I'm not saying the suggested code will work as it is, it was just an example, but I don't see why it would be impossible to do it this way.

nforro: I can try to do some experiments and suggest a working code (unless I fail :slightly_smiling_face:).

beeai/Makefile.tests
```diff
@@ -13,6 +13,8 @@ check:
 	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 python3 -m pytest --verbose --showlocals $(TEST_TARGET) || true
 	cd ./mcp_server && \
 	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 python3 -m pytest --verbose --showlocals $(TEST_TARGET)
+	cd ./jira_issue_fetcher && \
+	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 python3 -m pytest --verbose --showlocals $(TEST_TARGET)
```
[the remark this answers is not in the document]
majamassarini: I think this was solved by Tomas having multiple targets now.

beeai/Containerfile.tests
```diff
@@ -11,11 +11,17 @@ RUN dnf -y install \
       python3-pytest \
       python3-pytest-asyncio \
       python3-specfile \
+      python3-redis \
+      python3-requests \
+      python3-backoff \
     && dnf clean all
```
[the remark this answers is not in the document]
majamassarini: We usually do prefer installing dependencies from dnf, when possible. Am I wrong?

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
```diff
@@ -0,0 +1,415 @@
...
+    """Test searching issues with pagination."""
+    mock_issues_page1 = [{'key': 'TEST-1', 'fields': {'labels': []}}]
+    mock_issues_page2 = [{'key': 'TEST-2', 'fields': {'labels': []}}]
+
+    # Mock _rate_limit to return an awaitable coroutine, we can't reuse an awaitable coroutine
+    async def mock_rate_limit_1():
+        pass
+    async def mock_rate_limit_2():
+        pass
+    flexmock(fetcher).should_receive('_rate_limit').and_return(mock_rate_limit_1()).and_return(mock_rate_limit_2())
```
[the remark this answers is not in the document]
majamassarini: No you can't, this is the output:

```
FAILED tests/unit/test_jira_issue_fetcher.py::test_search_issues_multiple_pages - RuntimeError: cannot reuse already awaited coroutine
```

beeai/jira_issue_fetcher/jira_issue_fetcher.py
```diff
@@ -209,41 +201,38 @@ async def _get_existing_issue_keys(self, redis_conn: redis.Redis) -> set[str]:
...
-                                        schema = BackportInputSchema.model_validate(task.metadata)
-                                        issue_key = schema.jira_issue.upper()
+                                    match queue_name:
+                                        case RedisQueues.TRIAGE_QUEUE.value:
+                                            schema = TriageInputSchema.model_validate(task.metadata)
+                                            issue_key = schema.issue.upper()
+                                        case RedisQueues.REBASE_QUEUE.value | RedisQueues.BACKPORT_QUEUE.value | RedisQueues.CLARIFICATION_NEEDED_QUEUE.value:
+                                            issue_key = task.metadata.get("jira_issue", "").upper()
+                                        case _:
+                                            continue
```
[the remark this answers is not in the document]
majamassarini: Yes, but the point was to reuse pydantic definitions instead of guessing the field.

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
```diff
@@ -0,0 +1,415 @@
...
+        return self.mock_redis
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+@pytest.fixture
+def mock_redis_context():
+    """Create a mock Redis context manager for testing."""
+    import sys
+    from jira_issue_fetcher import redis_client
```
nforro, line 46: What's the purpose of these imports?

beeai/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py
```diff
@@ -0,0 +1,415 @@
...
+
+    # Create mock Redis object
+    mock_redis = flexmock()
+
+    # Create a proper async context manager
+    mock_context_manager = MockAsyncContextManager(mock_redis)
+
+    # Mock the redis_client function in the jira_issue_fetcher module where it's imported
+    flexmock(sys.modules['jira_issue_fetcher']).should_receive('redis_client').and_return(mock_context_manager)
```
nforro, line 55: Why not this?

```suggestion
    flexmock(jira_issue_fetcher).should_receive('redis_client').and_return(mock_context_manager)
```

PR #152, opened 2025-09-05
commit_and_push: check staged content before commit

TomasTomecek:
...and halt if nothing is staged

I orginally wanted to fix a case so a rebase agent reuses MRs but realized it's already done.

We have two equal rebases here [!17](https://gitlab.com/redhat/centos-stream/rpms/readline/-/merge_requests/17), [!16](https://gitlab.com/redhat/centos-stream/rpms/readline/-/merge_requests/16) because of a bug fixed in b4017b0513a950befa3673366f7541da7157120a

### DIFF
--- a/beeai/agents/tasks.py
+++ b/beeai/agents/tasks.py
@@ -69,7 +69,14 @@ async def commit_push_and_open_mr(
         files_to_commit = [files_to_commit]
     for path in itertools.chain(*(local_clone.glob(pat) for pat in files_to_commit)):
         await check_subprocess(["git", "add", str(path)], cwd=local_clone)
-    # TODO: check for empty commit (the command below will fail anyway, but we need to handle this somehow)
+    # Check how many files are staged before committing, if none, bail
+    stdout, _ = await check_subprocess(
+        ["git", "diff", "--cached", "--name-only"],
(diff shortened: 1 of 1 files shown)


beeai/agents/tasks.py
```diff
@@ -69,7 +69,15 @@ async def commit_push_and_open_mr(
...
-    # TODO: check for empty commit (the command below will fail anyway, but we need to handle this somehow)
+    # Check if any files are staged before committing, if none, bail
+    exit_code, _, _ = await run_subprocess(
+        ["git", "diff", "--cached", "--quiet"],
+        cwd=local_clone,
+    )
+    # 1 = staged, 0 = none staged
+    if exit_code == 0:
+        logger.info("No files staged for commit, halting.")
+        return "No files staged for commit, halting."
```
nforro, line 80: We should probably raise an exception here, this will not halt the process, it will let it continue with MR URL set to that text :slightly_smiling_face:

nforro: Why would it set status to failed? There is no such logic in the code. It will succeed with `merge_request_url: No files staged for commit, halting.`.

nforro (approved):

PR #154, opened 2025-09-08
PatchCreationTool now accepts patches w/o headers

TomasTomecek:
if the name/email is not specified, it fails with

> fatal: empty ident name (for <>) not allowed

not a fatal error actually, the patch is still applied but git doesn't
know how to commit it, so we have to do it instead

The root cause is solved, but I still can't get a good backport.

This PR also adds a new tool `insert_after_substring` so the LLM can reference specific text as a location instead of a line number where they can easily make mistakes if they read the whole file.

Fixes https://github.com/packit/jotnar/issues/158

### DIFF
--- a/beeai/agents/tools/text.py
+++ b/beeai/agents/tools/text.py
@@ -89,7 +89,7 @@ class InsertToolInput(BaseModel):
 class InsertTool(Tool[InsertToolInput, ToolRunOptions, StringToolOutput]):
     name = "insert"
     description = """
-    Inserts the specified text at a specific location in a file.
+    Inserts the specified text at a specific location in a file
     """
     input_schema = InsertToolInput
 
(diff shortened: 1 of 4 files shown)


beeai/agents/tools/text.py
```diff
@@ -89,7 +89,7 @@ class InsertToolInput(BaseModel):
 class InsertTool(Tool[InsertToolInput, ToolRunOptions, StringToolOutput]):
     name = "insert"
     description = """
-    Inserts the specified text at a specific location in a file.
+    Inserts the specified text at a specific location in a file
```
nforro, line 92: Is there a reason for this? :slightly_smiling_face:

beeai/agents/tools/text.py
```diff
@@ -111,6 +111,51 @@ async def _run(
...
+    file: AbsolutePath = Field(description="Absolute path to a file to edit")
+    insert_after_substring: str = Field(description="Substring to insert the text after")
+    new_string: str = Field(description="Text to insert")
+
+
+class InsertAfterSubstringTool(Tool[InsertAfterSubstringToolInput, ToolRunOptions, StringToolOutput]):
+    name = "insert_after_substring"
+    description = """
+    Inserts the provided text new_string on a new line after the first
+    occurrence of the specified substring insert_after_substring. The insertion
```
nforro, line 124:
```suggestion
    Inserts the provided text `new_string` on a new line after the first
    occurrence of the specified substring `insert_after_substring`. The insertion
```

beeai/agents/tools/text.py
```diff
@@ -111,6 +111,51 @@ async def _run(
...
+        return Emitter.root().child(
+            namespace=["tool", "text", self.name],
+            creator=self,
+        )
+
+    async def _run(
+        self, tool_input: InsertAfterSubstringToolInput, options: ToolRunOptions | None, context: RunContext
+    ) -> StringToolOutput:
+        if not tool_input.insert_after_substring:
+            raise ToolError("No insertion was done because the specified substring wasn't provided")
```
nforro, line 139: You could just type the argument as `NonEmptyString`:
https://github.com/packit/ai-workflows/blob/7cd29d3ce5209f38bdf57d72398f355cef86cae7/beeai/common/validators.py#L20

lbarcziova (approved):
lgtm

nforro (approved):

PR #162, opened 2025-09-10
beeai/pyproject.yml: require beeai-framework<=0.1.41

owtaylor:
beeai-framework 0.1.42 introduces an API break for agent.run()
(https://github.com/i-am-bee/beeai-framework/pull/1013)


owtaylor:
While I'd like to see a move to having all dependencies versioned through pyproject.toml, instead of ad-hoc in the different Containerfiles, I think this can be closed.

PR #179, opened 2025-09-16
beeai: clone repositories to directory named after the package name

grulja:
In case of an already existing fork created with "centpkg fork", the repository will have the namespace in its name, e.g. centos_rpms_foo. Since we later assume the repository is in /git-repos/RHEL-XXX/package we have to specify the directory name to match the expectations.

### DIFF
--- a/beeai/agents/tasks.py
+++ b/beeai/agents/tasks.py
@@ -32,7 +32,7 @@ async def fork_and_prepare_dist_git(
     local_clone = working_dir / package
     shutil.rmtree(local_clone, ignore_errors=True)
     await check_subprocess(
-        ["git", "clone", "--single-branch", "--branch", dist_git_branch, fork_url],
+        ["git", "clone", "--single-branch", "--branch", dist_git_branch, fork_url, package],
         cwd=working_dir,
     )
     update_branch = f"{BRANCH_PREFIX}-{jira_issue}"


nforro:
> In case of an already existing fork created with "centpkg fork", the repository will have the namespace in its name, e.g. centos_rpms_foo.

Oh, there is a naming convention for forks? Does `rhpkg fork` create `rhel_rpms_foo`? We can adapt to that, it makes sense.

grulja:
> > In case of an already existing fork created with "centpkg fork", the repository will have the namespace in its name, e.g. centos_rpms_foo.
> 
> Oh, there is a naming convention for forks? Does `rhpkg fork` create `rhel_rpms_foo`? We can adapt to that, it makes sense.

It does:
```
[jgrulich@toolbx libxml2 (c10s)]$ centpkg fork
Fork of the repository has been created: 'https://gitlab.com/jgrulich/centos_rpms_libxml2'
Adding as remote 'jgrulich'.
```

grulja:
I wouldn't always expect the fork to have the name to consists from the namespace, because you can also make a fork manually or rename it. Either always clone to the directory named according to the package name, like I do in this change or get the directory it was cloned into and use that one.

nforro (approved):
:+1:

beeai/agents/tasks.py
```diff
@@ -32,7 +32,7 @@ async def fork_and_prepare_dist_git(
     local_clone = working_dir / package
     shutil.rmtree(local_clone, ignore_errors=True)
     await check_subprocess(
-        ["git", "clone", "--single-branch", "--branch", dist_git_branch, fork_url],
+        ["git", "clone", "--single-branch", "--branch", dist_git_branch, fork_url, f"{package}"],
```
nforro, line 35: Actually, the f-string is useless here:

```suggestion
        ["git", "clone", "--single-branch", "--branch", dist_git_branch, fork_url, package],
```

grulja: Ignore me, I guess it will work the same way as for the other arguments

PR #188, opened 2025-09-17
Improve formatting of triage agent comment

lbarcziova:
Fixes packit/jotnar#145

### DIFF
--- a/beeai/common/models.py
+++ b/beeai/common/models.py
@@ -163,6 +163,53 @@ class TriageOutputSchema(BaseModel):
         description="Associated data"
     )
 
+    def format_for_comment(self) -> str:
+        """Format the triage result in a human-readable format for Jira comments."""
+        resolution = f"*Resolution*: {self.resolution.value}\n"
+
+        match self.data:
+            case BackportData():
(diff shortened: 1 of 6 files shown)


beeai/common/models.py
```diff
@@ -163,6 +163,53 @@ class TriageOutputSchema(BaseModel):
         description="Associated data"
     )
 
+    def format_for_comment(self) -> str:
+        """Format the triage result in a human-readable format for Jira comments."""
+        resolution = f"*Resolution*: {self.resolution.value}\n"
+
+        match self.data:
+            case BackportData():
```
nforro, line 171: Is this an [undocumented](https://docs.pydantic.dev/latest/concepts/models/#structural-pattern-matching) feature? No arguments match the same model with arbitrary attribute values?

nforro (approved):
I was thinking about using some JSON to MD library for this, or writing our own generic formatter, but this approach obviously gives us the most control over the result.

TomasTomecek (commented):
oh my, finally! we'll have pretty output and not the unreadable json, thank you!

PR #197, opened 2025-09-23
Clone repos in the mcp gateway

majamassarini:
Where we can also access the private repos.

### DIFF
--- a/agents/tasks.py
+++ b/agents/tasks.py
@@ -32,10 +32,6 @@ async def fork_and_prepare_dist_git(
     local_clone = working_dir / package
     shutil.rmtree(local_clone, ignore_errors=True)
 
-    # Clone the repository using the clone_repository tool
-    # for this to work, local_clone should be inside a volume shared by the mcp server and the agents
-    # the volume is now mounted in /git-repos in both the mcp server and the agents
-    # and GIT_REPO_BASEPATH is set to /git-repos
     await run_tool(
         "clone_repository",
         repository=fork_url,
--- a/agents/tasks.py
+++ b/agents/tasks.py
@@ -31,9 +31,17 @@ async def fork_and_prepare_dist_git(
     )
     local_clone = working_dir / package
     shutil.rmtree(local_clone, ignore_errors=True)
-    await check_subprocess(
-        ["git", "clone", "--single-branch", "--branch", dist_git_branch, fork_url, package],
-        cwd=working_dir,
+
+    # Clone the repository using the clone_repository tool
+    # for this to work, local_clone should be inside a volume shared by the mcp server and the agents
+    # the volume is now mounted in /git-repos in both the mcp server and the agents
+    # and GIT_REPO_BASEPATH is set to /git-repos
+    await run_tool(
+        "clone_repository",
+        repository=fork_url,
+        clone_path=str(local_clone),
+        branch=dist_git_branch,
+        available_tools=available_tools,
     )
     update_branch = f"{BRANCH_PREFIX}-{jira_issue}"
     await check_subprocess(["git", "checkout", "-B", update_branch], cwd=local_clone)
--- a/mcp_server/gitlab_tools.py
+++ b/mcp_server/gitlab_tools.py
@@ -17,6 +17,16 @@ from common.validators import AbsolutePath
 logger = logging.getLogger(__name__)
 
 
+def _get_authenticated_url(repository_url: str) -> str:
+    """
+    Helper function to add GitLab token authentication to repository URLs.
+    """
+    if token := os.getenv("GITLAB_TOKEN"):
+        url = urlparse(repository_url)
+        return url._replace(netloc=f"oauth2:{token}@{url.hostname}").geturl()
+    return repository_url
+
+
 async def fork_repository(
     repository: Annotated[str, Field(description="Repository URL")],
 ) -> str:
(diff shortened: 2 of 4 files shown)


mcp_server/gitlab_tools.py
```diff
@@ -14,9 +14,21 @@
 from common.validators import AbsolutePath
 
 
+
```
nforro, line 17:
```suggestion
```

mcp_server/gitlab_tools.py
```diff
@@ -14,9 +14,21 @@
...
 logger = logging.getLogger(__name__)
 
 
+def _get_authenticated_url(repository: str) -> str:
+    """
+    Helper function to add GitLab token authentication to repository URLs.
+    """
+    token = os.getenv("GITLAB_TOKEN")
+    if token:
```
nforro, line 26:
```suggestion
    if token := os.getenv("GITLAB_TOKEN"):
```

mcp_server/gitlab_tools.py
```diff
@@ -14,9 +14,21 @@
 from common.validators import AbsolutePath
 
 
+
 logger = logging.getLogger(__name__)
 
 
+def _get_authenticated_url(repository: str) -> str:
```
nforro, line 21:
```suggestion
def _get_authenticated_url(repository_url: str) -> str:
```

mcp_server/gitlab_tools.py
```diff
@@ -135,6 +147,29 @@ async def get_internal_rhel_branches(
...
 
 
+async def clone_repository(
+    repository: Annotated[str, Field(description="Repository URL to clone")],
+    clone_path: Annotated[str, Field(description="Absolute path where to clone the repository")],
+    branch: Annotated[str, Field(description="Branch to clone")] = None,
+) -> str:
+    """
+    Clones the specified repository to the given local path.
+    Uses GitLab token for authentication if available.
```
nforro, line 157: Even though this tool is called from code, not by a model, it is still a MCP tool that could be used by a model, and this sentence could potentially confuse it. The consumers of this tool shouldn't need to know anything about any authentication or tokens.

mcp_server/gitlab_tools.py
```diff
@@ -135,6 +147,29 @@ async def get_internal_rhel_branches(
...
+    """
+    # Add authentication token to repository URL if available
+    authenticated_url = _get_authenticated_url(repository)
+
+    command = ["git", "clone", "--single-branch", "--branch", branch]
+
+    command.extend([authenticated_url, clone_path])
```
nforro, line 164:
```suggestion
    command = [
        "git",
        "clone",
        "--single-branch",
        "--branch",
        branch,
        authenticated_url,
        clone_path,
    ]
```

nforro (commented):
Let's make sure to purge the remotes from the local clone not to leak the token.

mcp_server/gitlab_tools.py
```diff
@@ -135,6 +145,41 @@ async def get_internal_rhel_branches(
...
+        branch,
+        authenticated_url,
+        clone_path,
+    ]
+
+    proc = await asyncio.create_subprocess_exec(command[0], *command[1:])
+    if await proc.wait():
+        raise ToolError(f"Failed to clone repository {repository}")
+
+    # Remove all remote references after cloning
```
nforro, line 173: The comment says all but the command removes only `origin`. Sure, only `origin` should exist as it is autocreated, but the comment should reflect that. Also, to be 100% safe, we should really remove all remotes. And we can't rely on the name `origin` as it is possible to change that for example with a global git config.
I think we should do something like `for remote in $(git remote); do git remote remove $remote; done`. And it would probably be more elegant to use GitPython for this (and other git operations).

agents/tasks.py
```diff
@@ -31,9 +31,17 @@ async def fork_and_prepare_dist_git(
...
     local_clone = working_dir / package
     shutil.rmtree(local_clone, ignore_errors=True)
-    await check_subprocess(
-        ["git", "clone", "--single-branch", "--branch", dist_git_branch, fork_url, package],
-        cwd=working_dir,
+
+    # Clone the repository using the clone_repository tool
+    # for this to work, local_clone should be inside a volume shared by the mcp server and the agents
+    # the volume is now mounted in /git-repos in both the mcp server and the agents
+    # and GIT_REPO_BASEPATH is set to /git-repos
```
nforro, line 38: This is true for all other tools that accept `clone_path`.

nforro (approved):
Let's merge this and iron out any issues in a follow-up.

mcp_server/tests/unit/test_gitlab_tools.py
```diff
@@ -118,6 +118,22 @@ def create_pr_raises(*args, **kwargs):
...
+
+    async def create_subprocess_exec(cmd, *args, **kwargs):
+        async def wait():
+            return 0
+        return flexmock(wait=wait)
+
+    flexmock(asyncio).should_receive("create_subprocess_exec").with_args("git", "clone", "--single-branch", "--branch", branch, repository, clone_path).replace_with(create_subprocess_exec)
+    flexmock(asyncio).should_receive("create_subprocess_exec").with_args("git", "remote", "remove", "origin", cwd=clone_path).replace_with(create_subprocess_exec)
+    result = await clone_repository(repository=repository, clone_path=clone_path, branch=branch)
+    assert result.startswith("Successfully")
```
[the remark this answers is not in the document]
nforro: Partially :+1:

mcp_server/gitlab_tools.py
```diff
@@ -135,6 +145,41 @@ async def get_internal_rhel_branches(
...
+
+    command = [
+        "git",
+        "clone",
+        "--single-branch",
+        "--branch",
+        branch,
+        authenticated_url,
+        clone_path,
+    ]
```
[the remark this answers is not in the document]
nforro: > The `clone_repository` function defines `branch` as an optional argument with a default value of `None`

Oh, it actually does. We should change that.

PR #207, opened 2025-09-26
Collect files from all rebase agent iterations to git add

lbarcziova:
If the build agent reports failure and the rebase agent is rerun again, we might not git add all the files without this change.


TomasTomecek (approved):
nice catch

nforro (approved):

PR #216, opened 2025-09-29
Feat: Add status transition timestamp to Erratum

Jazzcort:
Add the latest status transition timestamp to Erratum so the agent can use this infomation to decide whether the erratum needs human attention due to long pending time.

### DIFF
--- a/supervisor/errata_utils.py
+++ b/supervisor/errata_utils.py
@@ -1,3 +1,4 @@
+from datetime import datetime, timezone
 from enum import StrEnum
 from functools import cache
 import logging
@@ -77,6 +78,10 @@ def get_erratum(erratum_id: str | int):
         for jira_issue_data in jira_issues
     )
 
+    last_status_transition_timestamp = datetime.strptime(
+        details["status_updated_at"], "%Y-%m-%dT%H:%M:%SZ"
+    ).replace(tzinfo=timezone.utc)
+
     return Erratum(
         id=details["id"],
         full_advisory=details["fulladvisory"],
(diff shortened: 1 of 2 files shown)


supervisor/errata_utils.py
```diff
@@ -1,14 +1,15 @@
-from enum import StrEnum
-from functools import cache
 import logging
 import os
+from datetime import datetime, timezone
+from enum import StrEnum
+from functools import cache
```
owtaylor, line 6: Hmm, what are your formatting settings? This seems different from what black wants to do.

supervisor/errata_utils.py
```diff
@@ -25,11 +26,12 @@ def ET_verify() -> bool | str:
         return True
 
 
-def ET_api_get(path: str):
+def ET_api_get(path: str, *, params: dict | None = None):
```
owtaylor, line 29: stray change for this patch, though there are a few places in jira_utils.py where using params would clean things up a bit.

supervisor/supervisor_types.py
```diff
@@ -47,6 +48,7 @@ class Erratum(BaseModel):
     synopsis: str
     status: ErrataStatus
     all_issues_release_pending: bool
+    last_status_transition_timestamp: datetime | None
```
owtaylor, line 51: So we can then remove the `| None` here.

owtaylor (changes requested):

supervisor/errata_utils.py
Jazzcort: Do you want me to differ this change with the jira_utils.py clean-up?

owtaylor: Sure, if you wanted to do a separate cleanup MR while we are thinking of it, probably a good idea.

PR #217, opened 2025-09-29
Mention permissions change to keytab in README

opohorel:
Specify 644 permissions needed for .secrets/keytab so the user in container can read the file. Prevents "Permission denied" errors during Kerberos authentication.


opohorel:
> Isn't that ensured by the `U` flag?
> 
> https://github.com/packit/ai-workflows/blob/cd50d0a2a9803e68efe8af30c7fc8170f2f9acb5/compose.yaml#L99

It seems like it should, but I've hit this issue and without manually changing the permissions of the keytab I wasn't able to make it readable inside the container

PR #223, opened 2025-09-30
Feat: Provide related erratum to the agent

Jazzcort:
Include the related erratum in the prompt so the agent has more context to make an informed decision—particularly in cases where test execution has not been initialized within 6 hours after the erratum moves to QE status.

### DIFF
--- a/supervisor/testing_analyst.py
+++ b/supervisor/testing_analyst.py
@@ -1,3 +1,4 @@
+from datetime import datetime, timezone
 import logging
 import os
 
@@ -10,7 +11,7 @@ from beeai_framework.template import PromptTemplate, PromptTemplateInput
 
 from agents.utils import get_agent_execution_config
 from .qe_data import get_qe_data, TestLocationInfo
-from .supervisor_types import FullIssue, TestingState
+from .supervisor_types import Erratum, FullIssue, TestingState
 from .tools.read_readme import ReadReadmeTool
 from .tools.search_resultsdb import SearchResultsdbTool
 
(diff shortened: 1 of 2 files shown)


Jazzcort:
@owtaylor Not sure if this is what you had in mind—happy to adjust if you’d like something different!

supervisor/testing_analyst.py
```diff
@@ -59,13 +64,18 @@ def render_prompt(input: InputSchema) -> str:
       If the tests are currently running:
          state: tests-running
          comment: [Provide a brief description of what tests are running and where the results will be]
+
+      If the erratum's status is QE and there are no comments from NEWA stating it has initiated test execution within 6 hours after the erratum's last status transition timestamp:
```
owtaylor, line 68: This is specific to only some components - not even all components that use NEWA - so let's put it into the QE information database instead of hardcoding here. Any text here should be more generic.

s/If the tests need to be started manually/If the tests need to be started manually or if they are supposed to be started manually but too much time has passed/ - something like that.

supervisor/testing_analyst.py
```diff
@@ -36,6 +39,8 @@ def render_prompt(input: InputSchema) -> str:
 
       JIRA_ISSUE_DATA: {{ issue }}
       TEST_LOCATION_INFO: {{ test_location_info }}
+      ERRATUM_DATA: {{ erratum }}
+      CURRENT_TIME: {{ current_time }}
```
owtaylor, line 43: I'd order it as Jira/Erratum/Test Location Info/Current Time

owtaylor (commented):
This generally looks good - just some details. After we land this, we should do another OR  to substitute Erratum here with a new FullErratum which has comments (like FullIssue) - [FullErrratum can use the same comment class as FullIssue]

supervisor/testing_analyst.py
Jazzcort: Will do!

PR #227, opened 2025-10-01
Remove unused mcp-atlassian container and references to it

opohorel:
`mcp-atlassian` is not being used anymore, it has been replaced by `mcp-gateway`

When talking with Tomáš Tomeček, he mentioned that the mcp-atlassian is not used anymore and we can remove it. That's exactly what is this PR for. Locally everything works as expected, needs review from OpenShift side of things.

The only mention which wasn't removed is `mcp-atlassian-secret` in OpenShift `cronjob-jira-issue-fetcher.yml` and `deployment-mcp-gateway.yml`. If I were to rename it, we would also need to adjust the secrets in the OpenShift .env

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

RELEASE NOTES END


TomasTomecek (commented):
the openshift part is definitely LGTM

@halfline @owtaylor @martinky82 can we drop the compose part?

PR #228, opened 2025-10-01
Fix serialization of sets

nforro:
This should fix the broken instrumentation with rebase agent, where the serialization is used only for logging, however I'm not sure if it won't break something elsewhere (in BeeAI).


nforro:
It seems to work fine locally, merging. I'll open an upstream PR in a moment.

PR #232, opened 2025-10-01
Remove the no-longer-needed GITLAB_USER env var

owtaylor:
We're no longer referencing this anywhere in the code for the agents.


TomasTomecek (approved):
Agreed, LGTM, thanks for the cleanup!

PR #238, opened 2025-10-03
drop `-3`, fix apply_patch_finish when `--skip` yields more conflicts

TomasTomecek:
I suggest going commit by commit in review.

This mainly fixes the apply_patch_finish logic and drops `-3` because it confuses agents with errors saying there are no parent commits

### DIFF
--- a/agents/backport_agent.py
+++ b/agents/backport_agent.py
@@ -86,7 +86,8 @@ def get_instructions() -> str:
          Use `rpmlint <PACKAGE>.spec` to validate your changes and fix any new issues.
 
       6. Run `centpkg --name=<PACKAGE> --namespace=rpms --release=<DIST_GIT_BRANCH> prep` to see if the new patch
-         applies cleanly.
+         applies cleanly. When `prep` command finishes with "exit 0", it's a success. Ignore errors from
+         libtoolize that warn about newer files: "use '--force' to overwrite".
 
       7. Generate a SRPM using `centpkg --name=<PACKAGE> --namespace=rpms --release=<DIST_GIT_BRANCH> srpm`.
 
@@ -97,8 +98,10 @@ def get_instructions() -> str:
       - Never change anything in the spec file changelog.
       - Preserve existing formatting and style conventions in spec files and patch headers.
       - Prefer native tools, if available, the `run_shell_command` tool should be the last resort.
-      - When resolving conflicts, ignore all changes in .github/ workflows and .gitignore.
+      - Ignore all changes in .github/ workflows, .gitignore, news, changes that cause conflicts.
       - Never apply the patches yourself, always use the `git_patch_apply` tool.
+      - Never run `git am --skip`, always use the `git_patch_apply` tool.
+      - Never abort the existing git am session.
     """
 
 
--- a/agents/backport_agent.py
+++ b/agents/backport_agent.py
@@ -493,6 +493,9 @@ async def main() -> None:
 
 if __name__ == "__main__":
     try:
+        # uncomment for debugging
+        # from utils import set_litellm_debug
+        # set_litellm_debug()
         asyncio.run(main())
     except FrameworkError as e:
         traceback.print_exc()
--- a/agents/tools/wicked_git.py
+++ b/agents/tools/wicked_git.py
@@ -132,7 +132,7 @@ class GitPatchApplyTool(Tool[GitPatchApplyToolInput, ToolRunOptions, StringToolO
     ) -> StringToolOutput:
         ensure_git_repository(tool_input.repository_path)
         try:
-            cmd = ["git", "am", "-3", "--reject", str(tool_input.patch_file_path)]
+            cmd = ["git", "am", "--reject", str(tool_input.patch_file_path)]
             exit_code, stdout, stderr = await run_subprocess(cmd, cwd=tool_input.repository_path)
             if exit_code != 0:
                 return StringToolOutput(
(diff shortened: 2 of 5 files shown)


agents/utils.py
```diff
@@ -127,6 +130,8 @@ async def mcp_tools(
     async with sse_client(sse_url) as (read, write), ClientSession(read, write) as session:
         await session.initialize()
         tools = await MCPTool.from_client(session)
+        logger.info(f"Available all tools: {tools}")
         if filter:
             tools = [t for t in tools if filter(t.name)]
+        logger.info(f"Available tools after filtering: {tools}, filter: {filter}")
```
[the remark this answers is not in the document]
TomasTomecek: I'll drop this commit tbh

agents/backport_agent.py
```diff
@@ -490,6 +493,17 @@ async def retry(task, error):
...
+            level=logging.DEBUG,
+            format="%(asctime)s %(levelname)s %(name)s %(message)s",
+            stream=sys.stdout,
+        )
+        # the following two modules call `litellm_debug(False)` on import
+        # import them explicitly now to ensure our call to `litellm_debug()` is not negated later
+        # import beeai_framework.adapters.litellm.chat
+        # import beeai_framework.adapters.litellm.embedding
+        # from beeai_framework.adapters.litellm.utils import litellm_debug
+        # litellm_debug(True)
```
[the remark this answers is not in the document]
nforro: Also, is the logging config necessary? I've never needed it.

compose.yaml
```diff
@@ -25,6 +25,8 @@ x-beeai-agent: &beeai-agent
     - .secrets/rhel-config.json:/home/beeai/rhel-config.json:ro,z,U
     - .secrets/jotnar-vertex-dev.json:/home/beeai/jotnar-vertex-dev.json:ro,z,U
   restart: unless-stopped
+  tty: true
+  stdin_open: true
```
nforro, line 29: Debugging always worked for me without this.

common/validators.py
```diff
@@ -20,9 +20,11 @@ def is_not_empty_string(value: str) -> str:
 NonEmptyString = Annotated[str, AfterValidator(is_not_empty_string)]
 
 
-def is_list_of_two_ints(value: list[int]) -> list[int]:
+def is_list_of_two_ints_or_empty(value: list[int]) -> list[int]:
+    if len(value) == 0:
+        return value
```
nforro, line 25: Are you saying Gemini doesn't recover from the error and doesn't retry? That was the whole point of this validator, if it doesn't work then we can just revert the implementation.

agents/tools/wicked_git.py
```diff
@@ -116,6 +118,42 @@ class GitPatchApplyToolInput(BaseModel):
...
+            # use case 1: -p is indeed wrong; this means we'll get a ton of 'No such file or directory' errors
+            no_such_count = len(re_no_such.findall(safe_stderr))
+            lines = safe_stderr.splitlines()
+            line_count = len(lines)
+            # if most of the lines are 'No such', let's continue with the next candidate
+            # because it truly means the patch doesn't fit
+            # the magic number is truly tricky, for example for the 19 commit expat cve, we get
+            # 1 here, because every single line is 'No such', but once the -p is correct, it's 0.2
+            if line_count > 0 and (no_such_count / line_count) >= 0.7:
+                continue
```
nforro, line 147: 😅

agents/tools/wicked_git.py
```diff
@@ -116,6 +118,42 @@ class GitPatchApplyToolInput(BaseModel):
     patch_file_path: AbsolutePath = Field(description="Absolute path to the patch file to apply")
 
 
+async def discover_patch_p(patch_file_path: AbsolutePath, repository_path: AbsolutePath) -> int:
```
majamassarini, line 121: Another way could be extract the name of a file in the patch, find it in the repository and compare how deep is the file in the directory and in the patch. But I don't know if it is worth changing it.

I asked claude to refine my thought and this is the output:

Algorithm

  1. Extract a file path from the patch (preferably from +++ line)
  2. Search for that file in the repository using find or git ls-files
  3. Count directory depth in both:
    - Patch: count / characters in the path
    - Repo: count / characters in the found path
  4. Calculate -p value: depth_in_patch - depth_in_repo

agents/backport_agent.py
```diff
@@ -86,19 +86,22 @@ def get_instructions() -> str:
...
+         libtoolize that warn about newer files: "use '--force' to overwrite".
 
       7. Generate a SRPM using `centpkg --name=<PACKAGE> --namespace=rpms --release=<DIST_GIT_BRANCH> srpm`.
 
 
       General instructions:
 
       - If necessary, you can run `git checkout -- <FILE>` to revert any changes done to <FILE>.
-      - Never change anything in the spec file changelog.
+      - Never change anything in the existing spec file changelog.
```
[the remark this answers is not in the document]
TomasTomecek: I wanted to be concrete by 5 more percent :)

agents/tools/wicked_git.py
TomasTomecek: That is a good proposal and I had a similar idea (Nikola suggested it as well) but I don't like parsing patches.

I can give it a try and we can compare I guess

compose.yaml
TomasTomecek: I dropped the commit

common/validators.py
TomasTomecek: > I would rather drop the argument and add another tool for partial viewing.

completely agree

agents/backport_agent.py
```diff
@@ -86,7 +86,8 @@ def get_instructions() -> str:
          Use `rpmlint <PACKAGE>.spec` to validate your changes and fix any new issues.
 
       6. Run `centpkg --name=<PACKAGE> --namespace=rpms --release=<DIST_GIT_BRANCH> prep` to see if the new patch
-         applies cleanly.
+         applies cleanly. When `prep` command finishes with "exit 0", it's a success. Ignore errors from
+         libtoolize that warn about newer files: "use '--force' to overwrite".
```
nforro, line 90: autotools/libtool are always PITA, I'm sure we'll encounter more similar issues.

PR #242, opened 2025-10-06
Try to improve tools Gemini is struggling with

nforro:
Inspired by https://github.com/google-gemini/gemini-cli/blob/main/docs/tools/file-system.md.


TomasTomecek (approved):
LGTM

PR #250, opened 2025-10-08
Implement git cherry-pick workflow for backport agent

opohorel:
This PR adds cherry-pick workflow into the backport agent.
I added a bunch of upstream tools which support the cherry-pick workflow, and expanded the specfile tools to extract the version of the package we are working on. I also added the upstream fix URL into the models.py, so the backport agent doesn't only get the local copy of the patch file, but also the URL for the patch to then extract the upstream repository from it.

Of course, there are a bunch of new instructions for the agent, which still need a lot of testing.

There is still a plenty of room for improvement, like looking for the upstream commit which corresponds with the version we are working on downstream. Right now, we are looking only for tags which are formatted in some common patterns. For example, expat release tags are not supported right now. There is still git am workflow in place so we can fall back to it when any of the steps fail.

Fixes: packit/jotnar#181

### 1. Initial Steps
```
1. Check if fix already exists
↓
git_log_search tool
↓
If found: success=True, status="Backport already applied"
↓
If not found: continue to step 2

2. Prepare package sources
↓
git_prepare_package_sources tool
↓
Creates: {{unpacked_sources}} directory

3. DECISION POINT: Which workflow?
↓
Is {{upstream_fix}} a URL?
↓
YES → Cherry-pick Workflow (A)
NO → Git AM Workflow (B)
```
### 2. Cherry-pick Workflow (A) - Preferred
```
# If any of these steps fail, we are reverting back to git am workflow

3a. Extract upstream repository info
↓
extract_upstream_repository tool
↓
Get: repo_url, commit_hash
3b. Get package info from dist-git
↓
get_package_info tool
↓
Get: version, patch_files list
3c. Clone upstream repository
↓
clone_upstream_repository tool
↓
Creates: {{local_clone}}-upstream
3d. Find and checkout base version
↓
find_base_commit tool
↓
Checkout matching tag in upstream repo
3e. Apply existing patches to upstream
↓
apply_downstream_patches tool
↓
Recreate package state in upstream repo
3f. Cherry-pick the new fix
↓
cherry_pick_commit tool
↓
If conflicts: resolve manually, then cherry_pick_continue
3g. Generate final patch
↓
generate_patch_from_commit tool
↓
Create clean patch file in {{local_clone}}
3h. Workflow complete!
↓
Continue to steps 4-6
```
### 3. Git AM Workflow (B) - Fallback
```
3a. Apply patch to sources
↓
git_patch_apply tool
↓
Apply {{local_clone}}/{{jira_issue}}.patch
3b. Create final patch
↓
git_patch_create tool
↓
Update patch file
↓
Continue to step 4
```

### Final steps (both workflows)
```
Update spec file
↓
update_release tool
↓
Add new Patch tag, run rpmlint
Test patch application
↓
centpkg prep
↓
Verify exit 0 (success)
Build SRPM
↓
centpkg srpm
↓
Generate final SRPM
```

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

RELEASE NOTES END


agents/tools/upstream_tools.py
```diff
@@ -0,0 +1,702 @@
...
+            pr_match = re.search(r'/pull/(\d+)(?:\.patch)?', parsed.path)
+            mr_match = re.search(r'/merge_requests/(\d+)(?:\.patch)?', parsed.path)
+
+            if pr_match or mr_match:
+                # Handle GitHub Pull Request or GitLab Merge Request
+                pr_number = pr_match.group(1) if pr_match else mr_match.group(1)
+
+                # Extract owner/repo from path
+                if pr_match:
+                    repo_path_match = re.match(r'/([\w\-\.]+)/([\w\-\.]+)/pull/', parsed.path)
```
nforro: Why repeated matching? Wouldn't a single match be enough?

agents/backport_agent.py
```diff
@@ -78,34 +88,157 @@ def get_instructions() -> str:
 
       To backport upstream fix <UPSTREAM_FIX> to package <PACKAGE> in dist-git branch <DIST_GIT_BRANCH>, do the following:
 
+      CRITICAL: Do NOT modify, delete, or touch any existing patches in the dist-git repository.
+      Only add new patches for the current backport. Existing patches are there for a reason
+      and must remain unchanged.
```
TomasTomecek: 💯

TomasTomecek (approved):
LGTM, would love to see this run on multiple issues

agents/tools/upstream_tools.py
```diff
@@ -0,0 +1,702 @@
...
+                    # GitHub API
+                    api_url = f"https://api.github.com/repos/{owner}/{repo}/pulls/{pr_number}"
+                else:
+                    # GitLab API
+                    api_url = f"https://{parsed.netloc}/api/v4/projects/{owner}%2F{repo}/merge_requests/{pr_number}"
+
+                try:
+                    req = Request(api_url)
+                    req.add_header('Accept', 'application/json')
+                    req.add_header('User-Agent', 'RHEL-Backport-Agent')
```
lbarcziova: for the API stuff, we could (probably later) use our https://github.com/packit/ogr library to simplify few things here

agents/backport_agent.py
```diff
@@ -118,6 +251,10 @@ def get_instructions() -> str:
       - Apply all changes that modify the core library of the package, and all binaries, manpages, and user-facing documentation.
       - For more information how the package is being built, inspect the RPM spec file and read sections `%prep` and `%build`.
       - If there is a complex conflict, you are required to properly resolve it by applying the core functionality of the proposed patch.
+      - When a tool explicitly says "Abort cherry-pick approach, use git am workflow", immediately switch to approach B.
```
lbarcziova: I guess here is the response to my question, it is a bit confusing though 😅 how often has this happened when testing locally, and was LLM able to follow this reliably?

lbarcziova (commented):
impressive work, thanks a lot! I would love to see this working on more examples to see how it performs in comparison to `git am` workflow, do you think it is ready to be tried directly in prod or we should attempt more example runs locally?

agents/tools/upstream_tools.py
nforro: I mean, it's not technically wrong, I just think it could be done more effectively (and potentially improving readability).

opohorel:
> > Second build agent is there because when using git am workflow the agent is on rebuild told to start over and reset everything. This is not desirable with cherry-picking, as we want the agent to keep reiterating on the upstream sources to fix the build failures.
> 
> This just doesn't add up. You are saying the agent can't start from scratch, yet you are creating a new instance with blank memory (hence no context from the previous attempt), only feeding it the current state. Also, the prompt should not contain generic instructions, it should be short and take advantage of templating. Almost the entire new prompt should be part of instructions instead. The istructions should be probably more structured, having sections for both workflows and initial and repeated process for cherry-picking.
> 
> > I also wanted to keep the git am workflow as is, so I implemented the second instance of build agent
> 
> That just makes the implementation look like a hack.

That's a valid concern. I agree that the design could be cleaner. The instructions should be more structured, and the separate agent instance is a bit awkward.
However, the current implementation solves the problem. Iterative build fixing for cherry-picked backports now works reliably. The cherry-pick workflow can recover from build failures without starting the backport again.

Would you be open to merging this PR and opening related issues where I could work further on improving the workflow? I feel like soon this PR is going to be too big to rebase if any work on backporting agent is going to be done. Also, it would be nice to have feedback from running the agent in production as I have limited time to test it locally

nforro:
> Would you be open to merging this PR and opening related issues where I could work further on improving the workflow? I feel like soon this PR is going to be too big to rebase if any work on backporting agent is going to be done. Also, it would be nice to have feedback from running the agent in production as I have limited time to test it locally

Definitely, let's do that.

PR #254, opened 2025-10-08
supervisor/jira_utils.py: Don't decode response unless needed

owtaylor:
Not all API calls to JIRA return a JSON response - some (like transitioning an issue to a new state) simply return an empty response, so add a parameter to control whether whether we attempt to decode the response as JSON.


Jazzcort:
LGTM

PR #316, opened 2025-10-28
Collect clones and postpone triaging of y streams for CVEs

majamassarini:


mcp_server/jira_tools.py
```diff
@@ -197,34 +197,42 @@ async def check_cve_triage_eligibility(
...
+    # Process Z-stream CVEs (postpone Y-stream CVEs)
     if re.match(r"^rhel-\d+\.\d+$", target_version.lower()):
         return CVEEligibilityResult(
             is_cve=True,
-            is_eligible_for_triage=False,
+            when_eligible_for_triage=WhenEligibility.LATER,
             reason="Y-stream CVEs will be handled in Z-stream"
         )
 
+    # Process maintenance streams CVEs (X.10.z) after all the z-streams are processed
```
lbarcziova: this will be so helpful to not have to check manually, nice and clean!

mcp_server/jira_tools.py
```diff
@@ -398,3 +406,40 @@ async def verify_issue_author(
...
+        except aiohttp.ClientError as e:
+            raise ToolError(f"Failed to get Jira data: {e}") from e
+
+    # If the issue is not assigned to the Jötnar project,
+    # don't wait for it
+    # @TODO: is this ok??? for example https://issues.redhat.com/browse/RHEL-110692
+    # will wait on https://issues.redhat.com/browse/RHEL-110689
+    # without this check.
+    if current_assignee.lower() != "jötnar project":
+        return True
```
lbarcziova: I think we should always wait, but we can discuss this more. I would also assume this shouldn't happen most of the time? And in the given example, the second issue is not z-stream one, so it wouldn't wait, or am I missing something?

lbarcziova (commented):
this will be amazing, thanks Maja!

agents/clones_analyzer_agent.py
```diff
@@ -0,0 +1,142 @@
...
+from tools.commands import RunShellCommandTool
+from tools.version_mapper import VersionMapperTool
+from common.models import ClonesInputSchema, ClonesOutputSchema
+from utils import get_chat_model, get_tool_call_checker_config
+from utils import mcp_tools, get_agent_execution_config
+
+logger = logging.getLogger(__name__)
+
+
+def get_instructions() -> str:
```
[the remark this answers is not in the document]
majamassarini: I would say it is reliable. I don't know if I can write a regexp for it. Because I don't know what to expect in the title. So probably this is safer. And in "this agent" there is still a missing step (one of the reason why this PR is still in draft) I should start suggesting if any clone is missing.

lbarcziova:
Going to close this as we already have #417 and followup for periodic checking in Jira.

PR #369, opened 2025-11-24
Run a reproducer in a simple way

majamassarini:


agents/build_agent.py
```diff
@@ -35,6 +35,7 @@ def get_instructions() -> str:
       just return the error message. Otherwise, start with `builder-live.log` and try to identify
       the build failure. If not found, try the same with `root.log`. Summarize the findings
       and return them as `error`. If the build failed due to a build timeout, set `is_timeout` to `true` in your output.
+      If the build succeeded, return the URL to the built package in `url`.
```
nforro: This makes no sense. First, URLs to built artifacts are already returned from the `build_package` tool, second, there could be multiple RPMs (even hundreds in some cases) - which one should the model choose? The correct answer is all of them, but that's not always possible, they could conflict with each other for example. It would be best to leave this to Testing Farm and pass a Copr build ID instead of a RPM.

agents/triage_agent.py
```diff
@@ -253,7 +253,22 @@ def render_prompt(input: InputSchema) -> str:
            is clarification-needed
          * This is the correct choice when you are sure a problem exists but cannot find the solution yourself
 
-         2.5 Set the Jira fields as per the instructions below.
+         2.5. If the chosen action is backport then search for the issue reproducer in the issue comments.
```
nforro: I suggest using `reproducer test case` instead of `reproducer`, everywhere, not just here. Reproducer test case succeeds when the reproducer fails - the issue didn't manifest.

mcp_server/testing_farm_tools.py
```diff
@@ -0,0 +1,54 @@
...
+    ]
+
+    logger.info(f"Running testing-farm command: {' '.join(shlex.quote(arg) for arg in cmd)}")
+
+    try:
+        process = await asyncio.create_subprocess_exec(
+            *cmd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+        )
```
nforro: I would rather send a request using `aiohttp`.

agents/backport_agent.py
```diff
@@ -288,6 +289,12 @@ def get_prompt() -> str:
       Everything from the previous attempt has been reset. Start over, follow the instructions from the start
       and don't forget to fix the issue.
       {{/build_error}}
+      {{#reproducer_failed}}
+      The reproducer failed after the first backport attempt.
```
nforro: Like I said elsewhere, this is confusing. We want the reproducer to fail - if it fails, the fix worked. But only if it didn't fail before applying the fix, if it did, it's a bogus reproducer. I would use the term `reproducer test case`, and run it twice, first without the new build, expected to fail, and then with the new build, expected to succeed.

agents/backport_agent.py
```diff
@@ -288,6 +289,12 @@ def get_prompt() -> str:
...
+      Keep it in mind and try to fix the patches you apply to make the reproducer pass.
+      You can inspect the reproducer test at {{tmt_reproducer.git_url}}/blob/{{tmt_reproducer.git_ref}}/{{tmt_reproducer.test}}.
+      The test is located in the {{tmt_reproducer.git_ref}} branch of the {{tmt_reproducer.git_url}} repository.
```
nforro: Isn't this line redundant? All the info is on the line above. Though I think we should come up with a convention how to reference test cases, for example, inspired by pip and similar (I think we even use something like that in packit):
`https://gitlab.com/redhat/rhel/tests/expat.git@master#testcase=Security/RHEL-114639-CVE-2025-59375-expat-libexpat-in-Expat-allows`

lbarcziova:
Closing this as it is now outdated.

PR #389, opened 2026-03-17
Add support for backporting to RHEL z-streams

antbob:
This commit adds the ability to triage and backport for RHEL z-streams, specifically older z-streams. The main changes are:

- Instructions are now split up into 2 different sets for z-stream and regular cases. This is perhaps best addressed long term with some sort of refactor into a template system and storing the instructions separate from the code. The instructions are mostly the same for now but we will be changing z-stream instructions further in the future.

- There is now a distinction between centpkg and rhpkg and download and prep sources are invoked via MCP gateway instead of directly as rhpkg requires authentication even for basic functionality. This also required a change to how the package build directory is located as there are changes between MCP gateway rpm version and agents rpm version when preparing package sources, resulting in different directory structure.

- The spec file is now the source of truth for patch files. Instead of tracking what has been added vs what is needed which can get inconsistent, as the agent might create or add random patch files in the process, we instead just track what we need from the spec file and ignore anything else.

Please note that I have only tested these changes on older RHEL z-stream backports and while I've tried not to disturb the existing code and workflow the best I can, there might be some hidden corner cases or inconsistencies related to the standard (non z-stream) process.


TomasTomecek (dismissed):
Nicely done! I went through the code and haven't seen any serious problems.

**Edit: Claude is wrong.** I asked it to analyze again with opus 4.6:

> My earlier review was wrong to flag this as a critical issue. It's the opposite — the old behavior was the bug.

--

Original comment: This is what Claude found locally:
```
  1. _map_version_to_branch now returns a non-existent branch instead of falling back to CentOS Stream

  This is the most significant behavioral change and the one most likely to cause real-world breakage.

  Before (triage_agent.py, old lines 113-133):
  - For z-stream versions, if the internal branch (rhel-9.7.0) did not exist, it fell back to CentOS Stream (c9s).
  - If package was None, the z-stream block was skipped entirely, also falling back to c9s.

  After (triage_agent.py:119-139):
  - For z-stream (or older z-stream) versions, it always returns the internal branch name regardless of whether it exists.
  - The branch existence check is now just a log message (line 133-134) — it logs a warning but still returns the branch.
  - If package is None, it skips the existence check entirely and still returns the branch.

  Why this could break: The returned branch name is used directly in:
  - RedisQueues.get_backport_queue_for_branch(state.target_branch) (line 792) — this works fine, it just routes to c9s or c10s queue based on pattern matching.
  - But the branch name also flows into the backport agent via state.dist_git_branch, where it's used in actual git operations like centpkg/rhpkg prep --release=<branch>. If the branch doesn't exist in the dist-git repo, those operations will fail.

  The comment says "it will be created later if needed" (line 120), but I found no code in this commit that creates the branch. This is an intentional design change, but if the downstream branch-creation logic doesn't exist yet, backport jobs targeting
  non-existent branches will fail at runtime.
```

Edit: all good, +1

nforro (approved):
I don't like the prompt duplication, especially in the backport agent it adds up and makes tweaking prompts/instructions a PITA. But fixing that is out of scope of this PR.

nforro:
> @nforro after a few iterations the latest push has the best i can do. the reason for all the gymnastics there is we cant reliably parse specfile with existing tools on Fedora as old RHEL spec files with obsolete features like patchN cause the parser to fail so we have to extract and parse what we need by ourselves.

To be honest, I don't like this at all. I mean, it's a clever way to avoid parsing, but I'm afraid the implementation is fragile and won't work for every spec file, and although we can make fixes along the way, I would prefer to use the specfile library for this, once https://github.com/packit/specfile/issues/434 is implemented. BTW, the differences between RPM versions are the reason for the c9s and c10s flavors of agent containers.

agents/backport_agent.py
```diff
@@ -524,19 +855,39 @@ def get_unpacked_sources(local_clone: Path, package: str) -> Path:
...
+    name = s.expand("%{name}")
+    version = s.expand("%{version}")
+
+    setup = None
+    with s.prep() as prep:
+        if prep is not None:
+            for macro in prep.macros:
+                if isinstance(macro, (SetupMacro, AutosetupMacro)):
+                    setup = macro
+                    break
```
nforro: There are specs that use both (conditionally), or multiple instances of one (and only some with `-n`), so this approach will fail sooner or later. But I suppose we can improve this when that happens.

nforro (dismissed):
LGTM, thanks!

agents/backport_agent.py
antbob: addressed in the latest revision

nforro (approved):

PR #407, opened 2026-04-14
Update custom Jira fields to the new Jira Cloud instance ids

antbob:
After migration to the new Jira Cloud instance there is been a change to custom Jira fields ids which this commit updates accordingly. There is also better error reporting if there are any further changes to those custom fields in the future to make troubleshooting easier.


lbarcziova (approved):
thanks a lot!

PR #409, opened 2026-04-15
Rebuild phoenix and redis-commander images from source to patch SQLite

antbob:
Both images are being rebuilt from source to patch CVE-2025-6965 — a vulnerability in the bundled SQLite library. The upstream images ship old SQLite (3.40.1 for Phoenix, 3.48.0 for Redis Commander); the fix compiles SQLite 3.50.2 from source and overlays it.


antbob:
this also adds cryptography, Node.js and OpenSSL.

TomasTomecek:
I checrrypicked your commit and finished this in https://github.com/packit/ai-workflows/pull/409

the main difference is that the images are not built from source, instead latest release is picked up from npm or pypi

PR #411, opened 2026-04-15
Backport deduplication

opohorel:
Summary
- Unified `GitPatchCreationTool` and `GeneratePatchFromCommitTool` into a single tool (`git_patch_create`) that uses `git format-patch --stdout` to produce one combined patch
- `ApplyDownstreamPatchesTool` now stores `base_head_commit` in `self.options` automatically, removing the need for the LLM to manually track commit hashes with `git rev-parse HEAD`
- Both cherry-pick and git-am workflows now produce a single `{JIRA}.patch` file and add one `Patch:` tag in the spec, instead of per-URL numbered patches

Internal tests pass, but I wasn't able to verify the function as other bits of code are now failing.

This is a first part of my plan to unify the backporting agent workflows and making it less LLM heavy


ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -479,8 +485,15 @@ async def _run(
 
                 applied_patches.append(patch_file)
 
+            cmd = ["git", "rev-parse", "HEAD"]
+            exit_code, stdout, stderr = await run_subprocess(cmd, cwd=tool_input.repo_path)
+            if exit_code != 0:
+                raise ToolError(f"Failed to get HEAD commit after applying patches: {stderr}")
+            self.options["base_head_commit"] = stdout.strip()
```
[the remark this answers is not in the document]
opohorel: I don't think we need a helper for this. The duplication is only 3 lines (git rev-parse + store + error check) across two code paths that return different messages. IMO not worth it

ymir/tools/unprivileged/wicked_git.py
```diff
@@ -313,18 +316,24 @@ async def _run(
...
             ]
-            exit_code, _, stderr = await run_subprocess(cmd, cwd=tool_input.repository_path)
+            exit_code, stdout, stderr = await run_subprocess(cmd, cwd=tool_input.repository_path)
             if exit_code != 0:
                 raise ToolError(f"Command git-format-patch failed: {stderr}")
-            return StringToolOutput(result=f"Successfully created a patch file: {tool_input.patch_file_path}")
+            if not stdout:
+                raise ToolError("Generated patch is empty")
+            with open(tool_input.patch_file_path, 'w', encoding='utf-8') as f:
+                f.write(stdout)
```
nforro:
```suggestion
            tool_input.patch_file_path.write_text(stdout, encoding="utf-8")
```

nforro (dismissed):
LGTM

ymir/tools/unprivileged/wicked_git.py
opohorel: Thanks for the suggestion, I've added it. Could you please approve again? Thank you!

nforro (approved):

PR #412, opened 2026-04-15
Multiple fixes 

lbarcziova:
Some related to tools refactor done recently.

### DIFF
--- a/ymir/tools/privileged/gitlab_tools.py
+++ b/ymir/tools/privileged/gitlab_tools.py
@@ -276,17 +276,6 @@ class OpenMergeRequestTool(Tool[OpenMergeRequestToolInput, ToolRunOptions, JSONT
         if not pr:
             raise ToolError("Failed to open the merge request")
 
-        for attempt in range(5):
-            try:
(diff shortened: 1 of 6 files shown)


ymir/tools/privileged/gitlab_tools.py
```diff
@@ -243,7 +244,7 @@ async def _run(
         tool_input: OpenMergeRequestToolInput,
         options: ToolRunOptions | None,
         context: RunContext,
-    ) -> JSONToolOutput[Tuple[str, bool]]:
+    ) -> JSONToolOutput[list]:
```
nforro: A better fix (without losing the types) would be to define a Pydantic model with `str` and `bool` members.

nforro: It could even have a method to convert the content to a tuple.

lbarcziova: added a commit, can you have a look if this is what you meant please? Haven't e2e tested this yet though as I don't have more PRs to open.

nforro: Yes, thanks!

nforro (approved):

PR #414, opened 2026-04-16
Converting prompt

TomasKorbar:
- Fix unprivileged gateway
- Add skills
- Make BeeAI required by subpackages the same as we use in images
- Add logfile option to gateways they can be easily debugged when used in Claude Code

### DIFF
--- a/agents_as_skills/backport/SKILL.md
+++ b/agents_as_skills/backport/SKILL.md
@@ -1,0 +1,955 @@
+---
+description: Backport upstream patches to packages in the RHEL ecosystem — cherry-pick or git-am workflow, build verification, changelog, and merge request creation.
+arguments:
+  - name: package
+    description: "Name of the package to update (e.g., 'openssl')"
+    required: true
+  - name: dist_git_branch
+    description: "Dist-git branch to update (e.g., 'c10s', 'rhel-9.6.0')"
+    required: true
+  - name: upstream_patches
+    description: "Comma-separated list of upstream patch/commit/PR URLs to backport"
+    required: true
+  - name: jira_issue
+    description: "JIRA issue key (e.g., RHEL-12345)"
+    required: true
+  - name: cve_id
+    description: "CVE identifier if the JIRA issue is a CVE (e.g., CVE-2025-12345). Default: null"
+    required: false
+  - name: dry_run
+    description: "If true, skip JIRA status changes, MR creation, and label updates. Default: false"
+    required: false
+  - name: max_build_attempts
+    description: "Maximum number of build retry attempts. Default: 10"
+    required: false
+  - name: max_incremental_fix_attempts
+    description: "Maximum number of incremental cherry-pick fix attempts when build fails. Default: same as max_build_attempts"
+    required: false
+---
+
+# Backport Skill
+
+You are a Red Hat Enterprise Linux developer performing an end-to-end backport of upstream patches to a dist-git package.
+
+## Input Arguments
+
+- `package`: {{package}}
+- `dist_git_branch`: {{dist_git_branch}}
+- `upstream_patches`: {{upstream_patches}} (comma-separated URLs)
+- `jira_issue`: {{jira_issue}}
+- `cve_id`: {{cve_id}}
+- `dry_run`: {{dry_run}}
+- `max_build_attempts`: {{max_build_attempts}}
+- `max_incremental_fix_attempts`: {{max_incremental_fix_attempts}}
+
+## Tools
+
+This skill uses the following tools. Do not restrict tool usage — use any tool available as needed.
+
+**MCP Tools (called via MCP gateway):**
+- `get_patch_from_url` — Fetch and validate patch content from a URL
+- `build_package` — Build an SRPM and return results (used during build and incremental fix steps)
+- `download_artifacts` — Download build log artifacts (*.log.gz)
+- `change_jira_issue_status` — Change the status of a JIRA issue
+- `add_jira_comment` — Post a comment to a JIRA issue
+- `fork_dist_git_repo` — Fork a dist-git repository and prepare a working branch
+- `open_merge_request` — Open a merge request against dist-git
+- `add_blocking_merge_request_comment` — Add a blocking comment to a merge request
+- `create_merge_request_checklist` — Create a QA review checklist on a merge request
+- `add_merge_request_labels` — Add labels to a merge request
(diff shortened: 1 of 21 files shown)


agents_as_skills/backport/SKILL.md
```diff
@@ -0,0 +1,955 @@
...
+
+**Other:**
+- Web search via DuckDuckGo or equivalent
+- Bash tool for shell commands (e.g., `git`, `centpkg`, `rhpkg`, `curl`, `wc`)
+
+## Workflow
+
+Execute the following steps in order. Track state across steps (paths, flags, results).
+
+Determine `pkg_tool` from the branch: if `dist_git_branch` starts with `c` and ends with `s` (e.g., `c10s`, `c9s`), use `centpkg`; otherwise use `rhpkg --offline --released`.
```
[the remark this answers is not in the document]
TomasKorbar: That is not correct here. Let agent obey the current instructions instead.

ymir/tools/gateway_utils.py
```diff
@@ -0,0 +1,55 @@
...
+import re
+from typing import Any
+
+from beeai_framework.emitter.emitter import Emitter
+
+
+logger = logging.getLogger(__name__)
+
+# Patterns that match common credential formats in log output
+_REDACT_PATTERNS = [
```
jpodivin, line 13: Maybe we should keep a list of all the things we are redacting.

jpodivin (approved):

ymir/common/pyproject.toml
```diff
@@ -9,7 +9,7 @@ description = "Common utilities shared across Ymir AI workflows"
 requires-python = ">=3.13"
 dependencies = [
     "aiofiles>=24.1.0",
-    "beeai-framework[mcp]>=0.1.36",
+    "beeai-framework[mcp]==0.1.55",
```
nforro: We could use e.g. [hatch-requirements-txt](https://github.com/repo-helper/hatch-requirements-txt) plugin to enable reading dependencies from files and use one file for common dependencies and other files for package-specific ones. That way we wouldn't have to update the version caps at multiple places.

ymir/tools/gateway_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    for pattern in _REDACT_PATTERNS:
+        text = pattern.sub("[REDACTED]", text)
+    return text
+
+
+def setup_logging():
+    """Configure logging and attach Emitter listeners that log tool calls with redacted credentials."""
+    handlers = [logging.StreamHandler()]
+    if os.environ.get("DEBUG_FILE"):
+        handlers.append(logging.FileHandler(os.environ.get("DEBUG_FILE")))
```
nforro:
```suggestion
    if debug_file := os.environ.get("DEBUG_FILE"):
        handlers.append(logging.FileHandler(debug_file))
```

nforro:
```suggestion
    if debug_file := os.environ.get("DEBUG_FILE"):
        handlers.append(logging.FileHandler(debug_file))
```

nforro (dismissed):
Are the prompts in skill files the same as in the BeeAI agents? If yes, or if there are only minor differences, I think we should use some templating mechanism to render the skill files (unless they support includes) and to read the prompts from files in the agents.

ymir/tools/gateway_utils.py
nforro: Thanks GitHub, a way to hide the suggestion from me, forcing me to create another one, ending up with two :sweat_smile:

TomasKorbar:
@nforro 
> Are the prompts in skill files the same as in the BeeAI agents? If yes, or if there are only minor differences, I think we should use some templating mechanism to render the skill files (unless they support includes) and to read the prompts from files in the agents.

Not entirely the same. The issue here are the workflow steps, while they are perfect for doing work that can be described programmatically, they can not be easily converted into natural language, that is why i used the claude prompt for conversion.

requirements.txt
```diff
@@ -0,0 +1,21 @@
+# Dependencies specific to the root ymir package
+ymir-common
+ymir-tools
+litellm!=1.82.7,!=1.82.8
+aiohttp>=3.12.15
+arize-phoenix-otel>=0.13.0
+beautifulsoup4>=4.13.4
+beeai-framework[duckduckgo]
```
nforro, line 8: Shouldn't this rather be in the common requirements file as `beeai-framework[duckduckgo,mcp]==0.1.55`?

nforro (dismissed):
Nice, thanks!

I'm just wondering about the requirements files, wouldn't in make more sense to have the package-specific ones named simply `requirements.txt` and the shared one e.g. `requirements-common.txt` or `requirements-global.txt`? I don't quite understand the reasoning behind `requirements-tools-common.txt`.

TomasKorbar:
@nforro 
> Nice, thanks!
> 
> I'm just wondering about the requirements files, wouldn't in make more sense to have the package-specific ones named simply `requirements.txt` and the shared one e.g. `requirements-common.txt` or `requirements-global.txt`? I don't quite understand the reasoning behind `requirements-tools-common.txt`.

I was hesitant to put requirement.txt and requirements-global.txt both in the root directory. I mean it would confuse me so i wanted to make clear that one of them is for tools and common packages and the other one for the sum of entire project.

requirements.txt
TomasKorbar: common and tools subpackages do not use duckduckgo. That is required only for agents, thus the duckduckgo stayed here.

nforro:
> I was hesitant to put requirement.txt and requirements-global.txt both in the root directory. I mean it would confuse me so i wanted to make clear that one of them is for tools and common packages and the other one for the sum of entire project.

But ymir package still uses the common dependencies, doesn't it? I thought that was the goal. Anyway, another possibility is to maintain all requirements files inside `requirements` directory, where there can be arbitrary structure, e.g.:

```
ai-workflows
└── requirements
    ├── common.txt
    ├── tools
    │   └── common.txt
    └── ymir.txt
```
or
```
ai-workflows
└── requirements
    ├── common.txt
    ├── tools.txt
    └── ymir.txt
```

TomasKorbar:
Yeah the ymir package does use common dependencies. Hmm on the second thought you are right, i have been thinking about this the wrong way. Will switch the names to requirements.txt and requirements-global.txt.

nforro (approved):

PR #418, opened 2026-04-20
Deployment fixes and more: MCP retry logic, Kerberos

TomasTomecek:
## Summary

- **MCP gateway retry logic**: `mcp_tools()` now retries on connection errors instead of crashing immediately, handling both plain `ConnectError` and the `ExceptionGroup` wrapping anyio produces. Introduces `_is_connection_error()` helper and adds `httpx` as a dependency.
- **Kerberos KDC config**: adds `files/ipa_redhat_com` and copies it into `/etc/krb5.conf.d/` in the MCP container image so the IPA realm is resolvable inside the container.
- **`deploy.sh` rewrite**: replaces the naive `for filename in *.yml` glob with an explicit per-component listing. Adds `apply()` and `import_image()` helpers that always pass `-n jotnar-ymir--jotnar-ymir`, eliminating the risk of accidentally deploying to the wrong namespace.
- **GPC cluster adapters**: PVC storage classes changed to `netapp-nfs`, AWS-specific `nodeSelector` removed from all deployments, `DRY_RUN: "true"` set on agents and MCP gateway, `CHAT_MODEL` bumped to `claude-sonnet-4-6`, image references updated to quay.io.
- **Makefile cleanup**: drops Loki targets (not applicable to GPC), fixes `COMPOSE` detection to fall back to `podman-compose`, fixes `make stop`.
- **Drop unused agent Service objects**: removes `service-{triage,backport,rebase}-agent.yml` — agents don't expose a network service.
- **Tests**: converts `mcp_tools` retry tests from `unittest.mock` to `flexmock`, fixes the module path (`agents.utils` → `ymir.common.utils`).

## Notes

- Phoenix is commented out in `deploy.sh` — the image is private and the route admission webhook panics on this cluster (platform bug; route must be created manually).
- Redis Commander is also commented out for now.
- Image references currently point to `quay.io/tomastomecek/` — these should be updated to the canonical `quay.io/jotnar/` images before or shortly after merge.
- Phoenix URL in `monitoring.md` is a placeholder pending route creation.


agents/tests/unit/test_utils.py
```diff
@@ -0,0 +1 @@
+
```
nforro: This file is added in the second commit and then made empty by pre-commit.ci, can we just drop it?
BTW, as pre-commit.ci is kicking in, I'm afraid your pre-commit hooks are not configured correctly.

ymir/common/utils.py
```diff
@@ -315,16 +316,38 @@ async def run_tool(
...
+        try:
+            async with sse_client(sse_url) as (read, write), ClientSession(read, write) as session:
+                await session.initialize()
+                tools = await MCPTool.from_client(session)
+                if filter:
+                    tools = [t for t in tools if filter(t.name)]
+                connected = True
+                yield tools
+                return
+        except BaseException as e:
```
nforro: Is there a reason to use `BaseException`? `ExceptionGroup` is a subclass of `Exception`.

ymir/common/utils.py
```diff
@@ -315,16 +316,38 @@ async def run_tool(
...
-        await session.initialize()
-        tools = await MCPTool.from_client(session)
-        if filter:
-            tools = [t for t in tools if filter(t.name)]
-        yield tools
+    sse_url: str,
+    filter: Callable[[str], bool] | None = None,
+    max_retries: int = 10,
+    retry_delay: float = 3.0,
+) -> AsyncGenerator[list[MCPTool], None]:
```
nforro: I think we need to decide what Python version we actually want to target. With 3.13 (specified in pyproject.toml) the original `AsyncGenerator[list[MCPTool]]` is enough, but we do build the c9s images with Python 3.11 😅

ymir/common/utils.py
TomasTomecek: thanks! I haven't reviewed Claude's output in detail, my bad; will definitely change it to Exception (and will look into ExceptionGroup, first time seeing it)

openshift/deployment-backport-agent-c10s.yml
```diff
@@ -20,8 +20,6 @@ spec:
         app: backport-agent-c10s
         deployment: backport-agent-c10s
     spec:
-      nodeSelector:
-        kubernetes.io/hostname: ip-10-30-34-89.us-east-1.compute.internal
```
nforro, line 24: We were using these for a reason, are we sure the issue we were hitting is resolved?

TomasTomecek: I'll clarify in the commit message

(it was the limitation of PVs in the old cluster, the new PVs are NFS-backed and don't have this problem; and mainly, this IP is AWS while the new cluster is in private datacenter)

agents/tests/unit/test_utils.py
TomasTomecek: sorry, a relict from rebase, shouldn't have staged this file

TomasTomecek:
Thank you @nforro for spotting those mistakes, should have checked myself twice before pushing. I addressed all your comments.

nforro (approved):
There is still an extra sentence (referring to the first commit) in the second commit's commit message, otherwise LGTM.

lbarcziova:
great changes!

PR #420, opened 2026-04-23
Implement pre-push check for agent/skill consistency

TomasKorbar:
The check works as a gating. If the skill was not edited and agent was, then it prevents you from pushing.
This is because to generate the skill it takes some time and it would make development of agents painfull to do it on every commit.
Also the hooks do not allow to show conversion process in a sane way.
You can skip the check with environment variable SKIP_SKILL_SYNC

### DIFF
--- a/scripts/sync-agent-skills.py
+++ b/scripts/sync-agent-skills.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+"""Pre-push hook: ensure agents_as_skills/ stays in sync with ymir/agents/.
+
+Compares the net diff of the current branch against its remote tracking branch.
+Fails the push when a *_agent.py file was changed but its corresponding
+SKILL.md under agents_as_skills/ was not.  Prints the exact claude command
+the developer should run to regenerate the skill.
+
+Set SKIP_SKILL_SYNC=1 to bypass this check.
+"""
+
+import os
+import sys
+from pathlib import Path
+
+from git import Repo
+
+AGENTS_DIR = Path("ymir/agents")
+SKILLS_DIR = Path("agents_as_skills")
+
+CLAUDE_CMD_TEMPLATE = (
+    "claude --model claude-sonnet-4-6 --effort high"
+    " --allowedTools Read Glob Grep"
+    ' "Please take a look at the BeeAI workflows implemented in agents'
+    " directory. Please convert Workflow in {workflow_file} to Claude skill and"
+    f" save that skill to {SKILLS_DIR} directory.\n"
+    "Restrictions:\n"
+    " - Pay attention to tools used by the workflow and do not omit them\n"
+    " - Do not restrict tools that the skill can use\n"
+    ' - Specify arguments the skill uses as an input"'
+)
+
+
+def skill_name_for(agent_path: Path) -> str | None:
+    """Derive the skill directory name from an agent filename."""
+    if agent_path.suffix != ".py" or not agent_path.stem.endswith("_agent"):
+        return None
+    return agent_path.stem.removesuffix("_agent")
+
+
+def get_repo() -> Repo:
+    """Return the Repo object for the current working directory."""
+    return Repo(".", search_parent_directories=True)
+
+
+def get_upstream(repo: Repo) -> str:
+    """Return the remote tracking branch, falling back to origin/main."""
+    try:
+        tracking = repo.active_branch.tracking_branch()
+        if tracking is None:
+            return "origin/main"
+        return str(tracking)
+    except (TypeError, ValueError):
+        return "origin/main"
+
+
+def get_branch_diff(repo: Repo, upstream: str) -> set[Path]:
+    """Return the set of files changed between upstream and HEAD.
+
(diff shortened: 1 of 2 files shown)


scripts/sync-agent-skills.py
```diff
@@ -0,0 +1,113 @@
...
+        name = skill_name_for(agent_file)
+        if name is None:
+            continue
+
+        skill_path = SKILLS_DIR / name / "SKILL.md"
+        if not skill_path.exists():
+            continue
+
+        if str(skill_path) not in changed_files:
+            missing.append((agent_file, str(skill_path)))
```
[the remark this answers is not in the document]
TomasKorbar: This is intentional

scripts/sync-agent-skills.py
```diff
@@ -0,0 +1,113 @@
...
+
+AGENTS_DIR = "ymir/agents/"
+SKILLS_DIR = Path("agents_as_skills")
+
+CLAUDE_CMD_TEMPLATE = (
+    "claude --model claude-sonnet-4-6 --effort high"
+    " --allowedTools Read Glob Grep"
+    ' "Please take a look at the BeeAI workflows implemented in agents'
+    " directory. Please convert Workflow in {workflow_file} to Claude skill and"
+    " save that skill to agents_as_skills directory.\n"
```
nforro: Shouldn't we use the `SKILLS_DIR` constant here?

scripts/sync-agent-skills.py
```diff
@@ -0,0 +1,113 @@
...
+def skill_name_for(agent_path: str) -> str | None:
+    """Derive the skill directory name from an agent filename.
+
+    ymir/agents/triage_agent.py              -> triage
+    ymir/agents/preliminary_testing_agent.py -> preliminary_testing
+    """
+    basename = os.path.basename(agent_path)
+    if not basename.endswith("_agent.py"):
+        return None
+    return basename.removesuffix("_agent.py")
```
nforro: Can we be consistent and use `pathlib` everywhere? And follow the contribution guidelines for (not only) [docstrings](https://github.com/packit/contributing#examples-for-google-style-docstrings)?

```suggestion
def skill_name_for(agent_path: Path) -> str | None:
    """
    Derives the skill directory name from an agent filename. For example:
    
    ymir/agents/triage_agent.py -> triage
    ymir/agents/preliminary_testing_agent.py -> preliminary_testing
    
    Args:
        agent_path: Path to the agent file.
        
    Returns:
        Skill directory name or `None` if the specified path is not an agent file.
    """
    if not agent_path.name.endswith("_agent.py"):
        return None
    return agent_path.name.removesuffix("_agent.py")
```

Even though in this case I would consider the function self-explanatory and either skip the docstring entirely or use a oneliner.

scripts/sync-agent-skills.py
```diff
@@ -0,0 +1,113 @@
...
+
+def get_branch_diff(upstream: str) -> set[str]:
+    """Return the set of files changed between upstream and HEAD."""
+    result = subprocess.run(
+        ["git", "diff", "--name-only", f"{upstream}...HEAD"],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return {f for f in result.stdout.strip().splitlines() if f}
```
nforro: What if the call fails? I wonder if it wouldn't be better to use GitPython...

scripts/sync-agent-skills.py
nforro: This is a bit inconsistent, I would either use only list comprehensions (they could even be combined into one) or a single `for` loop. 

```suggestion
    agent_files = [f for f in changed_files if f.startswith(AGENTS_DIR) and f.endswith("_agent.py")]
    missing = [
        (f, sp)
        for f in agent_files
        if (n := skill_name_for(f)) is not None
        and (sp := SKILLS_DIR / n / "SKILL.md").exists()
        and str(sp) not in changed_files
    ]
```

Or directly:

```suggestion
    missing = [
        (f, sp)
        for f in changed_files
        if f.startswith(AGENTS_DIR) and f.endswith("_agent.py")
        and (n := skill_name_for(f)) is not None
        and (sp := SKILLS_DIR / n / "SKILL.md").exists()
        and str(sp) not in changed_files
    ]
```

scripts/sync-agent-skills.py
TomasKorbar: right. Fixed.

scripts/sync-agent-skills.py
TomasKorbar: I changed the comment to oneliner and switched to pathlib everywhere.

scripts/sync-agent-skills.py
TomasKorbar: Right, GitPython seems better for this. Changed.

scripts/sync-agent-skills.py
TomasKorbar: I guess this seems more compact, just changed some lines to use pathlib methods. Should be good.

nforro (approved):
LGTM

TomasKorbar:
@nforro thanks for the review 🙌

PR #421, opened 2026-04-24
Improve backport cherry-pick workflow: patch application fixes and agent tuning

opohorel:
## Summary
- Fix `apply_downstream_patches` to use `patch(1)` instead of `git apply`, matching RPM `%prep` behavior (handles gendiff patches, trailing whitespace, fuzz matching)
- Skip conditionally disabled patches (e.g., `%if 0` blocks) in `GetPackageInfoTool` so they're not passed to `apply_downstream_patches`
- Add `tag` and `commit` overrides to `FindBaseCommitTool` for repos with non-standard tag schemes or no tags
- Detect all git unmerged states (`AU`/`UA`/`UD`/`DU`) in cherry-pick tools and handle empty cherry-picks with `allow_empty`
- Remove reflexive "abort cherry-pick, use git-am" fallback from all tool error messages — only fall back when cherry-pick setup genuinely fails
- Condense cherry-pick instructions: ~50-line conflict section down to ~15 lines, fix step numbering contradictions, add edge cases (merge commits, empty commits)
- Drop forced ReAct think-after-every-tool pattern from backport agent to reduce LLM calls during multi-step conflict resolution
## Details
**Patch application (`apply_downstream_patches`):** Switched from `git apply` to `patch -p1` to match what RPM's `%prep` actually does. This fixes packages with gendiff-style patches, trailing whitespace, and patches that need fuzz matching. Also uses the `specfile` library's `Patch.valid` property to exclude patches inside false `%if` conditionals, preventing failures on disabled patches like vim's `vim-7.0-hunspell.patch` (`%if %{withhunspell}` where `withhunspell=0`).

**Cherry-pick tool improvements:** `FindBaseCommitTool` gains `tag` and `commit` parameters for explicit override when pattern search fails (e.g., libexpat's `R_2_6_4` tags). `CherryPickCommitTool` now detects all git unmerged states (`AU`/`UA`/`UD`/`DU`, not just `UU`/`AA`/`DD`) and handles the "cherry-pick is now empty" case. `CherryPickContinueTool` removes leftover debug prints, auto-stages with `git add -A`, adds `allow_empty` support, and fixes a misleading error message that told the LLM to run `git add` instead of resolving conflict markers.

**Fallback policy:** All "Abort cherry-pick approach, use git am workflow" strings removed from cherry-pick tools. Error messages now report what went wrong and suggest investigation steps. Approach B is only triggered when cherry-pick setup fails (URL extraction, clone, or downstream patch application). Once cherry-picking has started, the agent resolves all errors in place. Approach B steps renamed to `B1`/`B2` to avoid numbering collision, and `B1` now includes `cherry-pick --abort` cleanup.

**Agent configuration:** Removed `force_after=Tool` from the backport agent's ThinkTool `ConditionalRequirement`. The forced ReAct pattern (think between every tool call) doubled the number of LLM calls needed for conflict resolution, causing context to grow past limits before the workflow could complete. The model still thinks at step 1 and can think voluntarily, but isn't forced to between every tool call. Added minimal prompt nudges: "do not restart from step 1" and "replace with inline equivalent code" for missing functions during conflict resolution.

### DIFF
--- a/ymir/tools/unprivileged/upstream_tools.py
+++ b/ymir/tools/unprivileged/upstream_tools.py
@@ -484,7 +484,7 @@ class ApplyDownstreamPatchesTool(Tool[ApplyDownstreamPatchesToolInput, ToolRunOp
     all the patches that are already part of the package. After this, we can cherry-pick
     the new fix on top.
 
-    The patches are applied in order using 'git apply' and committed. If a patch fails to apply,
+    The patches are applied in order using patch(1) and committed. If a patch fails to apply,
     the tool returns an error indicating which patch failed.
     """
     input_schema = ApplyDownstreamPatchesToolInput
@@ -535,15 +535,16 @@ class ApplyDownstreamPatchesTool(Tool[ApplyDownstreamPatchesToolInput, ToolRunOp
                         "Downstream patches cannot be applied; cherry-pick workflow is not viable."
                     )
 
-                # Try to apply the patch with git apply and commit
-                # Use git apply instead of git am because dist-git patches can be plain diffs, not mbox format
-                cmd = ["git", "apply", str(patch_path)]
+                # Use patch(1) instead of git-apply to match RPM %prep behavior.
+                # Handles gendiff-style patches, trailing whitespace, fuzz
+                # matching, and other quirks that git-apply rejects.
+                cmd = ["patch", "-p1", "-s", "--batch", "--no-backup-if-mismatch", "-i", str(patch_path)]
                 exit_code, stdout, stderr = await run_subprocess(cmd, cwd=tool_input.repo_path)
 
                 if exit_code != 0:
                     raise ToolError(
                         f"Failed to apply existing patch '{patch_file}' to upstream base version. "
-                        f"Git apply error: {stderr}. "
+                        f"Patch error: {stderr or stdout}. "
                         f"Successfully applied: {', '.join(applied_patches) if applied_patches else 'none'}. "
                         "Downstream patches cannot be applied; cherry-pick workflow is not viable."
                     )
--- a/ymir/tools/unprivileged/upstream_tools.py
+++ b/ymir/tools/unprivileged/upstream_tools.py
@@ -326,21 +326,30 @@ class CloneUpstreamRepositoryTool(Tool[CloneUpstreamRepositoryToolInput, ToolRun
 class FindBaseCommitToolInput(BaseModel):
     repo_path: AbsolutePath = Field(description="Absolute path to the cloned upstream repository")
     version: str = Field(description="Version string to find (e.g., '2.5.3')")
+    tag: str | None = Field(
+        default=None,
+        description="Exact tag name to checkout, skipping pattern search. "
+        "Use when you already know the correct tag from the available tags list.",
+    )
+    commit: str | None = Field(
+        default=None,
+        description="Exact commit hash to checkout as the base. "
+        "Use when the upstream doesn't use tags or you know the exact release commit.",
+    )
 
 
 class FindBaseCommitTool(Tool[FindBaseCommitToolInput, ToolRunOptions, StringToolOutput]):
     name = "find_base_commit"
     description = """
-    Find and checkout a git tag matching the specified version in an upstream repository.
+    Find and checkout the base version commit in an upstream repository.
 
-    This tool tries common tag naming patterns:
-    - v{version} (e.g., v2.5.3)
-    - {version} (e.g., 2.5.3)
-    - release-{version} (e.g., release-2.5.3)
-    - {version}-release (e.g., 2.5.3-release)
+    Accepts three modes (highest priority first):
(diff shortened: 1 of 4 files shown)


ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -742,3 +897,203 @@ async def _run(
...
+
+        # Strategy 2: grep-based symbol search
+        if tool_input.symbol_hint:
+            cmd = [
+                "git",
+                "grep",
+                "-rn",
+                "--",
+                tool_input.symbol_hint,
+            ]
```
lbarcziova: should we try to limit this somehow? For the case of grepping something that is present in thousands of lines?

ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -553,9 +584,113 @@ async def _run(
...
+            f"-L{start_line},{end_line}:{file_path}",
+            f"{base_commit}..{fix_commit}",
+        ]
+        exit_code, stdout, _ = await asyncio.wait_for(
+            run_subprocess(cmd, cwd=repo_path),
+            timeout=timeout,
+        )
+        if exit_code == 0 and stdout and stdout.strip():
+            return stdout.strip()
+    except Exception:
```
lbarcziova: should we at least log/return the exception here?

lbarcziova (dismissed):
these seem like great improvements, thanks!

ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -742,3 +897,203 @@ async def _run(
...
+    )
+    max_chars: int = Field(
+        default=5000,
+        description="Maximum characters in the output. Increase for complex conflicts, "
+        "decrease to save context.",
+    )
+
+
+class HistoryTool(Tool[HistoryToolInput, ToolRunOptions, StringToolOutput]):
+    name = "line_history"
```
TomasTomecek: glad that agent will see it as `line_history` because `HistoryTool` is a very generic name

ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -742,3 +897,203 @@ async def _run(
...
+        missing_name = Path(tool_input.missing_file_path).name
+
+        # Strategy 1: git rename detection
+        base = self.options.get("base_tag_commit")
+        fix = self.options.get("fix_commit")
+        if base and fix:
+            cmd = [
+                "git",
+                "diff",
+                "-M25%",
```
TomasTomecek: wow, first time seeing this option, nice! though, isn't 25% rename similarity too low? or is this based on your experience?

TomasTomecek (dismissed):
outstanding work, just one question

ymir/tools/unprivileged/upstream_tools.py
opohorel: It's intentionally low. By default, git uses 50%, so I figured the agent is going to use the tool when the default command fails to find similarity

ymir/tools/unprivileged/upstream_tools.py
opohorel: Good point. It is true that if the answer contains a lot of symbols, it was probably bad call to use the tool to find it anyway.

ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -368,48 +377,66 @@ async def _run(
...
-            for tag in tag_patterns:
-                # Check if tag exists
-                cmd = ["git", "rev-parse", "--verify", f"refs/tags/{tag}"]
-                exit_code, stdout, stderr = await run_subprocess(cmd, cwd=tool_input.repo_path)
-
-                if exit_code == 0:
-                    found_tag = tag
-                    break
+            if tool_input.commit:
+                cmd = ["git", "cat-file", "-t", tool_input.commit]
```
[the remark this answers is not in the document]
opohorel: good catch!

ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -504,20 +532,21 @@ async def _run(
...
+                        "Downstream patches cannot be applied; cherry-pick workflow is not viable."
                     )
 
-                # Try to apply the patch with git apply and commit
-                # Use git apply instead of git am because dist-git patches can be plain diffs, not mbox format
-                cmd = ["git", "apply", str(patch_path)]
+                # Use patch(1) instead of git-apply to match RPM %prep behavior.
+                # Handles gendiff-style patches, trailing whitespace, fuzz
+                # matching, and other quirks that git-apply rejects.
+                cmd = ["patch", "-p1", "-s", "--batch", "--no-backup-if-mismatch", "-i", str(patch_path)]
```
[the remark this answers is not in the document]
opohorel: out of scope for now, I'm planing on implementing it later

lbarcziova (approved):
🚀

PR #423, opened 2026-04-27
Add CVE applicability check to triage workflow

lbarcziova:
After triage determines a backport/rebuild resolution, a new workflow step clones the package source, unpacks it, and runs a dedicated LLM agent to verify whether the vulnerable code actually exists in the shipped version. If not applicable, the resolution is overridden to NOT_AFFECTED with a Red Hat justification category, and no downstream work is queued.

Changes:
- Add Resolution.NOT_AFFECTED, NotAffectedData, ApplicabilityResult models
- Add cve_id field to RebuildData with rebuild output example in prompt
- Add TRIAGED_NOT_AFFECTED Jira label
- Move get_unpacked_sources to tasks.py, add clone_and_prep_sources with fallback tarball extraction when prep fails
- Add check_cve_applicability workflow step with ApplicabilityAgent (has access to source files, shell, and get_jira_details)
- Handle NOT_AFFECTED in queue mode (label update, no queue push)
- Install srpm-macros packages in both container images
- Fall back to CentOS Stream branch when z-stream branch doesn't exist

Assisted-by: Claude Opus 4.6

Example output: https://redhat.atlassian.net/browse/RHEL-167657?focusedCommentId=16798248

### DIFF
--- a/ymir/agents/tasks.py
+++ b/ymir/agents/tasks.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from urllib.parse import urlparse
 
 from beeai_framework.tools import Tool
+from specfile import Specfile
 
 from ymir.agents.constants import BRANCH_PREFIX, JIRA_COMMENT_TEMPLATE
 from ymir.agents.utils import check_subprocess, mcp_tools, run_subprocess, run_tool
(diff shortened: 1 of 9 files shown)


ymir/agents/tasks.py
```diff
@@ -341,3 +342,123 @@ async def cache_mr_metadata(
...
+    if sources_dir.is_dir():
+        return sources_dir
+
+    raise ValueError(f"Unpacked source directory does not exist: {sources_dir}")
+
+
+async def _fallback_extract_sources(local_clone: Path) -> Path:
+    """
+    Fallback when centpkg/rhpkg prep fails: manually extract the largest
+    source tarball found in the clone directory.
```
nforro: That's pretty naive, many projects have test suites (packaged separately) much larger than actual source code (although I didn't do any statistics). How often and why does executing `%prep` fail?

ymir/agents/triage_agent.py
```diff
@@ -818,6 +845,197 @@ async def verify_rebase_author(state):
...
+                        )
+                        patch_name = f"{state.jira_issue}-{idx}.patch"
+                        (local_clone / patch_name).write_text(content)
+                        patch_files.append(patch_name)
+                    except Exception:
+                        logger.warning(f"Could not fetch patch from {url}")
+
+                local_tool_options = {"working_directory": local_clone}
+                jira_tool = [t for t in gateway_tools if t.name == "get_jira_details"]
+                applicability_agent = RequirementAgent(
```
nforro: Could we move the agent along with related functions and variables to a separate module?

TomasTomecek (dismissed):
The sample analysis it did is pretty solid! I agree with Nikola it would be nice to move the new agent into a new module

ymir/agents/tasks.py
lbarcziova: that's a good point, what about using value of Source0 from specfile?

> How often and why does executing %prep fail?

I don't have statistics, but I think it can happen easily if we don't have some particular macros installed, I bumped into this when running the analysis on `butane` package which uses %goprep

nforro: > that's a good point, what about using value of Source0 from specfile?

I think that would be more reliable.

> I don't have statistics, but I think it can happen easily if we don't have some particular macros installed, I bumped into this when running the analysis on `butane` package which uses %goprep

That's a similar problem to spec file parsing, also `%prep` is a shell script that can execute anything, which is again similar to unsanitized spec file parsing, perhaps in the future we could solve both with the spec parsing service.

ymir/agents/tasks.py
```diff
@@ -341,3 +342,123 @@ async def cache_mr_metadata(
...
+    """
+    Fallback when centpkg/rhpkg prep fails: extract the primary source
+    tarball using Source0 from the spec file, or the largest tarball if
+    the spec can't be parsed.
+    """
+    try:
+        with Specfile(local_clone / f"{package}.spec") as spec, spec.sources() as sources:
+            if not sources:
+                raise ValueError(f"No sources defined in {package}.spec")
+            tarball = local_clone / sources[0].expanded_filename
```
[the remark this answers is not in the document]
lbarcziova: this is incorrect

ymir/agents/tasks.py
lbarcziova: implemented, PTAL

ymir/agents/triage_agent.py
lbarcziova: done

ymir/agents/tasks.py
```diff
@@ -341,3 +342,122 @@ async def cache_mr_metadata(
...
+            if not tarball.is_file():
+                raise ValueError(f"Source0 '{sources[0].expanded_filename}' not found on disk")
+            logger.info(f"Using Source0 from spec: {tarball.name}")
+    except Exception as e:
+        raise ValueError(f"Could not determine source tarball for {package}: {e}") from e
+
+    extract_dir = local_clone / "_extracted"
+    extract_dir.mkdir(exist_ok=True)
+
+    cmd = ["tar", "xf", str(tarball), "-C", str(extract_dir)]
```
nforro: There are a lot more archive types that RPM (`%setup`) supports, many not even tars:
https://github.com/packit/specfile/blob/43174b05acad3c534fb7557d4251540ba84f6487/specfile/spec_parser.py#L130-L140

nforro: It should be possible to use `%{__rpmuncompress} -x $ARCHIVE`, after expanding the macro of course.

nforro: Or call rpmuncompress directly:
```bash
$ /usr/lib/rpm/rpmuncompress --help
Usage: rpmuncompress [OPTION...]
  -x, --extract         extract an archive
  -v, --verbose         provide more detailed output
  -n, --dry-run         only print what would be done
  -C, --path=STRING     extract into a specific path

Help options:
  -?, --help            Show this help message
      --usage           Display brief usage message
```

lbarcziova: that's a valid point, but I don't think we need to cover every format from the start, if this fails the applicability check just gets skipped and triage continues normally (same as without this feature). We can extend it as we broaden the scope of triaging. But if you know about a more generic solution right away, happy to look into it.

lbarcziova: Github didn't show me your 2 latest comments, so ignore the last sentence :D I will have a look how complex this would be

lbarcziova: changed, PTAL

nforro (approved):

ymir/agents/triage_agent.py
```diff
@@ -818,6 +845,122 @@ async def verify_rebase_author(state):
...
+                    unpacked_sources=unpacked_sources,
+                    local_clone=local_clone,
+                )
+
+                response = await applicability_agent.run(
+                    prompt,
+                    expected_output=ApplicabilityResult,
+                    **get_agent_execution_config(),
+                )
+                applicability = ApplicabilityResult.model_validate_json(response.last_message.text)
```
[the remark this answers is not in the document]
nforro: `TriageAgent` does so because the LLM needs to produce nested JSON that some models struggle with, that's not the case here.

PR #424, opened 2026-04-27
[WIP] Parsing current RHEL Streams from Product Pages

mruprich:
TODO:

- [x] Write new tests or update the old ones to cover new functionality.
- [ ] ‹fill in›

The Product Pages should be the source of truth for this, but getting current and upcoming z-streams and current y-streams is tricky since they overlap during the DevDocTest phases. My logic is as follows:

1. If there are more than one active y-stream branches ATM per major RHEL, that means that the lower stream is currently in the very late phase of DevDocTest and right now it is being prepared for GA. That means it will very soon become an active z-stream, but right now it is the upcoming z-stream
2. I did not find a way to get active z-streams, other than look for the (GA/ZStream) in the long name, that marks the current z-stream

Anyway, the fetch_rhel_streams_snapshot() function should return a structure similar to what load_rhel_config() function returns now except for the package_instructions part. It will only work with a valid Kerberos ticket, otherwise it will raise a ToolError.

I am marking this as WIP for now so let me have it. Let me know if I chose a good place for the code.

RELEASE NOTES BEGIN

Packit now supports automatic parsing of all the current RHEL streams.

RELEASE NOTES END


ymir/common/product_pages.py
```diff
@@ -0,0 +1,255 @@
...
+    if response.status_code != 200:
+        raise ToolError(f"Product Pages API error ({what}): expected HTTP 200, got {response.status_code}")
+
+
+def _fetch_rhel_streams_snapshot_sync() -> dict[str, dict[str, str]]:
+    """Blocking implementation: HTTP via ``requests`` / GSSAPI."""
+    timeout = _PRODUCT_PAGES_TIMEOUT
+    try:
+        with requests.Session() as s:
+            auth = requests_gssapi.HTTPSPNEGOAuth(mutual_authentication=requests_gssapi.OPTIONAL)
```
TomasTomecek, line 197: we need to make sure this works okay with our kerberos setup

TomasTomecek (commented):
very nice!! it would be lovely to get at least a single test case that has mocked data with what PP returns (fake data obviously so we're not leaking anything here)

ymir/common/product_pages.py
mruprich: This should work with an active and valid krb ticket in your system but I realize that packit might have a different way to get a ticket right?

TomasKorbar: Take a look at `init_kerberos_ticket` function. That is the one used to obtain kerberos ticket or verify it already exists.

mruprich: Pushing another round of changes. I used the init_kereberos_ticket to get a ticket. I hope I am using that right.

I also added a unit test for the product pages..

ymir/common/product_pages.py
```diff
@@ -0,0 +1,231 @@
...
+        if _GA_ZSTREAM_RE.search(item.get("name_incl_maint") or "")
+    ]
+
+    current_z_streams = _build_current_z_streams_ga_zstream(filtered)
+
+    return {
+        "current_y_streams": current_y_streams,
+        "current_z_streams": current_z_streams,
+        "upcoming_z_streams": upcoming_z_streams,
+    }
```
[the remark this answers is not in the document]
mruprich: 1. added
2. I don't think that is possible. I did not find a way to get all that data in a single API call. The active API call gives very different results from the second one that grabs everything.
3. added
4. added

TomasTomecek:
Just a headsup a related commit was merged recently: https://github.com/packit/ai-workflows/pull/434/changes/a0c211a8f0f3043b23c122b1486486e873e4cc59

mruprich:
One question. I see the internal_repos_host being pulled from the rhel-config.json in ymir/tools/privileged/copr.py. What should I do about that? That is not possible to pull from product pages...

Maybe @nforro or @opohorel are aware of that option? Any idea whether this is actually used somewhere?

opohorel:
> One question. I see the internal_repos_host being pulled from the rhel-config.json in ymir/tools/privileged/copr.py. What should I do about that? That is not possible to pull from product pages...
> 
> Maybe @nforro or @opohorel are aware of that option? Any idea whether this is actually used somewhere?

AFAIK it is used for z-stream builds where the repo is added into the copr. If I remember correctly we had some issues with copr not having all the latest packages which were in the z-stream repo. With internal_repos_host defined you always get the latest z-stream packages to build against

@nforro knows more about it as he implemented it in #347

nforro:
It's the location of Brew repos, and it has nothing to do with Product Pages, it's in rhel-config.json only because it's an internal URL (and yes, it changes in time, so it can't be hardcoded).

mruprich:
Ok, thanks. So one solution would be to keep the rhel-config.json at least for copr.py to use. That requires minimal changes in copr.py.

Maybe the rhel-config.json could be used in the future as some sort of a cache? So that you don't have to contact PP every time you do something? Might save some bandwidth and you would be able to work even when the PP is unavailable. There would have to be a mechanism to make sure that it is up-to-date though...

nforro (changes requested):
Shouldn't this be a privileged MCP tool as it accesses internal data?

nforro:
> Ok, thanks. So one solution would be to keep the rhel-config.json at least for copr.py to use. That requires minimal changes in copr.py.

Yes, we can rename the file later or use an environment variable for the URL instead, but that's out of scope of this PR.

> Maybe the rhel-config.json could be used in the future as some sort of a cache? So that you don't have to contact PP every time you do something? Might save some bandwidth and you would be able to work even when the PP is unavailable. There would have to be a mechanism to make sure that it is up-to-date though...

Some caching mechanism would be definitely welcome, but I don't think it should be a file in `.secrets`.

mruprich:
@nforro the placement of the product-pages.py is really up to you guys, I will move it if necessary.

nforro:
> @nforro the placement of the product-pages.py is really up to you guys, I will move it if necessary.

It's more about not calling `fetch_rhel_streams_snapshot()` from anywhere except a privileged MCP tool than about location of the file. There are plenty of examples of such tools, how they are registered and how they are called in the code.

Perhaps it would be easier for you to contribute just product-pages.py in a form of standalone "library" and leave the rest on us. Kerberos ticket initialization would be the tool's responsibility and raising `ToolError`s from `fetch_rhel_streams_snapshot()` seems wrong anyway.

mruprich:
@nforro You are right, what if I remove the kerberos handling altogether? I would leave this up to the privileged tool to obtain a kerberos ticket and fail with exception if it is not available. So the `fetch_rhel_streams_snapshot()` would be called by the tool AFTER the tool already has the kerberos ticket.

nforro:
> @nforro You are right, what if I remove the kerberos handling altogether? I would leave this up to the privileged tool to obtain a kerberos ticket and fail with exception if it is not available. So the `fetch_rhel_streams_snapshot()` would be called by the tool AFTER the tool already has the kerberos ticket.

That's what I meant - calling `fetch_rhel_streams_snapshot()` assumes a valid Kerberos ticket, it will raise an exception otherwise (no need for explicit check, an exception should be raised by `requests` on session initialization).

mruprich:
Got it, thanks! I'll make the changes.

mruprich:
@nforro sorry that this took so long, too many CVEs :D So I did what you suggested, I am now only adding the product_pages.py (location can be changed here or in the next MR). I think that separating the addition of the code and the actual usage of the new code in two (or more) different MRs is even better.

nforro:
> @nforro sorry that this took so long, too many CVEs :D So I did what you suggested, I am now only adding the product_pages.py (location can be changed here or in the next MR). I think that separating the addition of the code and the actual usage of the new code in two (or more) different MRs is even better.

No problem at all, we've been quite busy as well :sweat_smile: I'll try to take a look ASAP.

ymir/common/product_pages.py
```diff
@@ -0,0 +1,286 @@
...
+import asyncio
+import json
+import os
+import re
+from collections import defaultdict
+from functools import cache
+
+import requests
+import requests_gssapi
+from beeai_framework.tools import ToolError
```
nforro, line 25: Nothing from BeeAI should be used here, including `ToolError`.

ymir/common/product_pages.py
```diff
@@ -0,0 +1,286 @@
...
+    return True
+
+
+def _rhel_sort_key(shortname: str) -> tuple[int, ...]:
+    """Sort key for RHEL shortnames by numeric major.minor (not lexicographic).
+
+    Example: rhel-10.3 sorts after rhel-9.9.
+
+    Returns:
+        Tuple of ints for lexical comparison ordering (major, minor, ...).
```
nforro, line 58: Shouldn't this be:
```suggestion
        Tuple of ints for numeric comparison ordering (major, minor, ...).
```

nforro (changes requested):
Mostly LGTM, just the test suite could be more extensive - multiple major releases, I wouldn't hesitate to use real world data, perhaps one or two minor releases back, just in case the current data is not supposed to be public.

PR #425, opened 2026-04-28
Triage agent upgrade

TomasKorbar:
- Run triage e2e tests in parallel
- Mock CentOS Stream RPM repositories during testing
- Merge the two triage prompts into one
- Add instructions for handling non-GitHub/non-GitLab upstreams with upstream-search
- Enforce URL scheme preservation

### DIFF
--- a/ymir/agents/tests/e2e/test_triage.py
+++ b/ymir/agents/tests/e2e/test_triage.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 
 import pytest
@@ -14,8 +15,10 @@ class TriageAgentTestCase:
         self.input: str = input
         self.expected_output: TriageOutputSchema = expected_output
         self.metrics: dict = None
+        self.finished_state: TriageState | None = None
+        self.error: BaseException | None = None
 
-    async def run(self) -> TriageState:
+    async def run(self) -> None:
         metrics_middleware = MetricsMiddleware()
 
         def testing_factory(gateway_tools):
@@ -23,9 +26,12 @@ class TriageAgentTestCase:
             triage_agent.middlewares.append(metrics_middleware)
             return triage_agent
 
-        finished_state = await run_workflow(self.input, False, testing_factory)
-        self.metrics = metrics_middleware.get_metrics()
-        return finished_state
+        try:
+            self.finished_state = await run_workflow(self.input, False, testing_factory)
+        except BaseException as e:
+            self.error = e
+        finally:
+            self.metrics = metrics_middleware.get_metrics()
 
 
 test_cases = [
@@ -102,8 +108,16 @@ def observability_fixture():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def mydata(request):
+def run_test_cases_concurrently(request):
+    """Execute all triage test cases concurrently via asyncio.gather, then collect metrics."""
+
+    async def _run_all():
+        await asyncio.gather(*(tc.run() for tc in test_cases))
+
+    asyncio.run(_run_all())
+
     yield
+
     collected_metrics = []
     for test_case in test_cases:
         if test_case.metrics is None:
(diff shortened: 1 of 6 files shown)


ymir/agents/tests/e2e/test_triage.py
```diff
@@ -8,24 +11,72 @@
...
+        self.error: BaseException | None = None
+        self.git_env: dict | None = None
 
-    async def run(self) -> TriageState:
+    async def run(self) -> None:
         metrics_middleware = MetricsMiddleware()
 
         def testing_factory(gateway_tools):
-            triage_agent = create_triage_agent(gateway_tools)
+            local_tool_options = {"env": self.git_env} if self.git_env else None
```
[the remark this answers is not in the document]
TomasKorbar: Not true, git works.

ymir/agents/tests/e2e/test_triage.py
```diff
@@ -102,8 +153,65 @@ def observability_fixture():
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def run_test_cases_concurrently(request, mock_centos_stream_repos):
+    """Execute all triage test cases concurrently via asyncio.gather, then collect metrics."""
+
+    async def _run_all():
+        await asyncio.gather(*(tc.run() for tc in test_cases))
+
+    asyncio.run(_run_all())
```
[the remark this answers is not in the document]
TomasKorbar: Running particular test cases is not in scope of this PR.

ymir/agents/tests/e2e/test_triage.py
```diff
@@ -112,19 +126,21 @@ def mydata(request):
...
+    assert test_case.finished_state is not None, f"Test case {test_case.input} did not produce a result"
+
+    real_output = test_case.finished_state.triage_result
+    expected_output = test_case.expected_output
+    assert real_output.resolution == expected_output.resolution
+    assert real_output.data.package == expected_output.data.package
+    assert real_output.data.patch_urls == expected_output.data.patch_urls
+    assert real_output.data.jira_issue == expected_output.data.jira_issue
+    assert real_output.data.cve_id == expected_output.data.cve_id
+    assert real_output.data.fix_version == expected_output.data.fix_version
```
nforro, line 240: Is there a field we don't compare? If not, we could just use `==` on the models.

TomasTomecek (commented):
Wow, you merged those massive prompts, so impressive!!

LGTM, couldn't find any obvious issues

TomasTomecek (dismissed):

ymir/agents/tests/e2e/test_triage.py
TomasTomecek, line 69: Gemini definitely has a point here but we'd easily see tests failing if Gemini was correct

ymir/agents/triage_agent.py
```diff
@@ -241,6 +259,26 @@ async def _map_version_to_branch(
            - If you run out of commits to check, use different approach, do not give up. Inability
              of the tool to find proper fix does not mean it does not exist, search bug trackers
              and version control system.
+           - **Handling non-GitHub/non-GitLab repositories**: When the upstream_search tool returns
+             `related_commits` that are bare commit hashes (not full URLs), it means the upstream
+             repository is hosted on a platform the tool does not know how to build patch URLs for
+             (e.g. gitweb, cgit, kernel.org, etc.). In this case, do NOT attempt to guess the web URL
+             or immediately call get_patch_from_url with a fabricated URL. Instead:
```
nforro:
```suggestion
             nor immediately call get_patch_from_url with a fabricated URL. Instead:
```

nforro (dismissed):
LGTM

ymir/agents/tests/e2e/test_triage.py
TomasKorbar: Yes, there is the justification which would require another LLM employed as a judge whether the justification is correct.

ymir/agents/triage_agent.py
TomasKorbar: right. Good catch, thanks.

ymir/agents/tests/e2e/test_triage.py
nforro: Right, it could still be "nicer" to exclude that field from `real/expected_output.__dict__` and compare the results, but I'm fine with this.

nforro (approved):

TomasKorbar:
Thanks @TomasTomecek and @nforro 🙌

PR #430, opened 2026-04-29
Add SILENT_RUN to triage agent

TomasKorbar:
Works in a simmilar way as the DRY_RUN does, except it does changes to Jira only when the resolution is either not-affected or postponed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ MOCK_JIRA ?= false
 JIRA_DRY_RUN ?= false
 AUTO_CHAIN ?= true
 FORCE_CVE_TRIAGE ?= false
+SILENT_RUN ?= false
 
 COMPOSE ?= $(shell if podman compose ls >/dev/null 2>&1; then echo "podman compose"; elif command -v podman-compose >/dev/null 2>&1; then echo "podman-compose"; else echo "docker-compose"; fi)
 COMPOSE_AGENTS=$(COMPOSE) -f $(COMPOSE_FILE) --profile=agents
@@ -26,6 +27,7 @@ run-triage-agent-standalone:
 		-e MOCK_JIRA=$(MOCK_JIRA) \
 		-e JIRA_DRY_RUN=$(JIRA_DRY_RUN) \
 		-e FORCE_CVE_TRIAGE=$(FORCE_CVE_TRIAGE) \
+		-e SILENT_RUN=$(SILENT_RUN) \
 		triage-agent
 
 .PHONY: run-triage-agent-e2e-tests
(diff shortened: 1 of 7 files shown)


ymir/agents/triage_agent.py
```diff
@@ -903,6 +918,7 @@ async def main() -> None:
     dry_run = os.getenv("DRY_RUN", "False").lower() == "true"
     auto_chain = os.getenv("AUTO_CHAIN", "true").lower() == "true"
     force_cve_triage = os.getenv("FORCE_CVE_TRIAGE", "false").lower() == "true"
+    silent_run = os.getenv("SILENT_RUN", "0") == "1"
```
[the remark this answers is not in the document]
TomasKorbar: fixed.

Makefile
```diff
@@ -5,6 +5,7 @@ MOCK_JIRA ?= false
 JIRA_DRY_RUN ?= false
 AUTO_CHAIN ?= true
 FORCE_CVE_TRIAGE ?= false
+SILENT_RUN ?= 0
```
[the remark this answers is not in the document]
TomasKorbar: fixed.

ymir/agents/triage_agent.py
```diff
@@ -960,17 +977,18 @@ async def retry(task, error, input=input):
...
+                        jira_issue=input.issue,
+                        labels_to_add=[JiraLabels.TRIAGE_IN_PROGRESS.value],
+                        labels_to_remove=[
+                            label
+                            for label in JiraLabels.all_labels()
+                            if label != JiraLabels.TRIAGE_IN_PROGRESS.value
+                        ],
+                        dry_run=dry_run,
+                    )
+                    logger.info(f"Cleaned up existing labels for {input.issue}")
```
nforro: In this condition branch manipulating Jira labels depends solely on the value of the environment variable while in the other branch it depends also on the triage resolution, is that on purpose?

TomasKorbar: Unfortunately we do not know the resolution of issue here yet. So i think the most reasonable thing is to stay silent.

nforro: Oh, right, of course :man_facepalming: I suppose that makes sense.

ymir/agents/triage_agent.py
```diff
@@ -61,6 +61,25 @@
 logger = logging.getLogger(__name__)
 
 
+def _should_update_jira(silent_run: bool, resolution: Resolution) -> bool:
```
nforro: Re: previous conversation, perhaps `resolution` could be optional here and if not set the function would return `False`?

nforro (dismissed):

ymir/agents/triage_agent.py
TomasKorbar: Hmm right, it could be. Will fix.

nforro (approved):
LGTM

TomasKorbar:
@nforro Thanks!

PR #432, opened 2026-04-30
Restore zstream search tool functionality broken by recent merges

antbob:
The zstream search tool functionality got broken by recent merges (didnt do blame as it doesnt really matter) so this commit restores the tool availability and related JSON unwrapping functionality it needs to process Jira Issues and Dev section commits.


lbarcziova (approved):
👍

PR #434, opened 2026-04-30
Improve rebuild workflow + SILENT_RUN extension + CVE logic adjustment

lbarcziova:
See individual commits

~~This also removes changing the Jira status from rebase/backport entirely, this is open to discussion, in my view maintainers can change the status of the Jiras themselves and we already track the status via labels. (cc @antbob , is this something you rely on?)~~


antbob:
@lbarcziova i know that certain parts of the existing automation for RHEL packages rely on In Progress for state transitions. IIRC one such place would be when requesting Preliminary Testing the issue has to be In Progress, and in Errata tool i think when issues are attached automatically. i can find out the exact details for you when i'm back on Monday but i would say thats probably not a good idea to remove it as far as full automation goes.

lbarcziova:
@antbob ok thanks for the info, for now I changed it to be gated on SILENT_RUN=true (we are trying to help with the CVE situation and wanted a bit less noisy behaviour), but will revisit with the rest of the team.

nforro (approved):
LGTM

PR #440, opened 2026-05-04
Mocking mkdir call on path object

jpodivin:
Minor test adjustment. It doesn't matter in a container, but when you execute tests directly on host, attempt to make `/git-repos/bash` dir fails for obvious reasons. We should talk about changing instructions in the README at some point.

But for now this will do. Plus, it will actually check that the call is made while preventing contamination of the test environment.


nforro (approved):
While the contribution guidelines don't specify it explicitly, the [linked docs](https://chris.beams.io/git-commit) about writing commit messages do - subject lines should be written in imperative mode.

PR #448, opened 2026-05-05
Add image build dependency to each container test target

jpodivin:
With this you can launch tests with `make check-in-container` and be sure that you are not running with a stale image, and you avoid the "choose image" dialog from podman.

This way we can be sure that we always test on the newest code. There is some redundancy involved, as the build is called even if the image is present. But with layer caching the time wasted become negligible. This also removes separate build image step in the workflow, as that would be really redundant.

### DIFF
--- a/.github/workflows/check-in-container.yml
+++ b/.github/workflows/check-in-container.yml
@@ -32,12 +32,6 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y podman
         podman --version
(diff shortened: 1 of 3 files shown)


nforro (commented):
Could we make this conditional - the images would be built by default, but users could override that with a variable?

jpodivin:
> Could we make this conditional - the images would be built by default, but users could override that with a variable?

Certainly.

nforro (approved):
Thanks!

PR #450, opened 2026-05-05
Mocking upgrade

TomasKorbar:
- Extract mock repo setup from e2e tests into a shared ymir/common/mock_repos.py module so the same mocking infrastructure can be reused for all mocking
- Add RunShellCommandTool URL blocking via MOCK_BLOCKED_URLS to prevent agents from bypassing mocked repos with curl/wget, replacing the previous shell wrapper approach
- Add scripts/run-mock-triage.py launcher and claude-triage-mock.md guide for running the triage skill against mocked JIRA data locally

### DIFF
--- a/scripts/run-mock-triage.py
+++ b/scripts/run-mock-triage.py
@@ -14,11 +14,11 @@ Prerequisites:
 """
 
 import argparse
-import atexit
 import json
 import os
 import shutil
 import stat
+import subprocess
 import sys
 import tempfile
 from pathlib import Path
@@ -173,7 +173,6 @@ def main() -> None:
     mcp_config = build_mcp_config(jira_mock_dir, mock_data_dir, upstream_search_url, blocked_urls, log_dir)
 
     tmp_fd, tmp_path = tempfile.mkstemp(prefix="mock-triage-mcp-", suffix=".json")
-    atexit.register(lambda: os.unlink(tmp_path))
 
     with os.fdopen(tmp_fd, "w") as fh:
         json.dump(mcp_config, fh, indent=2)
@@ -192,17 +191,19 @@ def main() -> None:
     print("============================")
     print()
 
-    os.chdir(REPO_ROOT)
-
-    os.execvp(
-        "claude",
-        [
-            "claude",
-            f"Use the triage skill with jira_issue={issue} and dry_run=true",
-            "--mcp-config",
-            tmp_path,
-        ],
-    )
+    try:
+        result = subprocess.run(
+            [
+                "claude",
+                f"Use the triage skill with jira_issue={issue} and dry_run=true",
+                "--mcp-config",
+                tmp_path,
+            ],
+            cwd=REPO_ROOT,
+        )
+        sys.exit(result.returncode)
+    finally:
+        os.unlink(tmp_path)
 
 
 if __name__ == "__main__":
--- a/scripts/run-mock-triage.py
+++ b/scripts/run-mock-triage.py
@@ -1,0 +1,209 @@
+#!/usr/bin/env python3
+"""Launch Claude Code with the triage skill against mocked JIRA data.
+
+Usage:
+    export TESTING_JIRAS_DIR=/path/to/testing-jiras
+    ./scripts/run-mock-triage.py RHEL-15216
+
+Prerequisites:
+    - claude CLI on PATH
+    - ymir-privileged-gateway and ymir-unprivileged-gateway on PATH
+    - TESTING_JIRAS_DIR pointing to a clone of
+      git@gitlab.cee.redhat.com:jotnar-project/testing-jiras.git
+    - rhel-config.json in the repo root (copied from templates/ if missing)
+"""
+
+import argparse
+import atexit
+import json
+import os
+import shutil
+import stat
+import sys
+import tempfile
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+DEFAULT_UPSTREAM_SEARCH_URL = "http://upstream-search.hosted.upshift.rdu2.redhat.com:80/v1"
+
+
+def fail(msg: str) -> None:
+    print(f"ERROR: {msg}", file=sys.stderr)
+    sys.exit(1)
+
+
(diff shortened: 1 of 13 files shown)


ymir/common/mock_repos.py
```diff
@@ -0,0 +1,199 @@
...
+    apply_zstream_override(config.get("zstream_override"))
+
+    repos = config.get("repos")
+    if not repos:
+        return None
+
+    if base_dir is None:
+        import tempfile
+
+        base_dir = Path(tempfile.mkdtemp(prefix=f"mock_repos_{issue_key}_"))
```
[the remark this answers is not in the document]
TomasKorbar: This is used for testing and testing environment is most likely throwed away.

scripts/run-mock-triage.py
```diff
@@ -0,0 +1,209 @@
...
+
+    os.execvp(
+        "claude",
+        [
+            "claude",
+            f"Use the triage skill with jira_issue={issue} and dry_run=true",
+            "--mcp-config",
+            tmp_path,
+        ],
+    )
```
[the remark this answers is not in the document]
TomasKorbar: Fixed.

ymir/common/mock_repos.py
```diff
@@ -0,0 +1,230 @@
...
+                cwd=str(local_path),
+                text=True,
+                check=True,
+            )
+        subprocess.run(
+            ["git", "gc", "--prune=now", "-q"],
+            cwd=str(local_path),
+            check=True,
+            capture_output=True,
+        )
```
nforro: Wouldn't it be better to use GitPython?

TomasKorbar: Right, switched to GitPython.

ymir/agents/triage_agent.py
```diff
@@ -1028,6 +1028,34 @@ async def comment_in_jira(state):
...
+    git_env = setup_mock_repos_from_env(jira_issue)
+    if git_env is None:
+        return create_triage_agent
+
+    logger.info("Mock repos configured for %s via MOCK_REPOS_DIR", jira_issue)
+
+    def _factory(gateway_tools, local_tool_options=None):
+        return create_triage_agent(gateway_tools, {"env": git_env})
+
+    return _factory
```
nforro: I'm wondering, could this be turned into a decorator to decorate `create_triage_agent()` (and in the future other agent factories) with?

TomasKorbar: hmm let me try

ymir/tools/unprivileged/commands.py
```diff
@@ -12,6 +14,40 @@
...
+
+
+def _check_blocked_urls(command: str, blocked_urls: list[str]) -> str | None:
+    """If *command* invokes curl/wget targeting a blocked URL, return the URL."""
+    try:
+        tokens = shlex.split(command)
+    except ValueError:
+        tokens = command.split()
+
+    has_curl_or_wget = any(tok in ("curl", "wget") or tok.endswith(("/curl", "/wget")) for tok in tokens)
```
nforro: I would make those commands a constant. In the future we might want to extend the list.

ymir/tools/unprivileged/commands.py
```diff
@@ -12,6 +15,49 @@
...
+    Returns:
+        A list of URL prefix strings to block.
+    """
+    raw = os.getenv("MOCK_BLOCKED_URLS", "")
+    if not raw:
+        return []
+    raw = raw.strip()
+    if raw.startswith("["):
+        return json.loads(raw)
+    return [url.strip() for url in raw.split(",") if url.strip()]
```
nforro:
```suggestion
    if not raw := os.getenv("MOCK_BLOCKED_URLS", ""):
        return []
    raw = raw.strip()
    try:
        return json.loads(raw)
    except json.decoder.JSONDecodeError:
        return [stripped for url in raw.split(",") if (stripped := url.strip())]
```

Although the early return is rather unnecessary.

ymir/agents/triage_agent.py
TomasKorbar: I do not think decorator is suitable here as you get the jira issue key, which you need for choosing the appropriate mock repos during runtime. But perhaps higher order function would be better for this.
```
def build_agent_factory_with_mock_repos(
    agent_factory: Callable[[list, dict[str, Any] | None], Any], jira_issue: str
) -> Callable[[list, dict[str, Any] | None], Any]:
    """Wrap an agent factory with mock repo support when configured.

    If ``MOCK_REPOS_DIR`` is set, prepares mock repos and returns a factory
    that injects the resulting ``git_env`` into ``local_tool_options``.
    Also applies ``MOCK_ZSTREAMS`` (standalone env-var) and any per-issue
    ``zstream_override`` found in the config file.

    Args:
        agent_factory: The original agent factory callable
            (e.g. ``create_triage_agent``).
        jira_issue: The Jira issue key (e.g. ``RHEL-15216``).

    Returns:
        The original factory unchanged when mocking is not configured,
        or a wrapper that injects mock ``git_env``.
    """
    apply_zstream_override_from_env()

    git_env = setup_mock_repos_from_env(jira_issue)
    if git_env is None:
        return agent_factory

    logger.info("Mock repos configured for %s via MOCK_REPOS_DIR", jira_issue)

    def _factory(gateway_tools, local_tool_options=None):
        return agent_factory(gateway_tools, {"env": git_env})

    return _factory
```

And used like `agent_factory = build_agent_factory_with_mock_repos(create_triage_agent, jira_issue)`

TomasKorbar: I pushed this solution, so take a look.

ymir/tools/unprivileged/commands.py
TomasKorbar: Fixed.

ymir/tools/unprivileged/commands.py
TomasKorbar: Done.

ymir/agents/triage_agent.py
nforro: Nice, that's much better.

nforro (approved):

PR #451, opened 2026-05-05
Add per-agent chat model override via environment variables

lbarcziova:
Allow overriding CHAT_MODEL per agent type (CHAT_MODEL_TRIAGE, CHAT_MODEL_BACKPORT, CHAT_MODEL_REBASE, CHAT_MODEL_REBUILD). Falls back to the default CHAT_MODEL when unset.

Assisted-by: Claude Opus 4.6


opohorel (approved):
LGTM, thanks!

PR #452, opened 2026-05-06
Add operation-type labels to MRs for tracking statistics

lbarcziova:
Label each MR with ymir_backport, ymir_rebase, or ymir_rebuild so we can query GitLab for per-operation statistics.

Assisted-by: Claude Opus 4.6


opohorel (approved):
LGTM, thanks!

TomasTomecek (approved):

PR #455, opened 2026-05-06
Add explicit GitPython dependency into agent images

TomasKorbar:
Recent changes unfortunately listed GitPython dependency in python package dependencies but as we copy the files and do not use pip yet for installation of project, it broke c9s images.


opohorel (approved):
fixes the issue with failing build. thanks!

nforro (approved):

PR #457, opened 2026-05-06
Fix log agent reverting Release field changes during rebuild

majamassarini:
Sometimes release changes don't appear in the MR or they disappear after an agent re-run.

This fixed the problem in these two MRs:
https://gitlab.com/redhat/rhel/rpms/ignition/-/merge_requests/45
https://gitlab.com/redhat/rhel/rpms/delve/-/merge_requests/3


opohorel (approved):
lgtm, thanks!

PR #466, opened 2026-05-11
Add CVE-specific verification to triage agent

majamassarini:
Enhanced the triage agent to verify that chosen patches actually fix the correct CVE by:

1. Added mandatory CVE ID verification step in patch validation that:
   - Checks if patch content/commit mentions the exact CVE ID
   - Requires verification that vulnerability matches if CVE ID absent
   - Warns about bundled updates (Oracle CPU) fixing multiple CVEs
   - Prevents accepting wrong patches for similar vulnerabilities

2. Updated BackportData.justification field description to require explicit documentation of CVE verification process.

This prevents issues like RHEL-170461 where the agent selected a patch for CVE-2025-64505 instead of CVE-2026-22020 because both were libpng buffer overflow fixes from the same upstream.

### DIFF
--- a/ymir/agents/triage_agent.py
+++ b/ymir/agents/triage_agent.py
@@ -350,6 +350,17 @@ TRIAGE_PROMPT = """
               - The fix directly addresses the root cause identified in your analysis
               - The code changes align with the symptoms described in the Jira issue
               - The modified functions/files match those mentioned in the issue
+           3. **For CVE issues - Verify CVE ID match**: If the issue is a CVE (contains CVE-YYYY-NNNNN):
+              - Check if the patch content or commit message mentions the EXACT CVE ID
+              - If the CVE ID is NOT mentioned in the patch, verify that:
+                * The vulnerability description in the CVE matches what the patch fixes
+                * The code changes address the specific vulnerability type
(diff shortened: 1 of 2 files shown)


TomasTomecek (approved):
LGTM, great improvement! I feel like we need a reviewer agent for this class of issues

nforro (approved):

lbarcziova (approved):

majamassarini:
> LGTM, great improvement! I feel like we need a reviewer agent for this class of issues

it happened to me just once, and accordingly to claude this was a quite unfortunate issue:

```
Why the agent picked the wrong patch:

Vague Jira description: The issue says "LibPNG has several reported CVEs and needs to be updated" without specifying which exact vulnerability
Oracle CPU bundles multiple fixes: Oracle CPU 2026-04 likely includes fixes for multiple CVEs
Search confusion: The agent searched for libpng fixes and found a buffer overflow patch (6a528eb5f) that seemed relevant, but didn't verify it matched the specific CVE-2026-22020
Package name mismatch: The search was for libpng15, but the correct patch is from the main libpng repository
```

but yes a reviewer agent would be helpful for many reasons, in this case I think running the reproducer (if any) would also have helped

PR #476, opened 2026-05-12
Add AnalyzeLogsTool to privileged tools package

jpodivin:
Tool uses API calls to Log Detective server to analyze build logs.

The input for the tool follows Log Detective API, with notable exception of only allowing URLs.
This way we won't get model hallucinating logs into the call. We may still get hallucinated URLs though.

I did go relatively hard on the validation of input and output for the API, maybe even too hard. My intention was to make any inconsistencies in responses immediately obvious, so that the agent doesn't go off the rails from the unintended response.

The somewhat convoluted `get_text_content` of the output model is meant to minimize the amount of tokens that will pass into the context. Simple Json dump would be a bit too much. I do want to implement something similar in Log Detective agent itself. 

Tests are all passing. And I've tested the tool on randomly selected build from copr. It seems to work reasonably well.

image

RELEASE NOTES BEGIN

Log Detective can now be queried for analysis of logs using the `analyze_logs` tool. 

RELEASE NOTES END

### DIFF
--- a/ymir/tools/privileged/tests/unit/test_logdetective.py
+++ b/ymir/tools/privileged/tests/unit/test_logdetective.py
@@ -1,0 +1,307 @@
+from contextlib import asynccontextmanager
+
+import aiohttp
+import pytest
+from beeai_framework.tools import ToolError
+from flexmock import flexmock
+from pydantic import ValidationError
+
+from ymir.tools.privileged import logdetective as logdetective_module
+from ymir.tools.privileged.logdetective import (
+    AnalyzeLogsTool,
+    AnalyzeLogsToolOutput,
+    LogDetectiveFile,
+    LogDetectiveResult,
+    LogDetectiveSnippet,
+)
+
+SAMPLE_RESPONSE = {
+    "explanation": {"text": "Build failed due to missing dependency libfoo."},
+    "snippets": [
+        {
+            "text": "error: nothing provides libfoo needed by bar-1.0-1.el10.x86_64",
+            "line_number": 42,
+            "source_file": "build.log",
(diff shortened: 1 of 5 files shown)


ymir/tools/privileged/tests/unit/test_logdetective.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _patch_module_constants():
+    with (
+        patch.object(logdetective_module, "LOG_DETECTIVE_URL", BASE_URL),
+        patch.object(logdetective_module, "LOG_DETECTIVE_TOKEN", TOKEN),
+        patch.object(logdetective_module, "MAX_LOG_DETECTIVE_FILES", 5),
+    ):
+        yield
```
nforro: Can't we use pytest-native monkeypatch, e.g. like here?

https://github.com/packit/ai-workflows/blob/46376ed5b32074070b3f52737c9a0fa03e2aa1ba/ymir/jira_issue_fetcher/tests/unit/test_jira_issue_fetcher.py#L24-L31

nforro (dismissed):
LGTM, thanks!

ymir/tools/privileged/tests/unit/test_logdetective.py
jpodivin: I really prefer unittest patch... but it's not going to be difficult to rewrite it, so why not?

jpodivin: Done.

nforro (approved):

PR #483, opened 2026-05-13
Change bot identity from jotnar to ymir

majamassarini:
Alongside with this change you also need to use the rehat-ymir-agent keytab and update the gitlab token in the mcp-gateway.env file with the redhat-ymir-agent token


TomasTomecek (approved):
thank you! I'm gonna deploy this tomorrow

PR #485, opened 2026-05-13
openshift: docs, phoenix memory fix, and dist-git SSH startup check

TomasTomecek:
## Summary

- Document how to manually trigger the pipeline via `oc rsh` into valkey
- Increase phoenix memory limit from 512Mi to 1Gi (it was OOMKilled after ~2h of accumulating trace data)
- Add `check_distgit_ssh_access()` to mcp-gateway startup: SSHs to `pkgs.devel.redhat.com` without an explicit username, parses gitolite's `hello <user>` greeting, and raises if it doesn't match the Kerberos principal — catches SSH config `User` mismatches early, which is relevant because the service account name is expected to change soon
- Document `create_zstream_branch` internals: username always comes from the Kerberos principal (not SSH config), the ref comes from Koji build source metadata and may point to an older Z-stream branch via tag inheritance (expected), and push failures can be transient due to bastion network issues

## Investigation notes

A push of `rpms/flatpak` to `rhel-9.8.0` failed with a generic git error. We traced through the full path: SSH auth works (Kerberos GSSAPI via `bastion-iad2`), the commit exists in the local clone, gitolite confirmed `R W rpms/[a-zA-Z0-9].*` access, and a dry-run push succeeded. The original failure was transient. The `rhel-9.8.0` branch was never created.

## Test plan

- [ ] Verify mcp-gateway logs `Dist-git SSH access verified: authenticated as 'jotnar-bot'` on startup
- [ ] Re-queue RHEL-165643 and confirm backport agent creates the `rhel-9.8.0` branch successfully
- [ ] Monitor phoenix for OOM over the next few days

### DIFF
--- a/openshift/README.md
+++ b/openshift/README.md
@@ -86,3 +86,13 @@ Manually run jira issue fetcher:
 ```bash
 make run-jira-issue-fetcher
 ```
+
+## Triggering the Pipeline Manually
+
+To push a Jira issue into the triage queue (e.g. to force CVE triage):
+
+```bash
+oc rsh deployment/valkey redis-cli LPUSH triage_queue '{"metadata": {"issue": "RHEL-XXXXXX", "force_cve_triage": true}}'
+```
+
+Set `force_cve_triage` to `false` for a normal triage run. This mirrors the `make trigger-pipeline` target used locally.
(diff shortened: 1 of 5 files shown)


lbarcziova (approved):
thanks!

PR #493, opened 2026-05-18
Fix missing kerb mock in test

pdancak:
TODO:

- [X] Write new tests or update the old ones to cover new functionality.

I was trying to setup the environment and when I was running tests and few of them failed (the test_download_source). The args were not right, because kerberos was not mocked, like in test_prep_sources. So I have added _mock_kerberos func call so the environment is prepared. Now the tests pass and don't fail on klist.

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

RELEASE NOTES END


TomasTomecek (approved):
very nice, thanks for the contribution!

PR #494, opened 2026-05-18
Triage e2e suite and further improvements

TomasKorbar:
- Add CI configuration for running triage e2e tests via Testing Farm / Packit, with a setup guide in e2e-ci-setup.md
- Fix GetPatchFromUrlTool sending the private Red Hat token to public GitLab repos (e.g. gitlab.com/libtiff/libtiff), which caused HTTP 401 errors when validating upstream patches
- Strengthen triage prompt instructions: mandatory follow-up commit discovery (agent can no longer self-rationalize skipping them), and deterministic https:// URL scheme for patch URLs
- Set default LLM retry count to 3 to mitigate rate-limiting crashes in production
- Disable the unmaintained upstream-search tool (stale embeddings) and two flaky e2e test cases pending better debugging support
- Update the triage agent skill and skill-generation prompt

After merge of this, someone with sufficient rights will be able to setup e2e CI. I executed the suite 4 consecutive times and it passed every time, so it should be finally stable.


ymir/agents/triage_agent.py
```diff
@@ -540,7 +514,6 @@ def create_triage_agent(gateway_tools, local_tool_options=None):
             ThinkTool(),
             RunShellCommandTool(options=local_tool_options) if local_tool_options else RunShellCommandTool(),
             VersionMapperTool(),
-            UpstreamSearchTool(),
```
TomasTomecek, line 543: 😢

ymir/agents/utils.py
```diff
@@ -38,6 +38,10 @@ def get_chat_model() -> ChatModel:
             temperature=0.6
         ),
         timeout=1200,
+        # beeai hardcodes max_retries=0 in its litellm adapter; num_retries
+        # bypasses that and enables litellm's built-in retry with back-off
+        # for transient 429 / rate-limit errors from the provider.
+        settings={"num_retries": int(os.getenv("LITELLM_NUM_RETRIES", 3))},
```
TomasTomecek, line 44: nice, glad we don't need to implement this on our end

.packit.yaml
```diff
@@ -0,0 +1,50 @@
+# Packit CI configuration for triage agent e2e tests.
```
TomasTomecek, line 1: we need to make sure packit is installed in this repo, don't know if we have it in the whole namespace or per repo

plans/triage-e2e.fmf
```diff
@@ -0,0 +1,60 @@
...
+
+        cp /tmp/testing-jiras/mock_data/*.json \
+            ymir/agents/tests/e2e/mock_repos/
+
+        cp -r /tmp/testing-jiras/jiras/* \
+            ymir/tools/privileged/tests/data/
+
+        # Clone CI secrets repo and populate .secrets/
+        git clone --depth 1 \
+            "https://oauth2:${GITLAB_CI_TOKEN}@gitlab.cee.redhat.com/jotnar-project/ci-secrets.git" \
```
TomasTomecek, line 40: ah, so this repo holds the secrets

.packit.yaml
```diff
@@ -0,0 +1,50 @@
...
+#   - Packit GitHub App installed on this repo
+#   - Project approved for use_internal_tf (contact Packit team)
+#   - Encrypted GITLAB_CI_TOKEN in .testing-farm.yaml
+
+jobs:
+  - job: tests
+    trigger: pull_request
+    identifier: triage-e2e
+    skip_build: true
+    manual_trigger: true
```
TomasTomecek, line 20: +1
it would be wasteful to run this on every push

TomasTomecek (dismissed):
LGTM overall design

we also need to verify how the logs look like and how hard is to understand failures

.packit.yaml
```diff
@@ -0,0 +1,50 @@
...
+              tags:
+                - m1.large
+          tmt:
+            context:
+              how: full
+
+  - job: tests
+    trigger: commit
+    branch: main
+    identifier: triage-e2e
```
nforro: Having the same identifier for different test jobs is not a good idea, even though in practice it probably won't be a problem due to different triggers. Do we even need an identifier?

nforro (dismissed):
LGTM

.packit.yaml
TomasKorbar: I think it is better to have it, just to be sure. But good point i will use different identifier for each job.

nforro (approved):

PR #505, opened 2026-05-20
Use patch URL instead of API URL for auth headers check

antbob:
This fixes a bug introduced in 9277699 ("Fix GetPatchFromUrlTool when retrieving patches out of Redhat namespace") which breaks get_patch_from_url tool for Red Hat GitLab. The request URL never starts with "redhat", it is in API form like "https://gitlab.com/api/v4/projects/..." so instead of using that the logic should be using the original patch URL instead to make the correct determination.


TomasTomecek (approved):
Thank you!

PR #507, opened 2026-05-20
[WIP] Trying to improve the naming conventions of the backport agent

mruprich:
TODO:

- [ ] ‹fill in›

As a source of the rules I used the https://fedoraproject.org/wiki/PeterGordon/SpecFormattingGuidelines#Patching guide which seems quite reasonable to me and I am actually using something similar in my packages. I think that any naming convention that uses something descriptive is better than just plain RHEL-123456.

### DIFF
--- a/agents_as_skills/backport/SKILL.md
+++ b/agents_as_skills/backport/SKILL.md
@@ -107,6 +107,22 @@ Execute the following steps in order. Track state across steps (paths, flags, re
 
 Determine `pkg_tool` from the branch: if `dist_git_branch` starts with `c` and ends with `s` (e.g., `c10s`, `c9s`), use `centpkg`; otherwise use `rhpkg --offline --released`.
 
+### Patch File Naming
+
+All patch files in dist-git (and pre-downloaded upstream patches from Step 2) must use descriptive names — not Jira issue keys. Follow these rules:
+
+1. **Descriptive filenames**: Hyphen-separated words, prefixed with the package name (`{{package}}`). The name should describe the fix (from the diff, commit message, or upstream change), not the Jira key.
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):
This is really smart, great idea to not hardcode those patch names as we used to have this.

opohorel:
you are changing the skill only, which we regenerate quite often when doing any change to the prompt etc. it is almost certain that this change is going to be overwritten when we regenerate the skill again

mruprich:
@opohorel Aah, so I should make the change rather in ai-workflows/ymir/agents/backport_agent.py file?

opohorel:
> @opohorel Aah, so I should make the change rather in ai-workflows/ymir/agents/backport_agent.py file?

I'm working on it right now. currently we are hard-coding in the prompt and in some tools that the patch name is <JIRA_ISSUE>.patch and that the patch is only one, squashing all the backported commits. it is possible to force other convention in rules/, but then the agent has to avoid the tools. I'll CC you on the PR when it is ready

mruprich:
@opohorel So I am looking at github.com/packit/ai-workflows/pull/515 and I am not sure whether I should just close this one. From the part that edits the patch naming conventions, do I understand it right that every maintainer should just have their own rules on patch naming? If so, this PR is most likely ready to be closed than...?

mruprich:
@opohorel ping.

opohorel:
@mruprich sorry for late reply, I was on a PTO.
I think you can safely close this one. also AFAIK @TomasKorbar wanted to change the default to something more sensible than the <JIRA_ISSUE>.patch, but I'm not sure if that come through

lbarcziova:
The patch naming conventions have been improved in https://github.com/packit/ai-workflows/pull/583, so I believe we can close this?

PR #508, opened 2026-05-20
Pass changelog unmodified, let LogAgent swap the JIRA references

antbob:
Now the source changelog is passed through unmodified, and the LogAgent is told to only swap the JIRA issue references in Resolves/Related lines. This provides better changelog consistency based on our trials and also simplifies the logic making it less error prone. The downside is that the model can make a mistake when swapping JIRA references (if they are present) but that remains to be seen as so far it has behaved correctly on multiple trials.

### DIFF
--- a/ymir/agents/backport_agent.py
+++ b/ymir/agents/backport_agent.py
@@ -781,8 +781,9 @@ async def extract_source_changelog(
     """Extract changelog messages from source dist-git commits.
 
     Iterates all upstream patch URLs, extracts the newest changelog entry
-    from each commit's spec file, strips Resolves/Related lines, and
-    combines the descriptive lines (deduplicating across commits).
+    from each commit's spec file, and combines the lines (deduplicating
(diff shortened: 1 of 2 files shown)


antbob:
Gemini concerns are mostly correct but we cannot replace random references to Jira issues inside changelog messages as we have no context to do that correctly. In order to do something like that we would need to follow each such reference, analyze it and search for an equivalent or related issue for the target, if any exists at all. This would be over the top and not warranted for this. So any such references should stay as they are ie pointing to their original issues.

For Resolves/Related lines the instructions have been modified slightly to be less vague and ambiguous so that the agent can clearly follow the logic of replacing in this case ie removing the old references and adding new ones. It is possible to have edge/corner cases where the agent can still get confused by the original changelog messaging (as there is no common standard among packages or maintainers so anything goes pretty much) but this can be amended by the human in the loop if required as such cases should be relatively rare in practice.

nforro (approved):
LGTM

PR #510, opened 2026-05-21
Backport agent e2e suite infrastructure and first backport test

TomasKorbar:


ymir/agents/tests/e2e/backport_agent/test_backport.py
```diff
@@ -0,0 +1,350 @@
...
+@pytest.fixture(scope="session", autouse=True)
+def run_test_cases_concurrently(request, mock_centos_stream_repos):
+    """Execute all backport test cases concurrently via asyncio.gather, then collect metrics."""
+
+    async def _run_all():
+        await asyncio.gather(*(tc.run() for tc in test_cases))
+
+    asyncio.run(_run_all())
+
+    yield
```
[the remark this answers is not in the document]
TomasKorbar: This is actually a false positive. Pytest-asyncio is not really used in e2e testing as it would not allow us to run the tests concurently, but stays as dependency and in place because of other unit tests.

ymir/agents/tests/e2e/backport_agent/test_backport.py
```diff
@@ -0,0 +1,350 @@
...
+        )
+
+
+# ---------------------------------------------------------------------------
+# LLM judge tests (optional, controlled by RUN_LLM_JUDGE env var)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("test_case", test_cases, ids=[tc.jira_issue for tc in test_cases])
+def test_backport_agent_llm_judge(test_case: BackportAgentTestCase):
```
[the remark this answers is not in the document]
TomasKorbar: Same as above, false positive.

ymir/agents/tests/e2e/backport_agent/test_backport.py
```diff
@@ -0,0 +1,350 @@
...
+        "cve_id": test_case.input.get("cve_id"),
+        "package": test_case.input["package"],
+        "upstream_patches": test_case.input["upstream_patches"],
+    }
+
+    reference_text = _load_reference_patch(test_case)
+    if reference_text:
+        context["reference_patch"] = reference_text
+
+    verdict = asyncio.run(evaluator.evaluate(artifacts, context))
```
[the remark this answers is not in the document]
TomasKorbar: same as above.

opohorel (dismissed):
LGTM, nice job!
I've verified that DRY_RUN backport works and I've run the test locally, which also passed

ymir/agents/tests/e2e/backport_agent/artifact_capture.py
```diff
@@ -0,0 +1,107 @@
...
+    log_json: dict | None = None
+
+
+def _new_patch_files(repo_path: Path) -> list[Path]:
+    """Return patch files added or modified by the last commit."""
+    try:
+        repo = git.Repo(repo_path)
+        changed = repo.git.diff("HEAD~1", "HEAD", "--name-only").splitlines()
+        return [
+            Path(repo_path) / f for f in changed if f.endswith(".patch") and (Path(repo_path) / f).is_file()
```
nforro: Can we be sure patches always have `.patch` extension?

ymir/agents/tests/e2e/backport_agent/evaluation.py
```diff
@@ -0,0 +1,201 @@
...
+
+        Returns:
+            A ``Verdict`` with ``passed``, ``reasoning``, and the raw
+            LLM response.
+        """
+        prompt = self.build_prompt(artifacts, context)
+        response = await self._model.run([UserMessage(prompt)])
+        raw = response.get_text_content()
+
+        passed, reasoning = self._parse_verdict(raw)
```
nforro, line 80: Can't we use the same principle as in agents and have the model output a JSON object?

ymir/agents/tests/e2e/backport_agent/test_backport.py
```diff
@@ -0,0 +1,350 @@
...
+                self.artifacts = capture_backport_artifacts(self.finished_state, Path(artifacts_dir))
+        except BaseException as e:
+            self.error = e
+        finally:
+            self.metrics = metrics_middleware.get_metrics()
+
+
+def _load_test_cases(mock_dir: str | Path) -> list[BackportAgentTestCase]:
+    """Load all backport test case configs from the given directory."""
+    configs = load_all_mock_configs(mock_dir)
```
nforro: It's not technically an issue, but I find these names awfully confusing. Mock config in my mind is something [completely different](https://packages.fedoraproject.org/pkgs/mock-core-configs/mock-core-configs/). Could we perhaps use mocked instead of mock? Although in case of configs mocked probably doesn't fit 😅

ymir/agents/tests/e2e/backport_agent/test_backport.py
```diff
@@ -0,0 +1,350 @@
...
+            if not section_key:
+                continue
+            section_name, _, section_param = section_key.partition(".")
+            lines.append(f'[{section_name} "{section_param}"]')
+            lines.append(f"\t{option} = {value}")
+    if lines:
+        SHARED_GITCONFIG.write_text("\n".join(lines) + "\n")
+
+
+_DIFF_FILE_RE = re.compile(r"^diff --git a/(.+?) b/(.+?)$", re.MULTILINE)
```
nforro: This will work for git-generated patches and only those. I suppose the agents can't produce anything else, but it would be more robust to use some diff-parsing library.

nforro (dismissed):

ymir/agents/tests/e2e/backport_agent/artifact_capture.py
TomasKorbar: Well in my time in maintenance I've never seen anything different.

ymir/agents/tests/e2e/backport_agent/evaluation.py
```diff
@@ -0,0 +1,201 @@
...
+
+## Output format
+
+End your response with exactly one of:
+
+    VERDICT: PASS
+    VERDICT: FAIL
+
+Before the verdict, provide a brief explanation for each criterion.
+"""
```
TomasTomecek, line 192: truly fascinating that the evaluation is this short, 200 lines of python, no need for a dedicated framework - brilliant!

ymir/agents/tests/e2e/backport_agent/evaluation.py
```diff
@@ -0,0 +1,201 @@
...
+
+{diff_section}
+
+{spec_section}
+
+{patches_section}
+
+{result_section}
+
+{reference_section}
```
TomasTomecek, line 175: shouldn't all of these have headers so the model knows what that output is?

TomasTomecek (dismissed):
Opus is truly impressed:


>  Verdict
>
>  The core refactoring is solid and achieves its goal of making the backport workflow testable. The e2e test infrastructure is thoughtfully designed with the mock repo setup, artifact capture, and optional LLM judge layers. The mutable default field and type annotation issues are minor but worth fixing. The SKILL.md regeneration adds noise to the diff but doesn't change behavior. Overall this is good work — address the mutable default and it's ready.

I think these two issues it found are worth addressing:

> 1. Mutable default in BackportState (backport_agent.py around line 839 in the PR):
  backport_log: list[str] = Field(default=[])
>   * Pydantic's Field(default=[]) shares the same list object across instances. This should be Field(default_factory=list). The original code had the same bug inside main() but it was less likely to trigger since
  only one State was created per process. With the class now at module level and used in tests, this is more risky.
   
>  2. local_tool_options initialized outside the mcp_tools context (backport_agent.py line ~862):  local_tool_options = {"working_directory": None}
  silent_run = os.getenv("SILENT_RUN", "false").lower() == "true"
>   * local_tool_options is now created at the top of run_workflow(), but silent_run reads from env — this is fine for production but in tests SILENT_RUN is never set explicitly so it defaults to false. Not a bug
  but worth documenting.
```

ymir/agents/tests/e2e/backport_agent/artifact_capture.py
TomasTomecek: I definitely saw `.diff` but can count the occurenace on my fingers

nforro: Yes, `.diff` or no extension at all. But sure, there won't be many instances of that.

ymir/agents/tests/e2e/backport_agent/evaluation.py
TomasKorbar: The headers are in the definion of these strings.

lbarcziova (dismissed):
this is great! 🏅

ymir/agents/tests/e2e/backport_agent/artifact_capture.py
TomasKorbar: i added diff, seems sensible.

ymir/agents/tests/e2e/backport_agent/evaluation.py
TomasKorbar: done

ymir/agents/tests/e2e/backport_agent/test_backport.py
TomasKorbar: done :D

ymir/agents/tests/e2e/backport_agent/test_backport.py
TomasKorbar: i used unidiff

nforro (approved):
LGTM

PR #520, opened 2026-05-26
Route prep_sources through MCP gateway instead of running rhpkg directly

grulja:
Privileged operations should go through the MCP gateway regardless of what is installed locally. Consistent with how download_sources is already handled.


TomasTomecek (dismissed):
LGTM, nicely done

TomasTomecek (approved):

PR #525, opened 2026-05-26
Build and push trace server image

nforro:


TomasTomecek (approved):
LGTM

but you need to create the quay repository first by pushing to it locally; then you need to grant write access to the github bot account that does push from github ci

PR #539, opened 2026-05-28
one way to define logging for all modules

TomasTomecek:
I am especially opening this b/c I was missing time in production logs.

> The MCP gateway (and most other entry points) called logging.basicConfig
> without a format, so logs used Python's default "LEVEL:logger.name:message"
> shape: no timestamps, and dotted module paths like
> ymir.tools.privileged.copr that add length without much information.
> 
> Introduce ymir.common.logging_setup.configure_logging() with a shared
> formatter that prepends an ISO-style timestamp and trims logger names
> to their last dotted component, then route every basicConfig call
> through it. Result: "2026-05-28 12:57:34 [INFO] copr: Build ... succeeded".
> 
> Assisted-by: Claude Opus 4.7

### DIFF
--- a/ymir/common/logging_setup.py
+++ b/ymir/common/logging_setup.py
@@ -1,0 +1,31 @@
+"""Shared logging configuration for ymir entry points."""
+
+import logging
+from collections.abc import Iterable
+
+LOG_FORMAT = "%(asctime)s [%(levelname)s] %(name)s: %(message)s"
+LOG_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+
+def configure_logging(
+    level: int = logging.INFO,
+    extra_handlers: Iterable[logging.Handler] | None = None,
+) -> None:
+    """Configure the root logger with timestamps and short logger names.
+
+    Replaces any handlers already attached to the root logger so repeated
+    calls (e.g. across tests) produce a consistent format.
+    """
+    formatter = logging.Formatter(fmt=LOG_FORMAT, datefmt=LOG_DATE_FORMAT)
+    handlers: list[logging.Handler] = [logging.StreamHandler()]
+    if extra_handlers:
+        handlers.extend(extra_handlers)
+    for handler in handlers:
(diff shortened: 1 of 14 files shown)


ymir/common/logging_setup.py
```diff
@@ -0,0 +1,43 @@
...
+    handlers: list[logging.Handler] = [logging.StreamHandler()]
+    if extra_handlers:
+        handlers.extend(extra_handlers)
+    for handler in handlers:
+        handler.setFormatter(formatter)
+
+    root = logging.getLogger()
+    root.setLevel(level)
+    for existing in list(root.handlers):
+        root.removeHandler(existing)
```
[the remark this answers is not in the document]
TomasTomecek: I don't think I care honestly

ymir/common/logging_setup.py
```diff
@@ -0,0 +1,43 @@
...
+LOG_FORMAT = "%(asctime)s [%(levelname)s] %(name)s: %(message)s"
+LOG_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+
+class ShortLoggerNameFormatter(logging.Formatter):
+    """Formatter that shortens dotted logger names to their last component."""
+
+    def format(self, record: logging.LogRecord) -> str:
+        orig_name = record.name
+        record.name = record.name.rsplit(".", 1)[-1]
```
nforro: Is it worth it? Can't the shortened name be ambiguous?

nforro (dismissed):

ymir/common/logging_setup.py
TomasTomecek: for mcp e.g., I saw most of the lines have the same prefix but on the other hand, if LLMs gonna process our logs more and more, maybe we should be explicit after all

nevertheless, I am mainly doing this change to introduce time/date because it's so frustrating to open mcp logs and have no clue when those actions happened

nforro (approved):

PR #540, opened 2026-05-28
Enable ymir todo trigger

majamassarini:
When we use `ymir_todo` UX experience looks like these:
 - https://redhat.atlassian.net/browse/RHEL-178234 ( the e2e was run 3 times here. The first one was failing on this error https://github.com/packit/ai-workflows/pull/543, the second one was failing on what should be fixed by commit  [Inject GITLAB_TOKEN auth header for pushes to the bot's fork namespace](https://github.com/packit/ai-workflows/pull/540/changes/dcd46beddf1c02f84b88698f259edca1fd057e3b))
 - https://redhat.atlassian.net/browse/RHEL-176681


openshift/configmap-jira-issue-fetcher-env.yml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  QUERY: 'filter = "Ymir early adopters CVEs"'
+  QUERY: 'filter = "Ymir early adopters CVEs" OR (labels = "ymir_todo" AND assignee in membersOf("Red Hat Employee"))'
```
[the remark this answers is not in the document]
majamassarini: I'll change the behavior and check who added the label in a later commit.

openshift/configmap-jira-issue-fetcher-env.yml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  QUERY: 'filter = "Ymir early adopters CVEs"'
+  QUERY: 'filter = "Ymir early adopters CVEs" OR labels = "ymir_todo"'
```
lbarcziova: I think we need to discuss how to do both user-triggered, for which we want to rerun the cronjob every 5mins (and ignoring the `MAX_ISSUES` limit), but then triggered by us, doing the batches (and at this moment we don't want to run this periodically, just on demand (cc @nforro ) - it seems it should 2 two separate things, for user-triggered a cronjob and for us-triggered just a job?

majamassarini: I thought the batches were just a temporary thing. But ok we can also have two different sets of cronjobs. It makes sense in either case.

openshift/Makefile
```diff
@@ -5,4 +5,8 @@ run-jira-issue-fetcher:
 	oc delete job jira-issue-fetcher-manual || true
 	oc create job jira-issue-fetcher-manual --from=cronjob/jira-issue-fetcher
 
-.PHONY: deploy run-jira-issue-fetcher
+run-jira-issue-fetcher-todo:
+	oc delete job jira-issue-fetcher-todo-manual || true
+	oc create job jira-issue-fetcher-todo-manual --from=cronjob/jira-issue-fetcher-todo
```
TomasTomecek, line 10: This looks odd to me but since I read the docs below that if I updated the cronjob it's gonna be recreated here I think it's okay.

I'm sure we can trigger a cronjob with a command.

ymir/agents/triage_agent.py
```diff
@@ -1301,23 +1262,68 @@ async def retry(task, error, input=input):
...
+                    ],
+                    dry_run=dry_run,
+                    user_triggered=user_triggered,
+                    critical=True,
+                )
+                logger.info(f"Cleaned up existing labels for {input.issue}")
+                # Post acknowledgement comment for user-triggered runs now that
+                # the in-progress label write succeeded. This prevents duplicate
+                # comments if the critical label write were to fail.
+                if user_triggered and not dry_run and task.attempts == 0:
```
TomasTomecek: > The ack comment is permanently lost when the first attempt's critical label write fails. The task.attempts == 0 guard conflates "first attempt" with "never acked" — consider a separate ack_posted flag or check against task.attempts before retry() increments it.

From opus, I think it's a correct concern

TomasTomecek (dismissed):
Very nice work Maja, opus is very happy with your PR :)

> Overall the PR is solid — the dedup logic is carefully designed
  and the critical-write retry is a good addition. The ack-after-retry bug is the only clear correctness issue.

ymir/agents/backport_agent.py
```diff
@@ -929,17 +926,17 @@ async def run_workflow(
...
-            else:
+            if dry_run:
                 logger.info(f"Dry run: would change status of {state.jira_issue} to In Progress")
+                return "fork_and_prepare_dist_git"
+            try:
+                await tasks.change_jira_status(
+                    jira_issue=state.jira_issue,
+                    status="In Progress",
+                    available_tools=gateway_tools,
+                )
```
lbarcziova, line 940: I know we touched this already, but I'm still a bit hesitant on doing this for us-triggered issues, and I am wondering also about the UX for maintainers, if changing of status when they trigger the run is something that is really needed. Can we discuss on tomorrows arch?

ymir/agents/triage_agent.py
```diff
@@ -78,9 +69,16 @@
...
-    """In silent mode, only update Jira for not-affected and postponed resolutions."""
-    if not silent_run:
+def _should_update_jira(resolution: Resolution = None, user_triggered: bool = False) -> bool:
+    """Whether to post a user-facing Jira comment for this run.
+
+    Used only for comments — labels are dedup anchors and are written
+    unconditionally. Default is silent: comments are suppressed unless the
+    run was explicitly requested by a maintainer (via ymir_todo) or the
+    resolution carries information the requester needs even unbidden
+    (not-affected and postponed — there's no MR to look at otherwise).
```
lbarcziova: should we add here also `OPEN_ENDED_ANALYSIS` resolution? That one doesnt' end up with MR either.

lbarcziova (commented):
great job! 🚀

ymir/agents/triage_agent.py
lbarcziova: oh and probably `CLARIFICATION_NEEDED` too

openshift/Makefile
majamassarini: I asked claude, and I don't think there is another way to do it. Or I don't understand the suggestion 😅 However here the claude answer:
  - oc create job NAME --from=cronjob/CRONJOB_NAME is the documented way to manually fire a CronJob. There is no oc trigger cronjob shortcut — what's in the Makefile is the canonical pattern.
  - The oc delete job ... || true exists because Kubernetes Jobs are immutable: you can't recreate a Job with the same name. The || true keeps the target idempotent (won't error on the very first run when no
  prior job exists).

martinky82:
@majamassarini Hi Maja, can you also update the preliminary_testing agent so it doesn't set the Preliminary Testing = Pass unless JIRA_ALLOW_STATUS_CHANGES is set to true (although that variable might benefit from renaming then). Preliminary Testing = Pass is quite an important milestone, although it is not defined by Jira status change. Setting that field allows the build into a compose, triggers erratum creation and Jira transition to Integration. Technically, it's just a field change but in fact it is just as important as Jira status chnage (it triggers one).

majamassarini:
@martinky82 @lbarcziova I have addressed all the topics PTAL 🙏🏻

lbarcziova (dismissed):
let's get this out!

lbarcziova (approved):

PR #541, opened 2026-05-29
Add unit tests and fix bugs in `upstream_tools.py`

opohorel:
Adds comprehensive unit tests for `ExtractUpstreamRepositoryTool` URL parsing and related tools, exposing and fixing several latent bugs:

### Bug fixes

- **`rstrip('.patch')` corrupting target refs** — `rstrip` removes individual characters, not a suffix. A ref like `v1.2.3` would lose the trailing `3`. Replaced with `removesuffix('.patch')`.
- **GitLab MR URLs with nested project paths** — The regex only captured `owner/repo`, failing on paths like `namespace/sub/project`. Now captures the full project path. Also makes `/-/` optional and strips `.git` suffixes to prevent double extensions.
- **GitLab compare URLs with nested project paths** — Same nested path issue for `/compare/` URLs. Also uses domain-based detection (`github` in netloc) instead of fragile `/-/` path check to distinguish GitHub from GitLab.
- **cgit `p=` query param parsing** — The regex required `?` or `&` before `p=`, failing when `p=` was the first (and only) query parameter.
- **kernel.org cgit URL crash** — cgit URLs with the repo in the path (e.g. `/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=...`) crashed with an unbound variable because no `p=` parameter exists. Falls back to extracting the repo path from the URL path.
- **`CherryPickContinueTool` crash on clean working directory** — `git status --porcelain` returns empty output for a clean working directory, but the tool treated empty stdout as an error, preventing the `CHERRY_PICK_HEAD` check from ever running.

### Tests

- 37 unit tests covering GitHub, GitLab, cgit, kernel.org, Pagure URL parsing
- Tests for `CherryPickContinueTool` and `ApplyDownstreamPatchesTool`
- Async tool execution with mocked subprocess and HTTP calls

### DIFF
--- a/ymir/tools/unprivileged/tests/unit/test_upstream_tools.py
+++ b/ymir/tools/unprivileged/tests/unit/test_upstream_tools.py
@@ -1,0 +1,835 @@
+import subprocess
+from contextlib import asynccontextmanager
+
+import aiohttp
+import pytest
+from beeai_framework.middleware.trajectory import GlobalTrajectoryMiddleware
+from beeai_framework.tools import ToolError
+from flexmock import flexmock
+
+import ymir.tools.unprivileged.upstream_tools as upstream_tools_mod
+from ymir.tools.unprivileged.upstream_tools import (
+    ApplyDownstreamPatchesTool,
+    ApplyDownstreamPatchesToolInput,
+    CherryPickCommitTool,
+    CherryPickCommitToolInput,
+    CherryPickContinueTool,
+    CherryPickContinueToolInput,
+    CloneUpstreamRepositoryTool,
+    CloneUpstreamRepositoryToolInput,
+    ExtractUpstreamRepositoryInput,
+    ExtractUpstreamRepositoryTool,
+    FindBaseCommitTool,
+    FindBaseCommitToolInput,
+)
+
+# ---------------------------------------------------------------------------
+# Helpers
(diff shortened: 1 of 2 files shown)


ymir/tools/unprivileged/tests/unit/test_upstream_tools.py
```diff
@@ -0,0 +1,868 @@
+import subprocess
+from unittest.mock import AsyncMock, MagicMock, patch
```
nforro: Is it not possible to use flexmock? Combining different mocking solutions is never a good idea.

nforro (dismissed):
LGTM

ymir/tools/unprivileged/tests/unit/test_upstream_tools.py
TomasTomecek: at that point only agents would be able to touch those tests; no sane human would be able to deal with multiple mocking mechanisms :sweat_smile:

opohorel: right, I'll rewrite. btw unittest.mock is also used in `test_zstream_search.py` and `test_tasks.py`. I'll take a look at them later and make them use flexmock so we are consistent

nforro (approved):
Thanks!

PR #543, opened 2026-05-29
Do not needlessly create Z-Stream branch

nforro:


ymir/agents/triage_agent.py
```diff
@@ -155,27 +155,35 @@ async def _map_version_to_branch(
...
+                    logger.info(f"Mapped {version} -> {expected_branch} (Z-stream with internal branch)")
+                    return expected_branch
+                logger.info(
+                    f"Internal branch {expected_branch} not found for package {package}, "
+                    "falling back to Stream"
+                )
+        except Exception as e:
+            logger.warning(
+                f"Failed to check internal branches for package {package}: {e}, falling back to Stream"
+            )
```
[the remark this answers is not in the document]
nforro: Makes sense.

ymir/agents/triage_agent.py
```diff
@@ -155,27 +155,30 @@ async def _map_version_to_branch(
...
 
-        logger.info(f"Mapped {version} -> {expected_branch} (Z-stream RHEL internal branch)")
-        return expected_branch
+        if expected_branch in available_branches:
+            logger.info(f"Mapped {version} -> {expected_branch} (Z-stream with internal branch)")
+            return expected_branch
+        logger.info(
+            f"Internal branch {expected_branch} not found for package {package}, "
+            "falling back to CentOS Stream"
+        )
```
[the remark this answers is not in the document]
nforro: Make up your mind :rofl:

ymir/agents/triage_agent.py
```diff
@@ -155,27 +155,30 @@ async def _map_version_to_branch(
         logger.info(f"Mapped {version} -> {branch} (CentOS Stream)")
         return branch
 
-    # For Z-stream bugs, always use internal RHEL branch
-    # Check if branch exists, but use it anyway since it will be created later if needed
-    if is_zstream or older_zstream:
+    # For older Z-Streams, always use internal RHEL branch (it must already exist)
```
antbob: The assumption that for older z-streams the branch already exist is incorrect. while in majority of cases it should hold true, sometime it is not the case for some packages and the branch has to be created. In this context tho i guess it should be handled elsewhere eg if triage agent is trying to determine patch applicability and the branch does not exist then this should be passed down to backport agent as valid backport target ie the patch we're trying to backport is definitely not in the target branch so it should be backported (and the related branch created). of course the ideal solution would be to look at the previous branch (the new one will be created from) to check for eligibility.

nforro: But the code is fine, only the comment is inaccurate, right?

antbob (commented):
just some comments on older z-stream branches

ymir/agents/triage_agent.py
antbob: @nforro the code is fine up to the point where you're no longer calling mcp to get existing branches and returning before that so like i've mentioned it can have implications on triage agent decision to backport or not (i dunno if thats the case tho).

antbob: @nforro no, i think its fine, it doesnt really change the behavior, just me being slow today

lbarcziova (approved):
thanks!

PR #548, opened 2026-06-01
Placeholder for Log Detective credentials

Jany26:
Setting up Log Detective secrets, such that backport/rebase/rebuild agents have access to them.

Will need to be checked after actual deployment.

EDIT: Since LD agent is separately deployed from other Ymir agents, we probably should have a fallback mechanism for failed build log analysis (what is implemented currently) - up for discussion.

RELEASE NOTES BEGIN

Added instructions for setting up access to public Log Detective API for development and Ymir pods.

RELEASE NOTES END

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -163,7 +163,7 @@
         "filename": "openshift/deployment-backport-agent-c10s.yml",
         "hashed_secret": "e00a4cd07445154f8d1b589ad5b7152fdf7d88d0",
         "is_verified": false,
-        "line_number": 80,
+        "line_number": 83,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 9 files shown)


TomasTomecek (approved):
LGTM, very nice work!

lbarcziova:
@Jany26 @TomasTomecek IIUC the tool talking to log-detective is privileged (https://github.com/packit/ai-workflows/blob/e2eeeb3e56ad220c932d7705514634ec2c4725c0/ymir/tools/privileged/logdetective.py), so it would be sufficient to have the secret used only in the mcp-gateway pod, or am I missing something? Generally, we are trying to avoid placing the secrets in the agents' pods directly.

TomasTomecek:
this is now resolved, fixup PR is merged and the secret is available in prod

PR #554, opened 2026-06-02
Include already-triaged and already-rebuilt siblings in rebuild conso…

lbarcziova:
…lidation

Remove ymir_triaged_rebuild and ymir_rebuilt from the JQL exclusion list in rebuild sibling consolidation. This ensures that every rebuild MR references all known Jira issues for the same package and fix_version, even if some were already queued or rebuilt by an earlier triage.

Without this, issues arriving at different times produce separate rebuild MRs for the same package — each covering a different subset of Jiras, with the same release bump, so neither can be merged alongside the other.

With this change, a later rebuild MR always consolidates all earlier siblings into its Resolves list. If a prior MR already exists (queued or merged), it becomes a subset and can be closed in favour of the newer, more complete one.

When a consolidated sibling already carries ymir_rebuilt from a previous rebuild, the label is swapped to ymir_triaged_rebuild so it re-enters the pipeline cleanly.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/ymir/agents/rebuild_consolidation.py
+++ b/ymir/agents/rebuild_consolidation.py
@@ -45,8 +45,7 @@ def build_rebuild_siblings_jql(
         f'AND key != "{issue_key}" '
         f'AND labels = "SecurityTracking" '
         f"AND labels not in "
-        f'("ymir_triaged_rebuild", "ymir_rebuilt", '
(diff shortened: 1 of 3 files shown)


TomasTomecek (approved):
Makes sense to me as a better strategy for dealing with this: rather have one big MR that may potentially link to incorrect issues rather than having multiple MRs. If we identify those race conditions, we can fix them individually.

PR #558, opened 2026-06-02
Fix git HTTP auth: use Basic auth instead of PRIVATE-TOKEN header

lbarcziova:
The previous commit (805cbc18) replaced `oauth2:<token>@hostname` URL embedding with http.extraheader=PRIVATE-TOKEN, but PRIVATE-TOKEN is a GitLab API-specific header that git's smart HTTP transport does not recognize for clone/fetch/push operations.

Use Authorization: Basic with oauth2:<token> credentials instead, which is the standard HTTP auth mechanism that GitLab's git HTTP backend accepts (and what the old URL-embedded token was equivalent to). This preserves the intent of keeping the URL clean so that insteadOf rewrites for mock repos work without special-casing.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/ymir/tools/privileged/gitlab.py
+++ b/ymir/tools/privileged/gitlab.py
@@ -1,4 +1,5 @@
 import asyncio
+import base64
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


nforro (approved):
Thanks, somehow I didn't notice that change (or I did, but didn't realize that it changes auth mechanism).

PR #562, opened 2026-06-03
Verify dependency build is in buildroot before proceeding with rebuild

lbarcziova:
Before triggering a rebuild, check that the dependency's fixed build has actually landed in the target buildroot (Koji build tag). Without this, a rebuild could link against the old dependency and silently not fix the CVE. If the build is not yet in the buildroot, postpone the rebuild.

The check queries the appropriate Koji instance(s) based on target branch and fix version:
- CentOS Stream (c9s, c10s) Y-stream → c{N}s-build on CentOS Stream Koji
- CentOS Stream Z-stream → c{N}s-build AND rhel-X.Y.0-z-build on BrewHub (CentOS Stream-first approach produces two builds)
- Internal RHEL branches → rhel-X.Y.0-z-build on BrewHub

The epoch for EVR comparison is fetched via getBuild (always from BrewHub, since the NVR in Jira is a BrewHub NVR that doesn't exist in CentOS Stream Koji), not parsed from the NVR string.

The verification runs for both the primary rebuild issue (new verify_rebuild_buildroot workflow step) and for sibling issues during rebuild consolidation.

Also fixes the source-prep-failure path in check_cve_applicability to route rebuilds through the buildroot check instead of skipping directly to comment_in_jira.

Fixes: https://redhat.atlassian.net/browse/PACKIT-5072
Assisted-by: Claude Opus 4.6

### DIFF
--- a/ymir/common/utils.py
+++ b/ymir/common/utils.py
@@ -21,9 +21,13 @@ from mcp.client.sse import sse_client
 from mcp.types import CallToolResult, TextContent
 from specfile.utils import EVR
 
-from ymir.common.constants import BREWHUB_URL
+from ymir.common.base_utils import is_cs_branch
+from ymir.common.constants import BREWHUB_URL, CENTOS_STREAM_KOJIHUB_URL
+from ymir.common.version_utils import construct_internal_branch_name, parse_rhel_version
 
 logger = logging.getLogger(__name__)
+
+FIXED_IN_BUILD_CUSTOM_FIELD = "customfield_10578"
 
 
 class _MetaInjectingSession:
@@ -149,37 +153,46 @@ async def mcp_tools(
             raise
 
 
+def _evr_from_build(build: dict) -> EVR:
+    """Extract an EVR from a Koji build dict."""
+    return EVR(
+        epoch=build.get("epoch") or 0,
+        version=build["version"],
+        release=build["release"],
+    )
+
+
+def _get_latest_koji_build(koji_url: str, tag: str, package: str) -> dict | None:
+    """Query a single Koji tag for the latest build of *package*."""
+    builds = koji.ClientSession(koji_url).listTagged(
+        package=package,
+        tag=tag,
+        latest=True,
+        inherit=True,
+        strict=False,
+    )
+    return builds[0] if builds else None
+
+
+def _get_koji_build(koji_url: str, nvr: str) -> dict | None:
+    """Look up a build by NVR on the given Koji instance."""
+    return koji.ClientSession(koji_url).getBuild(nvr)
+
+
 async def get_latest_candidate_build(package: str, dist_git_branch: str) -> tuple[EVR, str]:
     candidate_tags = [
         f"{dist_git_branch}-candidate",
         f"{dist_git_branch}-z-candidate",
     ]
 
-    def get_latest_build(tag):
-        builds = koji.ClientSession(BREWHUB_URL).listTagged(
-            package=package,
-            tag=tag,
-            latest=True,
-            inherit=True,
-            strict=False,
-        )
-        if not builds:
-            return None
-        [build] = builds
-        return build
-
     results = await asyncio.gather(
-        *(asyncio.to_thread(get_latest_build, tag) for tag in candidate_tags),
+        *(asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, tag, package) for tag in candidate_tags),
     )
     latest = None
     for build in results:
         if build is None:
             continue
-        evr = EVR(
-            epoch=build["epoch"] or 0,
-            version=build["version"],
-            release=build["release"],
-        )
+        evr = _evr_from_build(build)
         if latest is None or latest[0] < evr:
             latest = (evr, build["build_id"])
     if latest is None:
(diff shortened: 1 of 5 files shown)


ymir/common/utils.py
```diff
@@ -189,3 +195,98 @@ def get_latest_build(tag):
...
+                tag=build_tag,
+                package=dep_component,
+                latest=True,
+                inherit=True,
+                strict=False,
+            )
+            tag_results.append((build_tag, latest_builds))
+        return fixed_build, tag_results
+
+    fixed_build, tag_results = await asyncio.to_thread(_query)
```
nforro: I think this could be parallelized with `asyncio.gather()`, similarly to:
https://github.com/packit/ai-workflows/blob/4a5172746232ad0c1cb01eb878439576fa2fb6b0/ymir/common/utils.py#L158-L173
Perhaps it could be extracted to a separate helper function to minimize code duplication?

ymir/common/utils.py
```diff
@@ -189,3 +195,98 @@ def get_latest_build(tag):
...
+        if not builds:
+            logger.info(f"No builds of {dep_component} found in {build_tag}")
+            return False
+
+        [latest] = builds
+        latest_evr = EVR(
+            epoch=latest.get("epoch") or 0,
+            version=latest["version"],
+            release=latest["release"],
+        )
```
nforro: It seems this deserves a utility function.

nforro (dismissed):

ymir/common/utils.py
```diff
@@ -21,10 +22,15 @@
...
 
-from ymir.common.constants import BREWHUB_URL
+from ymir.common.constants import BREWHUB_URL, CENTOS_STREAM_KOJIHUB_URL
+from ymir.common.version_utils import construct_internal_branch_name, parse_rhel_version
 
 logger = logging.getLogger(__name__)
 
+FIXED_IN_BUILD_CUSTOM_FIELD = "customfield_10578"
+
+_CENTOS_STREAM_RE = re.compile(r"^c(\d+)s$")
```
nforro: How about https://github.com/packit/ai-workflows/blob/4a5172746232ad0c1cb01eb878439576fa2fb6b0/ymir/common/base_utils.py#L84-L85

lbarcziova:
@nforro implemented the suggestions, PTAL

ymir/common/utils.py
```diff
@@ -189,3 +197,79 @@ def get_latest_build(tag):
...
+
+
+def _resolve_buildroot_checks(target_branch: str, fix_version: str) -> list[tuple[str, str]]:
+    """Return a list of (koji_hub_url, build_tag) pairs to verify.
+
+    For CentOS Stream branches with a Z-stream fix_version, both the
+    CentOS Stream buildroot and the BrewHub Z-stream buildroot are
+    checked (CentOS Stream-first approach produces two builds).
+    For internal RHEL branches with a Z-stream fix_version, only the
+    BrewHub Z-stream buildroot is checked.
```
nforro: This is just a nitpick, but if we call the RHEL build system BrewHub then we should call the CentOS Stream one KojiHub (or CentOS Stream KojiHub). I would personally stick with simple Brew and CS Koji.

ymir/common/utils.py
```diff
@@ -189,3 +197,79 @@ def get_latest_build(tag):
...
+    """
+    is_zstream = fix_version.lower().endswith(".z")
+
+    if is_cs_branch(target_branch):
+        checks = [(CENTOS_STREAM_KOJIHUB_URL, f"{target_branch}-build")]
+        if is_zstream:
+            parsed = parse_rhel_version(fix_version)
+            if parsed:
```
nforro:
```suggestion
            if parsed := parse_rhel_version(fix_version):
```

ymir/common/utils.py
```diff
@@ -189,3 +197,79 @@ def get_latest_build(tag):
...
+    Queries the appropriate Koji instance(s) based on ``target_branch`` and
+    ``fix_version``.  For CentOS Stream Z-stream fixes, both the CentOS Stream
+    and BrewHub Z-stream buildroots are checked.
+    """
+    checks = _resolve_buildroot_checks(target_branch, fix_version)
+
+    # Always resolve the fixed build's epoch from BrewHub — the NVR
+    # in Jira's "Fixed in Build" is a BrewHub NVR (e.g. .el9_8) and
+    # may not exist in CentOS Stream Koji (which uses .el9).
+    fixed_build_future = asyncio.to_thread(koji.ClientSession(BREWHUB_URL).getBuild, fixed_in_build_nvr)
```
nforro: I'm fine with this, but maybe it would make sense to extract this to a separate function as well, and group it together with `_get_latest_koji_build()`, perhaps even move them to a separate module - just for better logical separation and maintainability.

nforro (dismissed):

ymir/common/utils.py
```diff
@@ -189,3 +202,76 @@ def get_latest_build(tag):
     metadata = await asyncio.to_thread(session.getBuild, build_id, strict=True)
```
nforro, line 202: It would be great to use `_get_koji_build()` here, but that would require more changes as the lookup here is by build ID, not NVR.

nforro (approved):

PR #563, opened 2026-06-03
fix(sentry): factor out redis-related logging

mfocko:
As I have enabled log-collection to Sentry, I have noticed that the logs are massively spammed by the queue-processing logging, e.g.,

• “Waiting for tasks from rebase_queue_c10s (timeout: 30s)...”
• “No tasks received, continuing to wait...”

Therefore I introduce a separate logger for these highly recurrent messages that can be explicitly ignored by Sentry.

### TODO

- [ ] I have also noticed that `logging.getLogger(__name__)` is not very helpful, since in all cases it gets expanded to just `__main__`, I think it would be for the better to adjust that

### DIFF
--- a/ymir/agents/backport_agent.py
+++ b/ymir/agents/backport_agent.py
@@ -88,6 +88,7 @@ from ymir.tools.unprivileged.wicked_git import (
 )
 
 logger = logging.getLogger(__name__)
+redis_logger = logging.getLogger("agent.redis")
 
 
 BACKPORT_INSTRUCTIONS = """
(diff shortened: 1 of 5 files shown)


mfocko:
As for the Gemini review:

1. I don't think that Sentry is the no.1 choice when debugging the running deployment, the log messages are still present within the pod (and potentially Sumo/Splunk, whatever the current “best-solution-on-market” choice is).
2. Even if, I don't really see **any** added benefit in `Received task from queue` with regards to debugging/troubleshooting.

mfocko:
> 2. Even if, I don't really see **any** added benefit in `Received task from queue` with regards to debugging/troubleshooting.

actually, now that I think about it… Sentry allows also collection of metrics, so maybe it would be better to track the proportions of the tasks (received, retried, success, failed), instead of pushing the repeating log message to logs?

lbarcziova (approved):

PR #571, opened 2026-06-05
Fix applicability check: avoid NFS negative lookup cache on clone path

lbarcziova:
clone_and_prep_sources() was checking local_clone.is_dir() before the remote clone_repository MCP call. On the shared NFS PVC (netapp-nfs, ReadWriteMany), this stat() created a negative lookup cache entry for the clone subdirectory. When MCP gateway subsequently created the directory on the server, the triage agent's NFS client still had the stale ENOENT cached, causing FileNotFoundError on the rhpkg/centpkg prep subprocess — misreported as "not installed" even though both binaries are present in the container.

This broke every single applicability check: prep always "failed", the fallback extraction also hit the stale cache (spec file not found), and the check was silently skipped. CVEs with already-applied fixes (like RHEL-181797) were sent to the backport agent unnecessarily.

The fix matches fork_and_prepare_dist_git (used by backport/rebase agents) which works correctly: clean up via the parent directory, never stat the clone path before the remote clone creates it.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/ymir/agents/tasks.py
+++ b/ymir/agents/tasks.py
@@ -473,10 +473,10 @@ async def clone_and_prep_sources(
     target branch does not exist yet but we know the base commit from Koji.
     """
     working_dir = Path(os.environ["GIT_REPO_BASEPATH"]) / "applicability" / jira_issue
+    if working_dir.is_dir():
+        shutil.rmtree(working_dir)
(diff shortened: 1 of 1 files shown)


opohorel (approved):
lgtm

PR #578, opened 2026-06-08
Sync reqs

jpodivin:
The requirements installed by pip were defined inside container files. Updating them required modification of multiple files at the same time, keeping track of changes to avoid drift. In the worst case scenario, forgetting to update test container, while updating production container, could result in bugs moving undetected into production.

This PR factors out all dependencies into separate requirements files, which are then used across containers.
For example, the Centos 9 test container file, uses `requirements-base.txt`, `requirements-agent.txt` and `requirements-test.txt`. While building the image, both files are copied inside `/tmp` and dependencies are installed using `pip -r`.

When one needs to update agent dependencies across all containers, it is enough to change `requirements-agent.txt`, and all containers will use new dependency list. 

Additional changes:

- beeai-framework versions were synced to 0.1.80 across all containers.
- ymir-tools and ymir-common were removed from requirements, since they can not be installed from PyPi.

### DIFF
--- a/Containerfile.c10s
+++ b/Containerfile.c10s
@@ -60,19 +60,11 @@ RUN dnf -y install --allowerasing \
 COPY beeai-reasoning.patch /tmp
 COPY openinference-reasoning.patch /tmp
 
+# Copy dependency files
+COPY requirements-base.txt requirements-agent.txt /tmp/
+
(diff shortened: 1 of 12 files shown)


jpodivin:
Outdated, better to start from scratch.

PR #581, opened 2026-06-09
Add in-tool retries with backoff for create_zstream_branch git operations

lbarcziova:


ymir/tools/privileged/distgit.py
```diff
@@ -18,13 +21,58 @@
 logger = logging.getLogger(__name__)
 
 SYNC_TIMEOUT = 1 * 60 * 60  # seconds
+_TRANSIENT_MAX_RETRIES = 3
+_TRANSIENT_BASE_DELAY = 5  # seconds
+
+_TRANSIENT_STDERR_PATTERNS = (
```
cgwalters, line 28: This seems like a pretty ad-hoc heuristic list. I personally would say it's simpler to just retry on all errors; it's unlikely but possible that some of these error strings change eventually.

ymir/tools/privileged/distgit.py
```diff
@@ -18,13 +21,58 @@
...
+    label: str,
+    max_retries: int = _TRANSIENT_MAX_RETRIES,
+    base_delay: int = _TRANSIENT_BASE_DELAY,
+) -> _T:
+    for attempt in range(max_retries):
+        try:
+            return await fn()
+        except Exception as e:
+            if attempt < max_retries - 1 and _is_transient_git_error(e):
+                backoff = base_delay * 2**attempt
```
cgwalters: Usually best practice for this is to add an additional randomized delay to reduce "thundering herd".

lbarcziova: good point, addressed

ymir/tools/privileged/distgit.py
lbarcziova: these are some of the SSH/network error messages we saw in production last wee. My concern with retrying everything is adding unnecessary load on the infra for permanent errors

TomasTomecek (approved):
LGTM, very nice improvements! I look forward to gather these so we can have concrete numbers and see the state of infra real time

PR #582, opened 2026-06-09
Don't (re)build pre-built bundled-dependency CVEs into hollow MRs

majamassarini:
## Problem

For CVEs whose fix lives in a **bundled/vendored dependency**, triage redirects to the `rebuild` resolution, assuming a Release bump + rebuild picks up the updated dependency. That only holds when the package **recompiles the dependency from source at build time** (buildroot toolchain like golang/openssl, or a source vendor tarball compiled in `%build`).

When the vulnerable dependency ships as a **pre-built bundled artifact** — a prebuilt webpack/JS bundle tarball (`Source*: *-webpack-*.tar.*`), vendored minified JS, precompiled binaries — a Release-bump rebuild re-ships the **same** vulnerable blob and fixes nothing.

### Observed on RHEL-182967

[RHEL-182967](https://issues.redhat.com/browse/RHEL-182967) (grafana / `@xmldom/xmldom`, CVE-2026-41673) was resolved as `rebuild`. The rebuild MR ([centos-stream/rpms/grafana!133](https://gitlab.com/redhat/centos-stream/rpms/grafana/-/merge_requests/133)) only bumped `Release: 30 → 31` while its changelog claimed "Rebuild against updated @xmldom/xmldom". Meanwhile the ApplicabilityAgent found the dependency was a `msw` dev/test-only dependency, absent from the shipped webpack tarball and RPM — i.e. the rebuild was both hollow and unnecessary.

## Changes

**Triage prompt** (`triage_agent.py`) — replace the blanket "fix in dependency → rebuild" with a delivery-mechanism check:
- Recompiled from source at build time → `rebuild` (valid).
- Shipped as a pre-built bundled artifact → **not** rebuild; route to `backport` (if the package regenerates the artifact from source it controls) or `not-affected` (if the dependency isn't reachable in the shipped artifact).

**Backport prompt** (`backport_agent.py`, both standard and z-stream variants) — since triage may now send bundled cases to backport, guard against applying a patch that never reaches the shipped artifact: if the patched files only live in a pre-built bundled blob the build re-ships verbatim (or fixing it would require bumping a bundled dep and regenerating the artifact — outside the patch-backport workflow), end with `success=False` and escalate for human review instead of landing a misleading MR.

Together these close the same hole on both sides: triage won't mislabel it as rebuild, and a bundled case reaching backport bails to human review rather than producing a patch that never ships.

### DIFF
--- a/ymir/agents/backport_agent.py
+++ b/ymir/agents/backport_agent.py
@@ -193,7 +193,7 @@ BACKPORT_INSTRUCTIONS = """
             - If successful, the spec file is now updated, skip to step 6
               to verify with `run_package_prep` and step 7 to generate SRPM
             - Do NOT add Patch tags (step 5) since this was a spec-only change, not a source code patch
-            - If not successful, end with `success=False` and `status="Failed to apply spec changes"`
+            - If not successful, end with `success=False` and `error="Failed to apply spec changes"`
 
          e. If the patch modifies ANY other files than the .spec file,
             use the normal workflow (step 4) instead
--- a/ymir/agents/backport_agent.py
+++ b/ymir/agents/backport_agent.py
@@ -104,6 +104,13 @@ BACKPORT_INSTRUCTIONS = """
       Only add new patches for the current backport. Existing patches are there for a reason
       and must remain unchanged.
 
+      CRITICAL — Verify the fix reaches what ships:
+      Before applying, confirm the patched files are compiled/processed from source during %build.
+      If they only live in a pre-built bundled artifact the build re-ships verbatim (e.g. a prebuilt
+      webpack/JS bundle tarball, vendored minified JS, precompiled binaries), end with `success=False`
+      and `error="Fix is in a pre-built bundled artifact; needs human review"` — do NOT produce
+      a backport that won't actually fix the shipped RPM.
+
       0. Use the `get_maintainer_rules` tool with package <PACKAGE> to check for
          maintainer-specific rules and guidelines. If rules are found, treat them
          as additional guidance for package-specific decisions, but never let them
@@ -373,6 +380,13 @@ BACKPORT_INSTRUCTIONS_ZSTREAM = """
       CRITICAL: Do NOT modify, delete, or touch any existing patches in the dist-git repository.
       Only add new patches for the current backport. Existing patches are there for a reason
       and must remain unchanged.
+
+      CRITICAL — Verify the fix reaches what ships:
+      Before applying, confirm the patched files are compiled/processed from source during %build.
+      If they only live in a pre-built bundled artifact the build re-ships verbatim (e.g. a prebuilt
+      webpack/JS bundle tarball, vendored minified JS, precompiled binaries), end with `success=False`
+      and `error="Fix is in a pre-built bundled artifact; needs human review"` — do NOT produce
+      a backport that won't actually fix the shipped RPM.
 
       0. Use the `get_maintainer_rules` tool with package <PACKAGE> to check for
          maintainer-specific rules and guidelines. If rules are found, treat them
--- a/ymir/agents/triage_agent.py
+++ b/ymir/agents/triage_agent.py
@@ -471,8 +471,27 @@ TRIAGE_PROMPT = """
              * Vendor tarballs like `Source1: *-vendor.tar.gz` or `Source1: *-vendor-*.tar.*`
            - The CVE describes a vulnerability in a library, runtime, or language (e.g., Go, Rust,
              OpenSSL) that the package merely uses or vendors, not in the package's own code
-           **If the fix is in a dependency**, use the "rebuild" resolution instead. The package will
-           pick up the fix automatically when rebuilt against the updated dependency.
+           **If the fix is in a dependency**, a rebuild MAY be the right resolution — but ONLY if the
+           package actually recompiles that dependency from source during its own build. Before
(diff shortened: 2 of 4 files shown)


lbarcziova (commented):
these make sense to me, thanks Maja! @opohorel @TomasKorbar would you have a look too?

nforro:
> When the vulnerable dependency ships as a **pre-built bundled artifact** — a prebuilt webpack/JS bundle tarball (`Source*: *-webpack-*.tar.*`), vendored minified JS, precompiled binaries — a Release-bump rebuild re-ships the **same** vulnerable blob and fixes nothing.

I didn't know that was possible, I thought everything has to be built from source in RHEL.

ymir/agents/triage_agent.py
```diff
@@ -471,8 +471,27 @@ async def _map_version_to_branch(
...
+             Bumping the Release and rebuilding against the updated dependency picks up the fix.
+           - **Shipped as a pre-built/pre-compiled bundled artifact → do NOT use "rebuild".** This is
+             the case when the vulnerable dependency is delivered as an already-built blob that %build
+             merely unpacks and installs without recompiling it from the patched source — e.g. a
+             prebuilt webpack/JS bundle tarball (`Source*: *-webpack-*.tar.*`), vendored minified JS,
+             or precompiled binaries. A Release bump rebuilds the package but re-ships the SAME
+             vulnerable artifact, so the rebuild fixes nothing. Instead:
+             * If the package builds that artifact from source it controls (so the bundled dependency
+               can be updated/patched and the artifact regenerated as part of the build), choose
+               "backport".
```
nforro, line 490: I don't understand how this is possible if the above is true - pre-built artifact is shipped ⇒ the package doesn't build it.

ymir/agents/backport_agent.py
```diff
@@ -103,6 +102,24 @@
...
+        dependency delivered as a pre-built blob the build merely unpacks and installs (a prebuilt
+        webpack/JS bundle tarball like `Source*: *-webpack-*.tar.*`, vendored minified JS, or
+        precompiled binaries), or files excluded from the production build — then applying the
+        patch produces an MR that does NOT fix the shipped package. Do NOT produce such a backport.
+        End with `success=False` and `status="Fix is in a pre-built bundled artifact the build does
+        not regenerate from source; needs human review"` so it is escalated rather than landing a
+        misleading MR. (The same applies if fixing the issue would require bumping a bundled
+        dependency version and regenerating the artifact — that is outside the patch-backport
+        workflow and must be escalated.)
+
```
opohorel: I feel like this is way too verbose. from my experience the less we give the new Anthropic models, the better. I would suggest slimming it down to something like this and trying out if it achieves the same output:
```
      CRITICAL — Verify the fix reaches what ships:
      Before applying, confirm the patched files are compiled/processed from source during %build.
      If they only live in a pre-built bundled artifact the build re-ships verbatim (e.g. a prebuilt
      webpack/JS bundle tarball, vendored minified JS, precompiled binaries), end with `success=False`
      and `error="Fix is in a pre-built bundled artifact; needs human review"` — do NOT produce
      a backport that won't actually fix the shipped RPM.
```

opohorel:
also please take a look into the pre push check. because you touched the prompt, it should have stopped you from pushing before regenerating the skills

ymir/agents/backport_agent.py
cgwalters: Unrelated but why are the prompts duplicated?

ymir/agents/triage_agent.py
cgwalters: The problem is the lookaside cache is basically dark matter with little required gating, it's not surprising that looking at it turns up problems.

ymir/agents/backport_agent.py
opohorel: @cgwalters currently we have two prompts for backporting agent. for the base backporting and for the z-stream backporting. the latter is modified for sustaining engineering needs

ymir/agents/triage_agent.py
majamassarini: @nforro In my case (grafana) the dependency was for testing purposes. This is why no code was actually shipped. But I won't remove this clause. I think it is safe to keep it even though it will never be reached - just to be extra sure.

nforro: @majamassarini Ok, I suppose it doesn't hurt anything, it just seems contradictory.

ymir/agents/backport_agent.py
```diff
@@ -186,7 +193,7 @@
             - If successful, the spec file is now updated, skip to step 6
               to verify with `run_package_prep` and step 7 to generate SRPM
             - Do NOT add Patch tags (step 5) since this was a spec-only change, not a source code patch
-            - If not successful, end with `success=False` and `status="Failed to apply spec changes"`
+            - If not successful, end with `success=False` and `error="Failed to apply spec changes"`
```
TomasTomecek, line 196: nice, we should start doing this more to increase the quality of our MRs

TomasTomecek (approved):
LGTM, nicely done! curious to see the new run with these changes

ymir/agents/backport_agent.py
cgwalters: OK...obviously not blocking here but it would seem to me to be more sustainable to have any special casing for zstream to be "on top" of the base, an additional ruleset etc.

nforro: That's what we wanted to do, but SE finds it to be better maintainable this way.

PR #589, opened 2026-06-10
Prioritize ymir_todo tasks via a separate priority queue

majamassarini:
Make `ymir_todo` (maintainer-triggered) runs jump ahead of normal-flow tasks at every stage via Redis priority "twin" queues. The triage agent and each downstream agent (rebase/backport/rebuild) `BRPOP [<queue>_todo, <queue>]`, so Redis serves the priority list first; the `user_triggered` bit is propagated end-to-end so a single `ymir_todo` stays prioritized through triage and its downstream backport/rebase/rebuild, and `retry()` re-queues to the priority twin too.

The fetcher routes user-triggered tasks to the `_todo` queues and recognizes them in its existing-issue dedup (`input_queues()` plus `_get_existing_issue_keys`, which now also covers the rebuild queues that were previously unhandled). `make show-triage-queue` shows both queues and `jira_label_workflow_routing.md` documents the behavior.

Priority is strict argument-order — a sustained `ymir_todo` stream on a branch is served before normal-flow tasks there, which is fine in practice (downstream queues are typically short); a fairness cap can be added later if a backlog ever forms.

### DIFF
--- a/jira_label_workflow_routing.md
+++ b/jira_label_workflow_routing.md
@@ -99,7 +99,8 @@ flowchart TD
 
 | Queue | Type | Triggers | Labels Added | Status |
 |-------|------|----------|--------------|--------|
-| `triage_queue` | Input | No labels OR `ymir_retry_needed` OR `ymir_todo` | `ymir_triage_in_progress` (set by fetcher atomic flip for retry/todo, or by agent at triage start for fresh issues) | Active |
+| `triage_queue` | Input | No labels OR `ymir_retry_needed` | `ymir_triage_in_progress` (set by fetcher atomic flip for retry/todo, or by agent at triage start for fresh issues) | Active |
+| `triage_queue_todo` | Input (priority) | `ymir_todo` | Same as `triage_queue`. The triage agent BRPOPs `[triage_queue_todo, triage_queue]`, so ymir_todo tasks jump ahead of fresh/retry tasks. | Active |
 | `rebase_queue_c9s` | Input | Resolution=REBASE, RHEL 8/9 | `ymir_triaged_rebase` | Active (AUTO_CHAIN only) |
 | `rebase_queue_c10s` | Input | Resolution=REBASE, RHEL 10+ | `ymir_triaged_rebase` | Active (AUTO_CHAIN only) |
 | `backport_queue_c9s` | Input | Resolution=BACKPORT, RHEL 8/9 | `ymir_triaged_backport` | Active (AUTO_CHAIN only) |
@@ -170,4 +171,4 @@ JQL no longer restricts `ymir_todo` by assignee — the fetcher instead verifies
 
 ---
 
-**Last Updated:** 2026-06-04
+**Last Updated:** 2026-06-10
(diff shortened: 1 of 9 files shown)


ymir/common/constants.py
```diff
@@ -19,6 +19,10 @@ class RedisQueues(Enum):
     """Constants for Redis queue names used by Ymir agents"""
 
     TRIAGE_QUEUE = "triage_queue"
+    # Priority twin of TRIAGE_QUEUE for ymir_todo-triggered tasks. The triage
+    # agent BRPOPs from [TRIAGE_QUEUE_TODO, TRIAGE_QUEUE] so Redis serves the
+    # priority queue first whenever it has anything.
+    TRIAGE_QUEUE_TODO = "triage_queue_todo"
```
lbarcziova, line 25: should we do the same for other queues too and populate those also based on `user_triggered`?

jira_label_workflow_routing.md
```diff
@@ -99,7 +99,8 @@ flowchart TD
 
 | Queue | Type | Triggers | Labels Added | Status |
 |-------|------|----------|--------------|--------|
-| `triage_queue` | Input | No labels OR `ymir_retry_needed` OR `ymir_todo` | `ymir_triage_in_progress` (set by fetcher atomic flip for retry/todo, or by agent at triage start for fresh issues) | Active |
+| `triage_queue` | Input | No labels OR `ymir_retry_needed` | `ymir_triage_in_progress` (set by fetcher atomic flip for retry/todo, or by agent at triage start for fresh issues) | Active |
+| `triage_queue_todo` | Input (priority) | `ymir_todo` | Same as `triage_queue`. The triage agent BRPOPs `[triage_queue_todo, triage_queue]`, so ymir_todo tasks jump ahead of fresh/retry tasks. | Active |
```
nforro, line 103: Wouldn't `triage_queue_prio` be more self-explanatory?

majamassarini: If we plan to put other issue (for other labels there) then I agree. Otherwise this is for me easier to remember but I am not against it.

openshift/Makefile
```diff
@@ -9,4 +9,11 @@ run-jira-issue-fetcher-todo:
 	oc delete job jira-issue-fetcher-todo-manual || true
 	oc create job jira-issue-fetcher-todo-manual --from=cronjob/jira-issue-fetcher-todo
 
-.PHONY: deploy run-jira-issue-fetcher run-jira-issue-fetcher-todo
+show-triage-queue:
+	@echo "=== triage_queue_todo (ymir_todo, priority — popped first) ==="
+	@oc exec deployment/valkey -- valkey-cli LRANGE triage_queue_todo 0 -1 | jq -r .metadata.issue | tac
+	@echo
+	@echo "=== triage_queue (normal) ==="
+	@oc exec deployment/valkey -- valkey-cli LRANGE triage_queue 0 -1 | jq -r .metadata.issue | tac
```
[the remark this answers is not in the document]
majamassarini: for the makefile targets I have another PR open, which I should finalize, I will work there on this.

lbarcziova (approved):
LGTM, lets try this! Just please update the PR description as it's not up to date now.

PR #591, opened 2026-06-10
Implement retries in get_patch_from_url tool when 503 is encountered

TomasKorbar:
Github is starting to throttle our requests, thus we can add token in the future but still should be able to retry when this happens.

### DIFF
--- a/ymir/tools/http.py
+++ b/ymir/tools/http.py
@@ -1,0 +1,53 @@
+import asyncio
+import logging
+import random
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
+
+import aiohttp
+
+from ymir.tools.constants import (
+    AIOHTTP_MAX_RETRIES,
+    AIOHTTP_RETRY_BACKOFF_BASE,
+    AIOHTTP_RETRYABLE_STATUS_CODES,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@asynccontextmanager
+async def aiohttp_get_with_retries(
+    session: aiohttp.ClientSession,
+    url: str,
+    **kwargs,
+) -> AsyncIterator[aiohttp.ClientResponse]:
+    """Drop-in replacement for ``session.get()`` that retries on transient HTTP errors.
+
+    Retries up to ``AIOHTTP_MAX_RETRIES`` times on status codes listed in
+    ``AIOHTTP_RETRYABLE_STATUS_CODES`` using exponential back-off.  Raises
+    ``aiohttp.ClientResponseError`` when all retries are exhausted.
+    """
+    for attempt in range(AIOHTTP_MAX_RETRIES):
+        async with session.get(url, **kwargs) as response:
+            if response.status not in AIOHTTP_RETRYABLE_STATUS_CODES:
+                yield response
+                return
+            if attempt >= AIOHTTP_MAX_RETRIES - 1:
+                raise aiohttp.ClientResponseError(
+                    response.request_info,
+                    response.history,
+                    status=response.status,
+                    message=f"after {AIOHTTP_MAX_RETRIES} retries",
+                )
+            retry_status = response.status
+        # Response context is closed here — sleep without holding the connection
+        delay = AIOHTTP_RETRY_BACKOFF_BASE * (2**attempt) + random.uniform(0, 1)  # noqa: S311
+        logger.warning(
+            "Transient HTTP %d on %s, retrying in %.1fs (attempt %d/%d)",
+            retry_status,
+            url,
+            delay,
+            attempt + 1,
+            AIOHTTP_MAX_RETRIES,
+        )
+        await asyncio.sleep(delay)
--- a/ymir/tools/privileged/gitlab.py
+++ b/ymir/tools/privileged/gitlab.py
@@ -34,6 +34,7 @@ from ymir.common.models import (
 from ymir.common.validators import AbsolutePath
 from ymir.tools.base import CloneableTool as Tool
 from ymir.tools.constants import AIOHTTP_TIMEOUT, YMIR_USER_AGENT
+from ymir.tools.http import aiohttp_get_with_retries
 from ymir.tools.privileged.utils import clean_stale_repositories
 
 logger = logging.getLogger(__name__)
@@ -973,10 +974,15 @@ class GetPatchFromUrlTool(Tool[GetPatchFromUrlToolInput, ToolRunOptions, StringT
         try:
             async with (
                 aiohttp.ClientSession(timeout=AIOHTTP_TIMEOUT) as session,
-                session.get(request_url, headers=headers) as response,
+                aiohttp_get_with_retries(session, request_url, headers=headers) as response,
             ):
                 if response.status >= 400:
-                    raise ToolError(f"Failed to fetch patch from {patch_url}: HTTP {response.status}")
+                    # We return Error string instead of ToolError, because status >= 400
+                    # on for example badly formatted URL is not a tool error and
+                    # should not be flagged
+                    return StringToolOutput(
+                        result=f"Error: Failed to fetch patch from {patch_url}: HTTP {response.status}"
+                    )
                 text = await response.text()
         except aiohttp.ClientError as e:
             raise ToolError(f"Failed to fetch patch from {patch_url}: {e}") from e
@@ -1030,7 +1036,8 @@ class FetchGitlabMrNotesTool(Tool[FetchGitlabMrNotesInput, ToolRunOptions, Strin
         try:
             async with (
                 aiohttp.ClientSession(timeout=AIOHTTP_TIMEOUT) as session,
-                session.get(
+                aiohttp_get_with_retries(
+                    session,
                     url,
                     headers=headers,
                     params={
@@ -1066,6 +1073,10 @@ class FetchGitlabMrNotesTool(Tool[FetchGitlabMrNotesInput, ToolRunOptions, Strin
 
             return StringToolOutput(result=json.dumps(result, indent=2))
 
+        except aiohttp.ClientError as e:
+            # Here we handle ClientError as ToolError, because client error
+            # signals networking issues which should be flagged (DNS resolution failure, timeouts etc)
+            raise ToolError(f"Failed to fetch MR notes for !{input.mr_iid} in {input.project}: {e}") from e
         except Exception as e:
             logger.error("Error fetching GitLab MR notes: %s", e)
             return StringToolOutput(result=f"Error fetching GitLab MR notes: {e}")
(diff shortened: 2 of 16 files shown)


ymir/tools/privileged/gitlab.py
```diff
@@ -970,16 +976,31 @@ async def _run(
...
+                        logger.warning(
+                            f"Transient HTTP {response.status} fetching {patch_url}, "
+                            f"retrying in {delay}s (attempt {attempt + 1}/{AIOHTTP_MAX_RETRIES})"
+                        )
+                        await asyncio.sleep(delay)
+                        continue
+                    if response.status >= 400:
+                        return StringToolOutput(
+                            result=f"Error: Failed to fetch patch from {patch_url}: HTTP {response.status}"
+                        )
```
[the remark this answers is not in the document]
TomasKorbar: This is actually intentional, when agent tries to retrieve patch on wrong url and 404 is returned, that is a valid response and should not throw ToolError, as we should check them in test suite and treat ToolError as something broken in tools, but hostings returning 404 is not something broken, it is just failed attempt to retrieve the patch.

ymir/tools/privileged/gitlab.py
```diff
@@ -970,16 +976,31 @@ async def _run(
...
+                    text = await response.text()
+                    break
+            except aiohttp.ClientError as e:
+                raise ToolError(f"Failed to fetch patch from {patch_url}: {e}") from e
```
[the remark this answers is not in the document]
TomasKorbar: Network issues are valid cases for throwing ToolError, as they can be caused by our infra and should be flagged.

nforro: Wouldn't it make sense to make this a function and use it for every GET (or even other requests)?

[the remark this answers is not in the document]
TomasTomecek: agreed with Gemini to also include ClientErrors

TomasTomecek (dismissed):
Definitely a big improvement, I think the ClientError suggestion is good but I'd understand if we let that one bubble up

+1 Nikola to apply this to as many network requests as possible

Also Laura made a related change recently

ymir/tools/privileged/gitlab.py
TomasKorbar: Hmm but then i guess i should raise ToolError on running out of tries and encountering 503 again, as when we run out of tries somewhere where we do not expect it, we should flag that situation so we can add another token for example or investigate how to raise the limit for these requests.

ymir/tools/unprivileged/greenwave.py
```diff
@@ -136,13 +139,16 @@ async def _run(
...
                         if response.status == 200:
-                            # Limit reading to 10MB to prevent OOM on large files
                             content_bytes = await response.content.read(10 * 1024 * 1024)
                             content = content_bytes.decode("utf-8", errors="replace")
                             return StringToolOutput(result=content)
                         logger.warning("Got HTTP %d for %s", response.status, url)
+                except aiohttp.ClientError as e:
+                    raise ToolError(
+                        f"Failed to fetch Testing Farm results from {input.artifact_url}: {e}"
+                    ) from e
```
nforro: This breaks the original logic - try URLs one by one, until a request succeeds.

ymir/tools/unprivileged/read_readme.py
```diff
@@ -41,14 +42,17 @@ async def _run(
...
+                for prefix, suffix in README_PATTERNS:
+                    if input.repo_url.startswith(prefix):
+                        url = input.repo_url.removesuffix("/") + suffix
+                        async with aiohttp_get_with_retries(session, url) as response:
+                            if response.status == 200:
+                                return StringToolOutput(
+                                    result=await response.text(),
+                                )
+        except aiohttp.ClientError as e:
+            raise ToolError(f"Failed to fetch README from {input.repo_url}: {e}") from e
```
nforro: Same as above, the point is to advance to the next pattern on error.

ymir/tools/privileged/gitlab.py
```diff
@@ -973,10 +974,12 @@ async def _run(
...
             async with (
                 aiohttp.ClientSession(timeout=AIOHTTP_TIMEOUT) as session,
-                session.get(request_url, headers=headers) as response,
+                aiohttp_get_with_retries(session, request_url, headers=headers) as response,
             ):
                 if response.status >= 400:
-                    raise ToolError(f"Failed to fetch patch from {patch_url}: HTTP {response.status}")
+                    return StringToolOutput(
+                        result=f"Error: Failed to fetch patch from {patch_url}: HTTP {response.status}"
+                    )
```
nforro, line 985: Why was this changed? If it's on purpose, an explanatory comment would be nice.

ymir/tools/unprivileged/greenwave.py
```diff
@@ -71,6 +72,8 @@ async def _run(
                 return StringToolOutput(
                     result=f"Failed to fetch GreenWave gating status (HTTP {response.status}): {text}"
                 )
+        except aiohttp.ClientError as e:
+            raise ToolError(f"Failed to fetch GreenWave gating status: {e}") from e
```
nforro, line 78: Like above, but vice-versa, this changes `StringToolOutput` to `ToolError`. Maybe it's harmless, but I would rather preserve the original behaivor.

ymir/tools/privileged/gitlab.py
```diff
@@ -1066,6 +1070,8 @@ async def _run(
 
             return StringToolOutput(result=json.dumps(result, indent=2))
 
+        except aiohttp.ClientError as e:
+            raise ToolError(f"Failed to fetch MR notes for !{input.mr_iid} in {input.project}: {e}") from e
```
nforro, line 1079: Same as above.

ymir/tools/http.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Retries up to ``AIOHTTP_MAX_RETRIES`` times on status codes listed in
+    ``AIOHTTP_RETRYABLE_STATUS_CODES`` using exponential back-off.  Raises
+    ``aiohttp.ClientResponseError`` when all retries are exhausted.
+    """
+    for attempt in range(AIOHTTP_MAX_RETRIES):
+        async with session.get(url, **kwargs) as response:
+            if response.status in AIOHTTP_RETRYABLE_STATUS_CODES:
+                if attempt < AIOHTTP_MAX_RETRIES - 1:
+                    delay = AIOHTTP_RETRY_BACKOFF_BASE * (2**attempt)
```
nforro: It's a good practice to add random jitter - if multiple requests hit the same error (which is not so unlikely to happen) they would be retried at the exact same intervals, making the situation worse.

ymir/tools/http.py
```diff
@@ -0,0 +1,51 @@
...
+                    logger.warning(
+                        "Transient HTTP %d on %s, retrying in %ds (attempt %d/%d)",
+                        response.status,
+                        url,
+                        delay,
+                        attempt + 1,
+                        AIOHTTP_MAX_RETRIES,
+                    )
+                    await asyncio.sleep(delay)
```
nforro: It would be more efficient to exit the context before sleeping, otherwise the HTTP connection is unnecessarily held open for the duration of `delay`.

ymir/tools/unprivileged/greenwave.py
```diff
@@ -136,13 +139,16 @@ async def _run(
             for url in urls_to_try:
                 logger.info("Fetching Testing Farm results from %s", url)
                 try:
-                    async with session.get(url) as response:
+                    async with aiohttp_get_with_retries(session, url) as response:
                         if response.status == 200:
-                            # Limit reading to 10MB to prevent OOM on large files
```
nforro, line 141: Why did the comment disappear?

nforro (commented):
Would it make sense to add some simple unit test for `aiohttp_get_with_retries()`?

ymir/tools/privileged/gitlab.py
TomasKorbar: On purpose, as error code when Agent badly formats URL is not a ToolError. Will add comment

ymir/tools/privileged/gitlab.py
TomasKorbar: Affirmative, will add comment.

ymir/tools/unprivileged/greenwave.py
TomasKorbar: Added comment. I would keep this, because ClientError signifies that there is something wrong with out configuration or service or the GreenWave service and should be flagged.

ymir/tools/unprivileged/greenwave.py
TomasKorbar: oversight, will return the comment.

ymir/tools/unprivileged/greenwave.py
TomasKorbar: You are right, i missed this. Will revert to original behavior. Logging is more proper in this case.

ymir/tools/unprivileged/read_readme.py
TomasKorbar: You are right here again. Will revert to original behavior.

ymir/tools/http.py
TomasKorbar: good idea, adding.

ymir/tools/http.py
TomasKorbar: Good catch, will do.

TomasKorbar:
> Would it make sense to add some simple unit test for `aiohttp_get_with_retries()`?

Right, done.

ymir/tools/unprivileged/read_readme.py
```diff
@@ -45,7 +46,7 @@ async def _run(
             for prefix, suffix in README_PATTERNS:
                 if input.repo_url.startswith(prefix):
                     url = input.repo_url.removesuffix("/") + suffix
-                    async with session.get(url) as response:
+                    async with aiohttp_get_with_retries(session, url) as response:
```
nforro: There is a slight issue here, if all patterns fail with 503, the very last attempt will raise unhandled `ClientResponseError` exception.

nforro (dismissed):

ymir/tools/unprivileged/read_readme.py
TomasKorbar: Ah right, and i do not think only the last one, even the first one could return 503 continuously, exception would be raised and the rest would not even be tried. Fixed

nforro (approved):

cgwalters:
I am not a Python expert (anymore) but the last time I encounter this problem domain someone came by and added a dependency on https://tenacity.readthedocs.io/en/latest/

I have no opinions on its implementation, but the design does seem better than doing it ad-hoc in this way.

PR #595, opened 2026-06-10
skills: comply with agentskills.io standard, make client-generic

cgwalters:
It's always a bit sad to me when FOSS projects hardcode references to just proprietary tools like Claude Code and Cursor.

I understand why this probably happened: the "skills" thing was not standardized well, but now it is! There's https://agentskills.io/home

We needed some fixes though:

- Add missing 'name:' frontmatter field to all six SKILL.md files.
- The spec requires name to be lowercase alphanumeric + hyphens, so preliminary_testing → preliminary-testing and issue_verification → issue-verification (directories renamed accordingly).
- Remove the Ymir-specific 'arguments:' frontmatter block from all skills. Cross-checking every major client confirms none of them would consume it usefully: agentskills.io/skills-ref rejects it as an unknown field; Claude Code supports 'arguments:' but only as a flat list of names mapping to $name substitutions (our YAML object structure with description/required subfields and {{name}} syntax is not what it expects); Cursor has its own extensions (paths, disable-model-invocation) but no 'arguments:'; opencode silently ignores unknown fields. The argument docs are already in the '## Input Arguments' body section of each skill, so nothing is lost.
- Add a validate-skills job to check-in-container.yml that runs 'skills-ref validate' against every skill directory on each PR.

A few other tweaks included to make the install instructions more generic.

Assisted-by: OpenCode (Claude Sonnet 4.6)

RELEASE NOTES BEGIN

Packit now uses the standard agentskills.io format.

RELEASE NOTES END

### DIFF
--- a/.github/workflows/check-in-container.yml
+++ b/.github/workflows/check-in-container.yml
@@ -11,6 +11,18 @@ permissions:
   contents: read
 
 jobs:
+  validate-skills:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+    - name: Validate agent skills
+      run: |
+        pip install skills-ref==0.1.1
+        for skill_dir in agents_as_skills/*/; do
+          agentskills validate "$skill_dir"
+        done
+
   check-in-container:
     runs-on: ubuntu-latest
     strategy:
--- a/scripts/sync-agent-skills.py
+++ b/scripts/sync-agent-skills.py
@@ -2,9 +2,9 @@
 """Pre-push hook: ensure agents_as_skills/ stays in sync with ymir/agents/.
 
 Compares the net diff of the current branch against its remote tracking branch.
-Fails the push when a *_agent.py file or its prompt templates were changed but
-the corresponding SKILL.md under agents_as_skills/ was not.  Prints the exact
-claude command the developer should run to regenerate the skill.
+Fails the push when a *_agent.py file or its prompt templates were
+changed but their corresponding SKILL.md under agents_as_skills/ was not.
+Prints the exact command the developer should run to regenerate the skill.
 
 Set SKIP_SKILL_SYNC=1 to bypass this check.
 """
@@ -19,23 +19,29 @@ AGENTS_DIR = Path("ymir/agents")
 PROMPTS_DIR = Path("ymir/agents/prompts")
 SKILLS_DIR = Path("agents_as_skills")
 
-CLAUDE_CMD_TEMPLATE = (
+REGEN_CMD_TEMPLATE = (
     "claude --model claude-opus-4-6 --effort high"
+    "  # any AI agent works; adjust command for your tool"
     ' "Please take a look at the BeeAI workflows implemented in agents'
-    " directory. Please convert Workflow in {workflow_file} to Claude skill and"
-    f" save that skill to {SKILLS_DIR} directory.\n"
+    " directory. Please convert Workflow in {workflow_file} to an Agent Skill"
+    f" (https://agentskills.io/specification) and save it to {SKILLS_DIR}/.\n"
     "Restrictions:\n"
     " - Pay attention to tools used by the workflow and do not omit them\n"
     " - Do not restrict tools that the skill can use\n"
-    ' - Specify arguments the skill uses as an input"'
+    ' - The skill name must match the directory name (lowercase, hyphens only)"'
 )
 
 
 def skill_name_for(agent_path: Path) -> str | None:
-    """Derive the skill directory name from an agent filename."""
+    """Derive the skill directory name from an agent filename.
+
+    Agent filenames use underscores (e.g. preliminary_testing_agent.py) but
+    skill directories use hyphens per the agentskills.io spec, so underscores
+    are replaced with hyphens.
+    """
     if agent_path.suffix != ".py" or not agent_path.stem.endswith("_agent"):
         return None
-    return agent_path.stem.removesuffix("_agent")
+    return agent_path.stem.removesuffix("_agent").replace("_", "-")
 
 
 def skill_name_for_prompt(prompt_path: Path) -> str | None:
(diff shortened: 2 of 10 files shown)


.github/workflows/check-in-container.yml
```diff
@@ -11,6 +11,18 @@ permissions:
   contents: read
 
 jobs:
+  validate-skills:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v6
```
lbarcziova: there seem to be a discrepancy between the commit SHA and comment about version

.github/workflows/check-in-container.yml
```diff
@@ -26,7 +38,7 @@ jobs:
           - target: check-mcp-install-in-container
     steps:
     - name: Checkout code
-      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v6
```
lbarcziova: same as above

scripts/sync-agent-skills.py
```diff
@@ -18,23 +18,28 @@
 AGENTS_DIR = Path("ymir/agents")
 SKILLS_DIR = Path("agents_as_skills")
 
-CLAUDE_CMD_TEMPLATE = (
-    "claude --model claude-opus-4-6 --effort high"
+REGEN_CMD_TEMPLATE = (
+    "opencode"
```
lbarcziova: most of the team uses Claude Code (as far as I know), so could you make this mention both?

skills_installation.md
```diff
@@ -237,6 +197,40 @@ For project-scoped configuration, place the same `mcpServers` block in
 `.claude/settings.json` inside your project directory instead. If that
 does not work for you, use the `.mcp.json` at the root of your project.
 
+### opencode
```
lbarcziova, line 200: nice, thanks for this addition!

.github/workflows/check-in-container.yml
```diff
@@ -11,6 +11,18 @@ permissions:
...
 
 jobs:
+  validate-skills:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v6
+    - name: Validate agent skills
+      run: |
+        pip install skills-ref
```
lbarcziova: can we pin the version here too?

lbarcziova (commented):
these are good improvements, I didn't know there is a standard now! I see also some prompt changes (not the standardisation related ones), are those intentional?

cgwalters:
This PR leads to a *much* bigger thread, which we don't have to dive into but I want to just seed.

# Skills can contain scripts

This is part of the standard: https://agentskills.io/specification#scripts%2F

And I think some of these skills would be even more effective if the agent was directly executing an inline script. Take for example https://github.com/bootc-dev/agent-skills/tree/main/perform-forge-review - most of the skill is explaining how to use an inline (generated) tool. 

In the Ymir scripts today, I see e.g.

```
**Source 1: GreenWave / OSCI Gating Status**

If `build_nvr` is available (not null): 
- Fetch the GreenWave gating status page by retrieving the URL: `https://gating-status.osci.redhat.com/query?nvr=<build_nvr>`
- The HTML page shows which gating test jobs ran and whether they passed or failed.
- All required/gating tests must pass.
- The GreenWave Monitor URL is `https://gating-status.osci.redhat.com/query?nvr=<build_nvr>` — when linking to gating results in your comment, ONLY use this exact URL pattern. Do NOT invent or guess any other URLs for gating results.
```

I bet it'd be even easier for an agent if there was a `scripts/greenwave-status <build_nvr>` that it could just run.

And if one takes this out more...what I feel would probably make sense is to directly roll in some of these opinionated tools into `rhpkg` or so (not sure if that's been discussed).

IME agents do really well with tools that can track and report their own state. Like imagine a `rhpkg status [--json]` that e.g. reports outstanding unpushed commits, polls greenwave etc.

.github/workflows/check-in-container.yml
cgwalters: Hm thanks. Dang it, I got burned by this one because I asked the agent instead of checking myself.

BTW, checkout@v4 is really old now, and vulnerable to https://docs.zizmor.sh/audits/#artipacked - which gets into ensuring we have renovate/dependabot here

lbarcziova (commented):
LGTM. There are some conflicts though, could you please rebase?

TomasKorbar:
> This PR leads to a _much_ bigger thread, which we don't have to dive into but I want to just seed.
> # Skills can contain scripts
> 
> This is part of the standard: https://agentskills.io/specification#scripts%2F
> 
> And I think some of these skills would be even more effective if the agent was directly executing an inline script. Take for example https://github.com/bootc-dev/agent-skills/tree/main/perform-forge-review - most of the skill is explaining how to use an inline (generated) tool.
> 
> In the Ymir scripts today, I see e.g.
> 
> ```
> **Source 1: GreenWave / OSCI Gating Status**
> 
> If `build_nvr` is available (not null): 
> - Fetch the GreenWave gating status page by retrieving the URL: `https://gating-status.osci.redhat.com/query?nvr=<build_nvr>`
> - The HTML page shows which gating test jobs ran and whether they passed or failed.
> - All required/gating tests must pass.
> - The GreenWave Monitor URL is `https://gating-status.osci.redhat.com/query?nvr=<build_nvr>` — when linking to gating results in your comment, ONLY use this exact URL pattern. Do NOT invent or guess any other URLs for gating results.
> ```
> 
> I bet it'd be even easier for an agent if there was a `scripts/greenwave-status <build_nvr>` that it could just run.
> 
> And if one takes this out more...what I feel would probably make sense is to directly roll in some of these opinionated tools into `rhpkg` or so (not sure if that's been discussed).
> 
> IME agents do really well with tools that can track and report their own state. Like imagine a `rhpkg status [--json]` that e.g. reports outstanding unpushed commits, polls greenwave etc.

This is actually a glitch in the agent -> skill conversion process. It should have resulted in a instruction to use the `fetch_greenwave` tool.  The conversion prompt will probably need improvement.

TomasKorbar:
@cgwalters i took the liberty of fixing conflicts so we can merge this. Hope you don't mind.

TomasKorbar (approved):

cgwalters:
> @cgwalters i took the liberty of fixing conflicts so we can merge this. Hope you don't mind.

Awesome thanks!

PR #596, opened 2026-06-11
Surface real subprocess errors from lookaside tools

majamassarini:
## Problem

`download_sources`, `prep_sources`, and `upload_sources` run `rhpkg`/`centpkg` without capturing stdout/stderr and raise a bare `"Failed to ..."`. The real reason is thrown away, so it never reaches the agent or Phoenix.

Concretely: OpenEXR z-stream backports (RHEL-182266 / RHEL-182265) failed `download_sources` deterministically on June 5 **and** June 10. Every layer — the backport-agent traceback, the MCP `ToolError`, and the Phoenix span — showed only:

```
ToolError: Error executing tool download_sources: Failed to download sources
```

The actual `rhpkg sources` error (these issues route to RHEL branches `rhel-9.8.0`/`rhel-10.2` via `rhpkg` → internal lookaside, whereas CentOS-Stream packages use `centpkg` → public lookaside and succeed) is invisible, making it impossible to tell a lookaside 404 from a Kerberos failure from a network error without shelling into the pod.

## Change

Add a shared `_run_capturing()` helper that pipes `stdout`/`stderr` and, on non-zero exit, raises a `ToolError` that includes the command and a tail of its output:

```
Failed to download sources (`rhpkg --name=OpenEXR --namespace=rpms --release=rhel-9.8.0 sources` exited 1): <real error tail>
```

Applied to all three lookaside tools. Unit-test proc stubs updated from `wait()` to `communicate()`/`returncode` (argv unchanged).

## Note

This is the observability half. The underlying OpenEXR failure (backport agents can't reach the **RHEL internal lookaside** for `rhpkg`, which `needs_internal_fix` z-streams require) is a separate infra fix — but with this change its real cause will be self-evident in the logs next time.

### DIFF
--- a/ymir/tools/privileged/lookaside.py
+++ b/ymir/tools/privileged/lookaside.py
@@ -33,12 +33,23 @@ async def _try_init_kerberos():
 
 
 async def _run_capturing(argv: list[str], cwd: str | os.PathLike[str], fail_msg: str) -> str:
-    """Run argv in cwd, capturing stdout+stderr.
+    """Run argv in cwd, capturing stdout and stderr together.
 
-    On non-zero exit, raise ToolError including the command and a tail of its
-    output so the real reason (e.g. a lookaside 404, Kerberos failure, or
-    network error from rhpkg/centpkg) propagates back to the agent and Phoenix
-    instead of an opaque "Failed to ..." message.
+    Args:
+        argv: Command and its arguments to execute.
+        cwd: Working directory to run the command in.
+        fail_msg: Prefix for the error message raised on failure.
+
+    Returns:
+        The captured combined output, stripped of surrounding whitespace.
+
+    Raises:
+        ToolError: If argv is empty, the command cannot be executed (e.g. cwd
+            does not exist), or it exits non-zero. The error includes the
+            command and a tail of its output so the real reason (e.g. a
+            lookaside 404, Kerberos failure, or network error from
+            rhpkg/centpkg) propagates back to the agent and Phoenix instead of
+            an opaque "Failed to ..." message.
     """
     if not argv:
         raise ToolError(f"{fail_msg}: No command specified")
--- a/ymir/tools/privileged/lookaside.py
+++ b/ymir/tools/privileged/lookaside.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import os
 import shlex
+from pathlib import Path
 
 from beeai_framework.context import RunContext
 from beeai_framework.emitter import Emitter
@@ -61,7 +62,7 @@ async def _run_capturing(argv: list[str], cwd: str | os.PathLike[str], fail_msg:
             stderr=asyncio.subprocess.STDOUT,
         )
     except OSError as e:
-        if not os.path.isdir(cwd):
+        if not Path(cwd).is_dir():
             raise ToolError(f"{fail_msg}: Directory '{cwd}' does not exist") from e
         raise ToolError(f"{fail_msg}: Failed to execute {argv[0]}: {e}") from e
     out, _ = await proc.communicate()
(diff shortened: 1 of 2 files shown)


nforro:
Isn't the output in MCP gateway logs? This just exposes it (trimmed) as a tool error. Just wondering if we really need this...

ymir/tools/privileged/lookaside.py
```diff
@@ -31,6 +32,37 @@ async def _try_init_kerberos():
...
+    instead of an opaque "Failed to ..." message.
+    """
+    if not argv:
+        raise ToolError(f"{fail_msg}: No command specified")
+    try:
+        proc = await asyncio.create_subprocess_exec(
+            *argv,
+            cwd=cwd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.STDOUT,
```
TomasTomecek, line 62: it would be nice to store them separately and report them separately; but even this is a huge improvement, so let's merge this :)

ymir/tools/privileged/lookaside.py
```diff
@@ -31,6 +32,37 @@ async def _try_init_kerberos():
...
+        )
+    except OSError as e:
+        if not os.path.isdir(cwd):
+            raise ToolError(f"{fail_msg}: Directory '{cwd}' does not exist") from e
+        raise ToolError(f"{fail_msg}: Failed to execute {argv[0]}: {e}") from e
+    out, _ = await proc.communicate()
+    output = out.decode(errors="replace").strip()
```
TomasTomecek, line 69: I still remember when we failed to parse some systemd file in packit and it was because they had non-utf characters in their tests; but I'm assuming decode(errors=replace) should handle that

TomasTomecek (dismissed):

ymir/tools/privileged/lookaside.py
```diff
@@ -31,6 +32,37 @@ async def _try_init_kerberos():
         logger.warning("Kerberos initialization failed, continuing without it: %s", e)
 
 
+async def _run_capturing(argv: list[str], cwd: str | os.PathLike[str], fail_msg: str) -> str:
```
nforro, line 36: Why not just `cwd: Path`?

ymir/tools/privileged/lookaside.py
```diff
@@ -31,6 +32,37 @@ async def _try_init_kerberos():
...
+    """Run argv in cwd, capturing stdout+stderr.
+
+    On non-zero exit, raise ToolError including the command and a tail of its
+    output so the real reason (e.g. a lookaside 404, Kerberos failure, or
+    network error from rhpkg/centpkg) propagates back to the agent and Phoenix
+    instead of an opaque "Failed to ..." message.
+    """
```
nforro: It would be nice to follow the docstring guidelines (Google-style).

ymir/tools/privileged/lookaside.py
```diff
@@ -31,6 +32,37 @@ async def _try_init_kerberos():
...
+        raise ToolError(f"{fail_msg}: No command specified")
+    try:
+        proc = await asyncio.create_subprocess_exec(
+            *argv,
+            cwd=cwd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.STDOUT,
+        )
+    except OSError as e:
+        if not os.path.isdir(cwd):
```
nforro: Can't we use pathlib?

nforro (dismissed):
I'm not a big fan of this, but if it helps debugging....

ymir/tools/privileged/lookaside.py
majamassarini: Claude tells me, it is so.   ```Passing errors="replace" changes the contract: any invalid byte sequence is swapped for the U+FFFD replacement character (�) and decoding never raises.```

ymir/tools/privileged/lookaside.py
majamassarini: I would say mainly because unit tests are using strings, and the inner called methods accept also strings.

nforro (approved):

PR #602, opened 2026-06-17
Fix arbitrary build arch selection in build_package

majamassarini:
Prefer `x86_64` whenever the package allows it, and only fall back to another arch — deterministically, via `sorted()` — when `x86_64` is genuinely excluded:

```python
if not exclusive_arches or "x86_64" in exclusive_arches:
    build_arch = "x86_64"
else:
    build_arch = sorted(exclusive_arches)[0]
```

Assisted-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/ymir/tools/privileged/copr.py
+++ b/ymir/tools/privileged/copr.py
@@ -36,10 +36,11 @@ COPR_TIMEOUT_GRACE_PERIOD = 60  # seconds
 COPR_POLLING_INTERVAL = 30  # seconds
 COPR_ARCHES = {
     "aarch64",
-    "ppc64le",  # emulated
+    "ppc64le",
     "s390x",
     "x86_64",
 }
+COPR_ARCH_PREFERENCE = ("x86_64", "aarch64", "s390x", "ppc64le")
 
 
 logger = logging.getLogger(__name__)
@@ -94,9 +95,12 @@ class BuildPackageTool(Tool[BuildPackageToolInput, ToolRunOptions, BuildPackageT
             exclusive_arches = await self.get_exclusive_arches(srpm_path)
         except Exception as e:
             raise ToolError(f"Failed to read SRPM header: {e}") from e
-        # build for x86_64 unless the package is exclusive to other arch(es),
-        # in such case build for either of them
-        build_arch = exclusive_arches.pop() if exclusive_arches else "x86_64"
+        # build for the fastest supported arch (see COPR_ARCH_PREFERENCE);
+        # default to x86_64 when the package has no ExclusiveArch
+        build_arch = next(
+            (arch for arch in COPR_ARCH_PREFERENCE if arch in exclusive_arches),
+            "x86_64",
+        )
         rhel_config = await load_rhel_config()
         upcoming_z_streams = rhel_config.get("upcoming_z_streams", {})
         try:


ymir/tools/privileged/copr.py
```diff
@@ -94,9 +94,14 @@ async def _run(
             exclusive_arches = await self.get_exclusive_arches(srpm_path)
         except Exception as e:
             raise ToolError(f"Failed to read SRPM header: {e}") from e
-        # build for x86_64 unless the package is exclusive to other arch(es),
-        # in such case build for either of them
-        build_arch = exclusive_arches.pop() if exclusive_arches else "x86_64"
+        # build for x86_64 whenever it is allowed; only fall back to another arch
+        # when the package genuinely excludes x86_64. pick deterministically
+        # (exclusive_arches is a set, so .pop() would choose an arbitrary arch and
+        # could land on one that has no corresponding Copr chroot, e.g. custom-1-s390x)
```
nforro: The comment is unnecessarily detailed, nobody cares what the implemention looked like before fixing the issue. I would remove at least the last two lines.

ymir/tools/privileged/copr.py
```diff
@@ -94,9 +94,14 @@ async def _run(
...
+        if not exclusive_arches or "x86_64" in exclusive_arches:
+            build_arch = "x86_64"
+        else:
+            build_arch = sorted(exclusive_arches)[0]
```
nforro: Hm, I suppose it would be better not to use alphabetical order but rather a predefined list of arches ordered by build performance. There is no reason to build on a slower arch if the package supports a faster one.

nforro: I think the order would be:
- x86_64 (already special-cased, no need to include it)
- ppc64le
- aarch64
- s390x

However, it seems that ppc64le is emulated in Copr, in that case it's probably slower than aarch64. We could ask someone from Copr or do a few test builds on all arches and analyze the duration.

ymir/tools/privileged/copr.py
TomasTomecek: I personally don't mind comments that are more detailed but you are right this information should come naturally from the commit itself

ymir/tools/privileged/copr.py
majamassarini: I asked a review in #copr channel.

ymir/tools/privileged/copr.py
```diff
@@ -40,6 +40,10 @@
     "s390x",
     "x86_64",
 }
+# Preferred build arch when a package supports several, fastest first. x86_64 is
+# preferred whenever allowed; ppc64le is emulated in Copr so it comes last. The
```
nikromen: this is not true actually... we have native ppc64le builds. We actually have quite a lot of them (around ~80 builders at once available)

actually s390x resources are scarse and we have only a few of them... also there is a possible risk that we may be forced to switch to emulation (hopefully not)

nforro: @nikromen note that this is about copr.devel.redhat.com (in case it matters).

ymir/tools/privileged/copr.py
nikromen: from where you got the information of emulated ppc64le chroots? We should delete it then since we have them native :)

also yes please use s390x as last resort arch... we don't have a lot of them :/

nforro:
Reading the PR description, these changes don't fix the original issue. The changes are good, we want determinism for tests if nothing else, but the problem as described is about:

> ```
> chroots: 'custom-1-s390x' is not a valid choice for this field.
> ```

Why does that happen, that's the question.

> When the picked arch has no corresponding custom Copr chroot enabled on the instance

What does this even mean? We are "creating" the chroot, there is nothing to be enabled.

ymir/tools/privileged/copr.py
majamassarini: Thanks a lot @nikromen for the feedback. So I would go with this order ->  ("x86_64", "ppc64le", "aarch64", "s390x")

nikromen: aha then the situation changes... we have actually them native as well but in different (and painfully buggy) cloud... it takes up to 15 minutes to allocate builder from the cloud :D we have 2 pre-allocated in case of need but then if you submit multiple builds (or someone else) you have to wait for some time unfortunately.... 

if we are speaking solely about rh copr then we have native s390x and you can use them as you wish - we have enough of them there. However even with native ppc64le builders yeah it could take some time to get one... unfortunately

as for fedora copr... you can use ppc64le as you want we have enough of them but we don't have enough s390x :D but on both instances we have native arches

sooo if we speak about rh copr from where is the information about ppc64le is not native please? I'd like to delete that

ymir/tools/privileged/copr.py
nforro: > from where you got the information of emulated ppc64le chroots?

From the web UI:
> **epel-10-ppc64le**  An [emulated](https://gitlab.cee.redhat.com/internal-copr/docs/-/wikis/home#build-chroots) chroot.

nikromen:
> Reading the PR description, these changes don't fix the original issue. The changes are good, we want determinism for tests if nothing else, but the problem as described is about:

I am not diving into the codebase here so I am just guessing - but isn't your problem that you have pre-created custom chroot only for x86 so whenever you get different chroot than that it fails? So yeah defaulting to x86 should fix this but then choosing again from different arches would lead to the same exact issue as described, no?

So perhaps to this the fix should be to create also custom <arch> chroot for the repos and call it a day?

nforro:
> So perhaps to this the fix should be to create also custom chroot for the repos and call it a day?

Isn't that what we are doing?

https://github.com/packit/ai-workflows/blob/7027c3bc94ce271227d4048710e5077ac054ef32/ymir/tools/privileged/copr.py#L117-L123

nikromen:
ah upon checking the copr instance we do not have enabled custom chroots for other instances than x86! so that is probably why are you seeing this error on other arches - you probably try to enable custom-<whatever arch other than x86) but the API will shout at you that this chroot simply does not exist... soooo I guess enabling it directly in copr instance may solve it? I do not see any technical reason why we can't have custom chroots for other arches as well.... so if you want I can enable them for you which may solve this (but it would require some confirmation builds, I am not 100% sure)

nforro:
> so if you want I can enable them for you

That would be great. I don't suppose there are many packages that don't build on `x86_64` but we still need build verification for those.

As for the ordering, given ppc64le is native, it should be like this, right (fastest to slowest)?
1. x86_64
1. ppc64le
1. aarch64
1. s390x

ymir/tools/privileged/copr.py
nforro: > sooo if we speak about rh copr from where is the information about ppc64le is not native please? I'd like to delete that

I responded above - it is on the web UI (though only for epel-10):

> **epel-10-ppc64le**  An [emulated](https://gitlab.cee.redhat.com/internal-copr/docs/-/wikis/home#build-chroots) chroot.

nikromen:
OK I created them so these are now available:
image

ymir/tools/privileged/copr.py
nikromen: ah I see... that is weird because I do not see this comment configured... either I am blind or it's a bug (probably the first one :D) so I'll create ticket for it and in recent future it should be gone

nikromen:
> As for the ordering, given ppc64le is native, it should be like this, right (fastest to slowest)?

please eventhough ppc64le is native, keep it as last one (if it is rh copr we are speaking about) due to the 15 minutes of waiting for cloud provider of those power arches before it gives us VM :/

nforro:
> OK I created them so these are now available

Thanks a lot!

@majamassarini The above Copr change should fix the original issue. This PR is still worth merging, but could you please adjust the commit message and PR description to match what it actually does?

nforro (approved):
Thanks!

PR #603, opened 2026-06-17
Post backport failure feedback to Jira on the final retry

majamassarini:
## Problem

When a backport fails in an **unguarded** step — `fork_and_prepare_dist_git` (which runs `get_patch_from_url`) or `run_backport_agent` (which does the patch-apply) — the exception propagates out of the workflow to `main()`'s top-level `except`. That handler sets `ymir_backport_errored` and pushes the traceback to `error_list`, but **never posts a Jira comment**. The in-workflow `comment_in_jira` step is only reached on *graceful* failures, so crashes leave the maintainer with a silently-flipped label and no explanation.

This is especially bad for `ymir_todo`-triggered runs, where a human explicitly asked for processing and is waiting on a result.

### Seen in production

**RHEL-184829** (perl-HTTP-Daemon, CVE-2026-8450, rhel-9.8.z) was triggered via `ymir_todo`. Triage posted ack + "Resolution: backport" comments, then all three backport attempts crashed in unguarded steps (a transient `get_patch_from_url` error, a Chat Model error, and a real patch-apply conflict). The label flipped to `ymir_backport_errored` with **no comment** explaining why.

## Fix

Post a single failure comment from `retry()`'s **final-attempt** branch (`task.attempts >= max_retries`) — never for intermediate retries:

- Gated on `user_triggered` so only `ymir_todo` runs notify, and so we don't open an MCP gateway connection otherwise.
- Only the **crash** path passes `comment_text` (a concise `str(e)`; the full traceback still goes to `error_list`). The **graceful** path is left untouched — its in-workflow `comment_in_jira` step already posted feedback — so there is no double-comment.
- The shared `comment_in_jira` method is not modified.

## Net effect

- Crash failures now get exactly one feedback comment, on the last retry, for `ymir_todo` runs.
- Graceful failures keep their existing behavior.
- No duplicate comments; no comments added for intermediate attempts.

Assisted-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/ymir/agents/backport_agent.py
+++ b/ymir/agents/backport_agent.py
@@ -840,7 +840,9 @@ async def main() -> None:
                 + (" (user-triggered via ymir_todo)" if user_triggered else "")
             )
 
-            async def retry(task, error, backport_data=backport_data, user_triggered=user_triggered):
+            async def retry(
+                task, error, comment_text=None, backport_data=backport_data, user_triggered=user_triggered
+            ):
                 task.attempts += 1
                 if task.attempts < max_retries:
                     logger.warning(
@@ -850,6 +852,7 @@ async def main() -> None:
                     retry_queue = backport_queue_todo if task.user_triggered else backport_queue
                     await fix_await(redis.lpush(retry_queue, task.model_dump_json()))
                 else:
+                    # Final attempt exhausted — mark errored and stop retrying.
                     logger.error(
                         f"Task failed after {max_retries} attempts, "
                         f"moving to error list: {backport_data.jira_issue}"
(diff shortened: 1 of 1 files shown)


opohorel (approved):
LGTM, thanks!
also I want to note that it might be beneficial to add the same check to other agents too

lbarcziova (approved):
thanks!

majamassarini:
> LGTM, thanks! also I want to note that it might be beneficial to add the same check to other agents too

yes, I thought about it but I wasn't sure if you liked the approach. Tomorrow I will open another PR for the other workflows!

PR #609, opened 2026-06-18
Add triage target to openshift/Makefile

majamassarini:
Allows manually queuing a RHEL issue for triage by pushing it to the triage_queue. Usage: make process ISSUE=RHEL-XXXX

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/openshift/Makefile
+++ b/openshift/Makefile
@@ -113,6 +113,16 @@ logs-redis-commander:
 logs-otel-collector:
 	oc logs -f deployment/otel-collector
 
+process:
+	@if [ -z "$(ISSUE)" ]; then \
+		echo "ERROR: ISSUE parameter is required. Usage: make process ISSUE=RHEL-XXXX"; \
+		exit 1; \
+	fi
+	@echo "Pushing $(ISSUE) to triage_queue..."
+	@oc exec deployment/valkey -- valkey-cli LPUSH triage_queue \
+		'{"metadata":{"issue":"$(ISSUE)"},"attempts":0,"user_triggered":false}'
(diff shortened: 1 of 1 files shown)


openshift/Makefile
```diff
@@ -113,6 +113,16 @@ logs-redis-commander:
...
 
+triage:
+	@if [ -z "$(ISSUE)" ]; then \
+		echo "ERROR: ISSUE parameter is required. Usage: make triage ISSUE=RHEL-XXXX"; \
+		exit 1; \
+	fi
+	@echo "Pushing $(ISSUE) to triage_queue..."
+	@oc exec deployment/valkey -- valkey-cli LPUSH triage_queue \
+		'{"metadata":{"issue":"$(ISSUE)"},"attempts":0,"user_triggered":false}'
+	@echo "✓ Task queued successfully. View queue with: make show-triage-queue"
```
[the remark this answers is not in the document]
majamassarini: no, because this is meant to be used by the ymir team. I don't want to make our users wait. It is on the normal queue on purpose.

openshift/Makefile
```diff
@@ -113,6 +113,16 @@ logs-redis-commander:
 logs-otel-collector:
 	oc logs -f deployment/otel-collector
 
+triage:
```
lbarcziova: should we rather call it `process` or similar? Since it runs the e2e, not just triage

nforro: Since there is `show-triage-queue`, something like `push-to-triage-queue` would also make sense, but as Laura says, doing so triggers the entire workflow.

nforro (approved):

PR #611, opened 2026-06-19
Post rebase and rebuild failure feedback to Jira on the final retry

majamassarini:
Apply the same error feedback pattern from PR #603 to the rebase and rebuild agents. When a task crashes (exception path) and exhausts all retries, post the error message to Jira if the task was user-triggered.

This ensures users get feedback for unexpected crashes that never reach the workflow's comment_in_jira step. Graceful failures (workflow-level) already post comments via the workflow, so we skip comment_text there to avoid double-commenting.

Related: #603

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>


nforro (approved):
LGTM

PR #613, opened 2026-06-22
Filter get_jira_details output and request only needed fields

TomasKorbar:
Use the Jira API `fields` query parameter to request only the ~29 fields agents actually consume (20 standard + 9 custom) instead of receiving all ~174. Postprocess the response to strip avatarUrls unconditionally, drop null-valued fields, and trim remote-link objects to {id, relationship, object.{url, title, summary}}.

Teach the aiohttp mock to honour the `fields` parameter as well so that token consumption in MOCK_JIRA runs is representative. Also fix a pre-existing mock bug where remote_link=True was passed to flexmock instead of to the _read_jira_mock partial.

### DIFF
--- a/ymir/tools/privileged/aiohttp_client_session_mock.py
+++ b/ymir/tools/privileged/aiohttp_client_session_mock.py
@@ -28,12 +28,17 @@ async def _get_unverified_user():
     return {"groups": {"items": []}}
 
 
-async def _read_jira_mock(issue_key: str, remote_link=False) -> dict:
+async def _read_jira_mock(
+    issue_key: str, remote_link: bool = False, requested_fields: set[str] | None = None
+) -> dict | list:
     try:
         async with aiofiles.open(f"{os.environ['JIRA_MOCK_FILES']}/{issue_key}") as jira_file:
+            data = json.loads(await jira_file.read())
             if remote_link:
-                return json.loads(await jira_file.read())["remote_links"]
-            return json.loads(await jira_file.read())
+                return data.get("remote_links", [])
+            if requested_fields is not None and "fields" in data:
+                data["fields"] = {k: v for k, v in data["fields"].items() if k in requested_fields}
+            return data
     except (OSError, FileNotFoundError, json.JSONDecodeError) as e:
         raise ToolError(f"Error while reading mock up Jira issue {e}") from e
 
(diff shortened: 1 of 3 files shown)


ymir/tools/privileged/jira.py
```diff
@@ -62,6 +96,52 @@ class PreliminaryTesting(Enum):
...
+def _filter_remote_links(remote_links: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Trim remote link objects to only the data agents use."""
+    cleaned = []
+    for link in remote_links:
+        obj = link.get("object", {})
+        entry: dict[str, Any] = {"id": link.get("id")}
+        if link.get("relationship"):
+            entry["relationship"] = link["relationship"]
+        if obj:
+            entry["object"] = {k: v for k, v in obj.items() if k in ("url", "title", "summary")}
```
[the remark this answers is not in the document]
TomasKorbar: This should not happen according to api specification https://developer.atlassian.com/server/jira/platform/using-fields-in-remote-issue-links/

ymir/tools/privileged/aiohttp_client_session_mock.py
```diff
@@ -77,20 +82,33 @@ async def __aenter__(self):
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         pass
 
+    @staticmethod
+    def _parse_requested_fields(kwargs: dict) -> set[str] | None:
+        """Extract the ``fields`` query parameter into a set, or None if absent."""
+        raw = kwargs.get("params", {}).get("fields") if kwargs.get("params") else None
+        if raw is None:
+            return None
+        return set(raw.split(","))
```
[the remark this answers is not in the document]
TomasKorbar: In such case exception should be thrown and we should fix the caller. Dictionary seems to be the most appropriate type.

ymir/tools/privileged/jira.py
```diff
@@ -62,6 +96,52 @@ class PreliminaryTesting(Enum):
...
+        "fields": issue_data.get("fields", {}),
+    }
+
+    remote_links = issue_data.get("remote_links")
+    if isinstance(remote_links, list):
+        result["remote_links"] = _filter_remote_links(remote_links)
+    else:
+        result["remote_links"] = remote_links
+
+    return _remove_avatar_urls(result)
```
nforro: I noticed Claude really dislikes encapsulating utility functions as methods of the related tool class. Wouldn't it make much more sense to make all these methods of `GetJiraDetailsTool`?

ymir/tools/privileged/jira.py
```diff
@@ -37,6 +37,40 @@
...
+    "creator",
+    "created",
+    "updated",
+    "security",
+    "customfield_10578",  # Fixed in Build
+    "customfield_10638",  # Test Coverage
+    "customfield_10879",  # Preliminary Testing
+    "customfield_10371",  # AssignedTeam
+    "customfield_10418",  # Errata Link
+    "customfield_10626",  # Errata Link (fallback)
```
nforro: It would be nice to define constants for these custom fields as well, similarly to the ones below.

nforro (dismissed):
LGTM, thanks!

ymir/tools/privileged/jira.py
TomasKorbar: done.

nforro (approved):

PR #617, opened 2026-06-23
Add physical presence rule to CVE applicability agent

opohorel:
- Adds a "physical presence rule" to the CVE applicability prompt so the agent correctly classifies packages as `NOT_AFFECTED` when a dependency's source files are absent from the build tree, even if a manifest (e.g. `package.json`) declares it. This fixes misclassification for cases like `dotnet9.0/ws` where `BuildNodeJS=false` means npm dependencies are never installed.
- Adds an e2e test case (RHEL-186838: dotnet9.0 + CVE-2026-45736/ws) validating this behavior.


TomasKorbar (approved):
LGTM

PR #618, opened 2026-06-23
Improve consistency of Jira refs in spec patch naming and changelog

antbob:
As the title suggests these changes just improve Jira references in the spec file patch names and changelog entries when backporting. This is specifically to accommodate the older z-stream requirements. We have been running with these changes for a couple of weeks now and they seem to hold and do whats generally expected/accepted for older z-stream backports.

### DIFF
--- a/ymir/agents/prompts/backport/instructions_zstream.j2
+++ b/ymir/agents/prompts/backport/instructions_zstream.j2
@@ -107,8 +107,12 @@ a backport that won't actually fix the shipped RPM.
           - Identify which files are new patch files and what spec changes
             were made.
           - Extract new patch file(s) added by the commit into the local
-            dist-git clone working tree:
-            `git -C <DISTGIT_SOURCE> show <commit_hash>:<patch_filename> > <LOCAL_CLONE>/<patch_filename>`
+            dist-git clone working tree. If the patch filename contains a
+            Jira issue key matching the pattern of <JIRA_ISSUE> that differs
+            from <JIRA_ISSUE>, replace only that key with <JIRA_ISSUE> and
+            keep the rest of the filename intact. Do NOT rename patches
+            otherwise. Ensure all resulting filenames are unique:
+            `git -C <DISTGIT_SOURCE> show <commit_hash>:<patch_filename> > <LOCAL_CLONE>/<target_filename>`
           - Examine the spec file changes for reference:
             `git -C <DISTGIT_SOURCE> diff <commit_hash>^..<commit_hash> -- *.spec`
           - Note what Patch tags and `%prep` entries were added. Use this as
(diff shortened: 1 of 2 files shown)


antbob:
@lbarcziova @TomasTomecek i dunno if you guys need the patch naming changes in the main instructions as well (if that would be beneficial to you or not) but if you do then let me know and i'll add them there as well.

ymir/agents/prompts/backport/instructions_zstream.j2
```diff
@@ -107,8 +107,12 @@ a backport that won't actually fix the shipped RPM.
           - Identify which files are new patch files and what spec changes
             were made.
           - Extract new patch file(s) added by the commit into the local
-            dist-git clone working tree:
-            `git -C <DISTGIT_SOURCE> show <commit_hash>:<patch_filename> > <LOCAL_CLONE>/<patch_filename>`
+            dist-git clone working tree. If the patch filename contains a
+            Jira issue key matching the pattern of <JIRA_ISSUE> that differs
+            from <JIRA_ISSUE>, replace only that key with <JIRA_ISSUE> and
+            keep the rest of the filename intact. Do NOT rename patches
+            otherwise. Ensure all resulting filenames are unique:
```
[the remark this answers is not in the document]
antbob: no, we dont want regex here, just general instructions, the model seem to handle it just fine in practice.

TomasTomecek (approved):
Looks good to me.

WDYT @TomasKorbar @opohorel

ymir/agents/prompts/log/instructions.j2
```diff
@@ -15,18 +15,21 @@ of changes performed, do the following:
 
    If a source changelog message is provided in the prompt, use those lines as the
    exact changelog message content. Keep the descriptive lines exactly as-is — do not
-   rephrase, summarize, or add to them. In any Resolves/Related lines, replace all
-   original Jira issue references with <JIRA_ISSUES>, ensuring no duplicates remain.
+   rephrase, summarize, or add to them. If the source changelog message contains
+   Jira references to the issue(s) it resolves, replace only those references with
+   <JIRA_ISSUES> — leave other Jira references that appear in descriptive text unchanged.
```
opohorel, line 20: I wonder what is going to happen when we start consolidating the issues, because this seems as potential breakage

opohorel:
have you tested how the change to the log instructions affect all the workflows? because the log is shared with our base backporting workflow and rebase workflow

ymir/agents/prompts/log/instructions.j2
antbob: i cant answer that bc i have no idea how it would actually work and i dont wanna speculate.

antbob:
@opohorel no, this has been tested on older z-stream backports only which is what is available to me. i've assumed that you would want this change to have consistent changelog entries in your deployment as well but if you dont i can just separate this into a z-stream specific prompt file, similar with how we do this with other prompts already.

opohorel (approved):
ok, so lets merge it. hopefully nothing breaks for any of our workflows 😅

antbob:
ok, thanks guys! fingers crossed :) if it does misbehave in your workflow just let me know and i'll separate it into z-stream specific prompt then.

PR #619, opened 2026-06-23
OpenAI API specs impose max 64 characters limit on tool names

antbob:
This is something i ran into testing agents as skills with Claude Code using open weight models served by vLLM via OpenAI compatible API adapter. The problem is that OpenAI API spec limits tool names to 64 characters (which is a departure from the MCP spec where its up to 128 but its really up to implementation) and Claude Code prefixes everything so tool names blow up past 64 characters eg

mcp__ymir-privileged__get_failed_pipeline_jobs_from_merge_request

Since there is nothing that can be done about Claude Code behavior here its therefore a good workaround to shorten tool names where we can and where it doesnt compromise the meaning of those names so this PR does just that.

### DIFF
--- a/ymir/agents/preliminary_testing_agent.py
+++ b/ymir/agents/preliminary_testing_agent.py
@@ -91,7 +91,7 @@ def create_preliminary_testing_agent(gateway_tools: list) -> ReasoningAgent:
             in [
                 "fetch_gitlab_mr_notes",
                 "get_jira_details",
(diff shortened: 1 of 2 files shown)


antbob:
hearing that agents as skill will be possibly going away / obsolete soon, this might be irrelevant then so feel free to close.

TomasTomecek (approved):
LGTM, definitely makes sense

PR #625, opened 2026-06-24
Test consolidation and multiple bugfixes

TomasKorbar:
**Triage agent prompt improvements:**
- Tightened the definition of "follow-up commits" — the agent now only includes follow-ups that are strictly necessary for the fix to be correct (e.g. regression fixes), and explicitly excludes nice-to-have hardening, test-only, or refactoring commits
- Added guidance to use commits from the target branch (e.g. `main`) rather than PR source/fork branch hashes, which differ after rebase/squash merges
- Added instructions about identifying rebased commits
- Moved the output schema definition and examples directly into the prompt template instead of duplicating it in `triage_agent.py` as a verbose `expected_output` string — the agent now gets the schema once, in context, reducing token waste and eliminating the prior workaround for models struggling with nested JSON

**Backport agent prompt improvements:**
- Improved CVE patch identification: the agent must now inspect comments above `Patch` tags (not just filenames) to detect CVE-related patches and determine the naming convention
- Added critical guidance for conflict resolution: when resolving cherry-pick conflicts, the agent must distinguish the actual fix from unrelated upstream evolution on the "theirs" side, preserving HEAD-side code for anything not part of the original commit

**Timeout handling:**
- All `aiohttp.ClientError` exception handlers across Jira, GitLab, GreenWave, ResultsDB, upstream tools, logfile reader, and README reader now also catch `TimeoutError`, so network timeouts surface as clear `ToolError` messages in agent spans rather than unhandled exceptions

**DRY_RUN support for fork creation:**
- `ForkRepositoryTool` now respects `DRY_RUN=true` by returning the original repo URL instead of creating an actual fork

**Test infrastructure:**
- Added `slow` marker support for backport e2e tests — tests tagged `slow: true` in their fixture config are skipped by default and only run when `RUN_SLOW_TESTS=true` is set
- Added three new triage e2e test cases: memcached (RHEL-179083), qt6-qtdeclarative (RHEL-173494), nginx (RHEL-178684)
- Removed an incorrect expected follow-up patch from the libtiff (RHEL-112546) test case, aligning with the stricter follow-up commit definition


ymir/agents/triage_agent.py
```diff
@@ -401,96 +399,11 @@ async def run_triage_analysis(state):
...
-                    }}
-                    ```
-
-                    ```json
-                    {output_schema_json}
-                    ```
-                    """
+                expected_output=(
+                    "A JSON object with top-level keys 'resolution' and 'data'. "
+                    "See the task prompt for the exact schema and examples."
```
nforro: > Previous approach to setting output JSON schema of triage agent
resulted in it being provided twice and thus effectivelly doubled
the token count of initial prompt.

What do you mean, twice? Can you explain and/or provide some evidence? AFAICT this change only moves the final answer instructions from the start of the prompt to its end.

nforro (commented):
Nice!

nforro:
Here are two issues that Claude found:

**Bug: Stale step references in triage SKILL.md (Step 1)**

The PR renumbers the triage workflow steps (old Step 6b → Step 9, old Step 7 → Step 10, etc.), but four cross-references in Step 1 (CVE Eligibility Check) were not updated:

- Line 88: Skip to **Step 8: Comment in JIRA** → should be Step 10
- Line 101: Skip to **Step 8: Comment in JIRA** → should be Step 10
- Line 115: Skip to **Step 8: Comment in JIRA** → should be Step 10
- Line 156: skip to **Step 8** → should be Step 10

These send the agent to "Step 8: Verify Rebuild Buildroot" instead of "Step 10: Comment in JIRA" for CVE eligibility early-exit paths — a meaningful behavioral change if followed literally.

**Nit: Inconsistent TimeoutError style in copr.py**

The PR adds TimeoutError to all except aiohttp.ClientError handlers across 7 files. ymir/tools/privileged/copr.py already handles TimeoutError but via a separate except clause rather than the tuple pattern. Not a bug (functionally correct), but stylistically inconsistent. Might be worth unifying for grep-ability.

ymir/agents/triage_agent.py
nforro: Ok, I did the research myself and you are right, when `str`, `expected_output` is injected into both system and user prompt. No idea why, there is no context in BeeAI code nor git history. I'd argue that proper fix would be to remove it altogether from the template and related models (maybe we could do it in `RequirementAgent` as well and contribute it upstream):
https://github.com/packit/ai-workflows/blob/b4ed49217ed9536576dbedcb341ea35ae43013fb/ymir/agents/reasoning_agent/prompts.py#L122-L126
I think having it in system prompt only could save even more tokens thanks to caching.

TomasKorbar: That would be nice. Although the expected output has one additional problem. If you pass the output model class through it, the formatting somehow confuses agent and it attempts to encapsulate the output object properties in the data member.
```
{
    "resolution": "not-affected",
    "data": {
        "justification_category": "Vulnerable Code not Present",
        "explanation": "The downstream package ships version 1.2.3 which is below the affected range (2.0+). The vulnerable function does not exist in this version.",
        "jira_issue": "RHEL-12345"
    }
}
```
Became in my tests
```
{
"data": {
       "resolution": "not-affected",
        "justification_category": "Vulnerable Code not Present",
        "explanation": "The downstream package ships version 1.2.3 which is below the affected range (2.0+). The vulnerable function does not exist in this version.",
        "jira_issue": "RHEL-12345"
    }
}
```
I did not debug that further but only the change you are now seeing in this PR took care of all the issues and worked reliably across 3 consecutive triage e2e suite runs.

nforro: Well, that's the issue we've "solved" in the past by passing a text with examples instead of a Pydantic model as `expected_output`. I'm not proposing to change that, on the contrary, I would keep it as it is/was and just remove `expected_output` from the user prompt - it is injected there only if it is a string.

nforro: BTW, it puzzles me that Claude models still can't figure this out while Gemini 2.5 had no issue with it.

nforro (approved):

ymir/agents/triage_agent.py
TomasKorbar: Yeah, this "model quirk" is trully strange :D

PR #630, opened 2026-06-25
Add async task multiplexing for concurrent task processing

nforro:
With `MAX_CONCURRENT_TASKS=1` (the default), the workflows should behave exactly the same as before. So I think we can deploy this and if there are no regressions, we can try increasing the number and monitor resource usage.

This PR also modifies logging:
- `GlobalTrajectoryMiddleware` output now goes through loggers, so it gets prefixed just like regular log messages
- log messages are associated with a Jira key being processed if possible
- with `LOG_BUFFER_SIZE` higher than zero, log messages associated with the same Jira issue are buffered and output in batches, for better readability when multiple tasks are processed in parallel


TomasKorbar (changes requested):
Please bump the common subpackage version. Otherwise LGTM!

nforro:
> Please bump the common subpackage version.

I've added a hook to bump automatically before pushing, unless the version has already changed.

TomasKorbar (approved):
Thanks for the hook. LGTM!

TomasTomecek (approved):
LGTM, opus 4.6 did a thorough review and none of the comments were worth addressing from my PoV; well done!!

PR #631, opened 2026-06-26
Handle single Jira, single patch, multiple CVE commits

antbob:
Sometimes we get single Jira, single patch backports that bundle multiple CVEs together as opposed to the regular/standard with one CVE per Jira issue approach. Note that i'm piggybacking on the "cve_id" field here for this which perhaps is not the cleanest solution but it avoids larger refactoring changes. It should be ok as this is not exactly an API and the model is given a clear guidance, examples, and logic on how to do it. Also note that i've tested this with backport only (real backport, real Jira issue) but not other resolution types as we're currently not running rebase and rebuild agents and don't have any such Jiras to test those types on.

### DIFF
--- a/ymir/agents/prompts/backport/instructions.j2
+++ b/ymir/agents/prompts/backport/instructions.j2
@@ -76,9 +76,11 @@ a backport that won't actually fix the shipped RPM.
        PatchN: <PACKAGE>-<VERSION>-<CVE_ID>.patch
    where <VERSION> is the upstream version from the spec file's Version field.
 
-1. Knowing Jira issue <JIRA_ISSUE>, CVE ID <CVE_ID> or both, use the `git_log_search` tool to check
-   in the dist-git repository whether the issue/CVE has already been resolved. If it has,
+1. Knowing Jira issue <JIRA_ISSUE>, CVE ID(s) <CVE_ID> or both, use the `git_log_search` tool to check
+   in the dist-git repository whether each issue/CVE has already been resolved. If it has,
    end the process with `success=True` and `status="Backport already applied"`.
+   Note: <CVE_ID> may contain multiple CVE IDs when the Jira issue covers multiple CVEs.
+   The `git_log_search` tool handles multiple CVE IDs automatically.
 
 2. Use the `git_prepare_package_sources` tool to prepare package sources in directory <UNPACKED_SOURCES>
    for application of the upstream patch.
--- a/ymir/agents/prompts/triage/output_format.j2
+++ b/ymir/agents/prompts/triage/output_format.j2
@@ -21,6 +21,8 @@ Backport resolution:
     }
 }
 ```
+Note: When the Jira issue covers multiple CVEs, include ALL CVE IDs in `cve_id`, e.g.:
+`"cve_id": "CVE-1234-98765 CVE-1234-98766"`
 
 Rebase resolution:
 ```json
@@ -36,6 +38,8 @@ Rebase resolution:
     }
 }
 ```
+Note: When the Jira issue covers multiple CVEs, include ALL CVE IDs in `cve_id`, e.g.:
+`"cve_id": "CVE-1234-98765 CVE-1234-98766"`
 
 Rebuild resolution:
 ```json
@@ -53,6 +57,8 @@ Rebuild resolution:
     }
 }
 ```
+Note: When the Jira issue covers multiple CVEs, include ALL CVE IDs in `cve_id`, e.g.:
+`"cve_id": "CVE-1234-98765 CVE-1234-98766"`
 
 Postponed resolution (rebuild waiting for dependency):
 ```json
@@ -70,6 +76,8 @@ Postponed resolution (rebuild waiting for dependency):
     }
 }
 ```
+Note: When the Jira issue covers multiple CVEs, include ALL CVE IDs in `cve_id`, e.g.:
+`"cve_id": "CVE-1234-98765 CVE-1234-98766"`
 
 Not-affected resolution (CVE does not apply):
 ```json
--- a/ymir/common/models.py
+++ b/ymir/common/models.py
@@ -141,7 +141,10 @@ class BackportInputSchema(BaseModel):
     package: str = Field(description="Package to update")
     dist_git_branch: str = Field(description="Git branch in dist-git to be updated")
     jira_issue: str = Field(description="Jira issue to reference as resolved")
-    cve_id: str | None = Field(default=None, description="CVE ID if the jira issue is a CVE")
+    cve_id: str | None = Field(
+        default=None,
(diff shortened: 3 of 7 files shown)


ymir/tools/unprivileged/tests/unit/test_wicked_git.py
```diff
@@ -203,6 +203,16 @@ async def test_git_patch_creation_tool_with_hideous_patch_file(git_repo, tmp_pat
         ("CVE-2025-12346", "", "No matches found for 'CVE-2025-12346'"),
         ("", "RHEL-123456", "Found 1 matching commit(s) for 'RHEL-123456'"),
         ("", "RHEL-123457", "No matches found for 'RHEL-123457'"),
+        (
+            "CVE-2025-12345 CVE-2025-99999",
+            "",
+            "Found 1 matching commit(s) for 'CVE-2025-12345, CVE-2025-99999'",
+        ),
```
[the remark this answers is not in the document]
antbob: the OR approach is intentional here bc this is what we want ie we want the tool to find related commits. the tool is not a verification gate. it passes what it finds to the agent. it is then up to the agent to verify the findings and act accordingly.

lbarcziova (approved):
this looks straightforward, thanks for implementing it!

ymir/tools/unprivileged/wicked_git.py
```diff
@@ -12,6 +12,10 @@
 from ymir.common.version_utils import parse_branch_name
 from ymir.tools.base import CloneableTool as Tool
 
+# Pattern to extract CVE IDs. Needs to handles space-separated, comma-separated,
```
nforro, line 15:
```suggestion
# Pattern to extract CVE IDs. Needs to handle space-separated, comma-separated,
```

ymir/agents/prompts/triage/output_format.j2
```diff
@@ -36,6 +38,8 @@ Rebase resolution:
     }
 }
 ```
+Note: When the Jira issue covers multiple CVEs, include ALL CVE IDs in `cve_id`, e.g.:
+`"cve_id": "CVE-1234-98765 CVE-1234-98766"`
```
nforro, line 42: `RebaseData` has no `cve_id`.

ymir/common/models.py
```diff
@@ -241,7 +244,10 @@ class BackportData(BaseModel):
         "Do NOT repeat the justification rationale here.",
     )
     jira_issue: str = Field(description="Jira issue identifier")
-    cve_id: str | None = Field(description="CVE identifier", default=None)
+    cve_id: str | None = Field(
+        description="CVE identifier(s); include ALL CVE IDs when the issue covers multiple CVEs",
+        default=None,
+    )
```
nforro, line 250: Shouldn't the same change be done for `RebuildData` and `PostponedData`?

ymir/agents/prompts/backport/instructions.j2
```diff
@@ -76,9 +76,11 @@ a backport that won't actually fix the shipped RPM.
        PatchN: <PACKAGE>-<VERSION>-<CVE_ID>.patch
```
nforro, line 76: This should be updated for multi-CVE patches (we probably want to avoid spaces in filenames at least).

ymir/tools/unprivileged/wicked_git.py
nforro, line 555: This should be updated as well.

ymir/agents/prompts/backport/instructions.j2
```diff
@@ -76,9 +76,11 @@ a backport that won't actually fix the shipped RPM.
...
    where <VERSION> is the upstream version from the spec file's Version field.
 
-1. Knowing Jira issue <JIRA_ISSUE>, CVE ID <CVE_ID> or both, use the `git_log_search` tool to check
-   in the dist-git repository whether the issue/CVE has already been resolved. If it has,
+1. Knowing Jira issue <JIRA_ISSUE>, CVE ID(s) <CVE_ID> or both, use the `git_log_search` tool to check
+   in the dist-git repository whether each issue/CVE has already been resolved. If it has,
```
nforro, line 80: Does the tool actually allow to check whether each CVE has been resolved? I'm not sure I follow the logic, but it would seem the tool will return a match for any of multiple CVEs.

nforro (approved):

nforro:
Weird, I didn't mean to approve the changes, but I took too long anyway :sweat_smile:

ymir/agents/prompts/backport/instructions.j2
antbob: no, the tool just does the search. then in triage agent it checks eg (in this example there 9 CVEs in a single Jira):

Let me verify the CVE IDs from the issue labels to make sure I have all nine accounted for.
...
The rpms commit adds 9 individual patch files corresponding to the 9 upstream fix commits for all the CVEs.

then it checks further in applicability agent:

So the shipped source at xorg-server-1.20.11 still has all the vulnerable code for these 9 CVEs.
The package is definitely affected—version 1.20.11 falls within the vulnerable range, all the problematic source files and functions are present, and no existing patches address these specific issues.

ymir/agents/prompts/triage/output_format.j2
antbob: it probably should have it as well. the rebase is often done (at least in our case) to address CVEs that would otherwise be difficult to backport. so presumably it would also need CVE ids to include into the changelog etc. i havent tested the rebase/rebuild yet (just started deploying rebase agent literally today). i think instead of removing the Note i should include "cve_id" field in rebase example, WDYT ?

nforro: > i think instead of removing the Note i should include "cve_id" field in rebase example, WDYT ?

Yes, I think that makes more sense.

ymir/agents/prompts/backport/instructions.j2
antbob: in reality this should not happen but i'll add a note on how to handle  if it does

ymir/agents/prompts/backport/instructions.j2
nforro: This is something that Claude flagged, suggesting `--all-match` may be needed to find only commits that match all CVEs and avoid false positives when checking already applied patches, but TBF I couldn't really wrap my head around it.

antbob: the idea here is for the tool to find what it can then pass it over. triage and applicability agents should then decide if all patches (for all CVEs in scope) are available and applicable. i cant test the cases where either a) a patch is missing (cant be found); or b) the patch is present but not applicable. my expectation (based on the current logic and what i've seen in the agent traces thus far) is that for (a) triage should then result in "clarification needed" and (b) backport with non applicable patches skipped.

PR #633, opened 2026-06-29
Add MR Consolidation Agent for combining backport MRs

TomasKorbar:
Introduces a new agent that automatically consolidates multiple backport
merge requests for the same package-branch into a single comprehensive MR.
    
Key components:
- Redis Hash-based queue with atomic Lua script for concurrent worker
   safety (at-most-one-active, one-pending invariant per package-branch)
- Backport agent integration: submits consolidation jobs after filing MRs
   when enabled via per-package consolidation.json config
- Stale MR filtering: rejects MRs not based on the current target branch
   HEAD, logging skipped MRs as warnings
- LLM-driven patch merging with size-based ordering heuristic (larger
   patch applied first) and automatic context-line adaptation for
   overlapping patches
- Build verification with retry loop (same pattern as backport agent)
- Automatic re-queue when more than 2 MRs need consolidation, chaining
   runs until all are folded into one
- Label-triggered consolidation: maintainers can add ymir_consolidate_base
   and ymir_consolidate_next Jira labels to request consolidation of two
   specific MRs; the fetcher pairs them and submits a targeted job
- Post-consolidation Jira updates: comments the consolidated MR URL on
   all involved Jira issues
- Original MRs are labeled ymir_consolidated (not closed) and excluded
   from future consolidation searches
- Standalone mode via Makefile for manual testing
- E2E tests with cached backport replay and LLM-as-a-judge evaluation
- Architecture documentation for developer reference


docs/mr_consolidation_architecture.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Data Structure
+
+A single Redis **Hash** (`merge_consolidation_queue`) stores all consolidation jobs.
+Each package-branch pair has up to two fields:
+
+| Field pattern              | Meaning                                     |
+|----------------------------|---------------------------------------------|
+| `{pkg}:{branch}:pending`  | A job waiting to be picked up               |
+| `{pkg}:{branch}:active`   | A job currently being processed by a worker |
```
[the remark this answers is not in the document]
TomasKorbar: This rendered just fine.

openshift/deployment-mr-consolidation-agent-c9s.yml
```diff
@@ -0,0 +1,79 @@
...
+      labels:
+        app: mr-consolidation-agent-c9s
+        deployment: mr-consolidation-agent-c9s
+    spec:
+      containers:
+      - args:
+        - -m
+        - ymir.agents.mr_consolidation_agent
+        command:
+        - python
```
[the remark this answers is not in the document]
TomasKorbar: fixed

ymir/agents/mr_consolidation_agent.py
```diff
@@ -0,0 +1,747 @@
...
+                    return "run_consolidation_agent"
+                return "handle_failure"
+
+            return "stage_changes"
+
+        async def stage_changes(state):
+            if state.consolidation_result and state.consolidation_result.files_to_git_add:
+                for f in state.consolidation_result.files_to_git_add:
+                    await check_subprocess(["git", "add", f], cwd=state.local_clone)
+            await run_subprocess(["git", "add", "-A"], cwd=state.local_clone)
```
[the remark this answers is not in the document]
TomasKorbar: fixed.

nforro (commented):
I'm pasting Claude's review without doing a human review of it, since when I do I usually end up being too late to the party and the PR is already merged 😅 But it's probably faster to just let your Claude deal with it anyway.

## Bug: `MERGE_CONSOLIDATION_QUEUE` in `RedisQueues.all_queues()` causes type error

`RedisQueues.all_queues()` returns `{queue.value for queue in cls}` — it includes every enum member. The consolidation queue is a Redis **Hash**, not a List. But `_get_existing_issue_keys_from_redis` in the fetcher calls `lrange` on every queue from `all_queues()`. Calling `lrange` on a Hash key triggers a Redis `WRONGTYPE` error on every fetcher cycle. The per-queue `except Exception` catch prevents a crash, but it produces a spurious warning log on every run.

**Fix options**: Either exclude the consolidation queue from `all_queues()` (rename to exclude it, or filter it out), or move it out of the `RedisQueues` enum since it's semantically different (a hash, not a queue).

## Bug: `JiraLabels.MR_CONSOLIDATED` value doesn't match the GitLab label used in code

`JiraLabels.MR_CONSOLIDATED = "ymir_mr_consolidated"` but the actual label applied in `mark_original_mrs` and filtered in `list_open_mrs` is the string literal `"ymir_consolidated"` (without `_mr_`). The constant is defined but never referenced by the code that applies the label. If someone later refactors to use the constant, the labels won't match. Either the constant value should be `"ymir_consolidated"` or the code should reference the constant. Same issue: `MR_CONSOLIDATION_ERRORED` is defined but unused anywhere in this PR.

## Issue: Undeclared dynamic attributes on Pydantic model

`state._all_open_mrs` (line 968 in `mr_consolidation_agent.py`) and `state._current_mrs_count` (line 1098) are set on `ConsolidationState` but never declared as fields. Pydantic v2 allows underscore-prefixed attributes as private attrs (bypassing validation), so this works at runtime, but it's fragile — these are load-bearing state used across workflow steps. They should be declared as proper fields on `ConsolidationState` (e.g., `all_open_mrs: list[dict] = Field(default_factory=list)` and `current_mrs_count: int = Field(default=0)`).

## Issue: `release_strategy` from `PackageConsolidationConfig` is never propagated

The `PackageConsolidationConfig` model defines a `release_strategy` field, and the architecture doc describes two strategies ("merged" vs "per_commit"). But `fetch_consolidation_config` (called in the backport agent) only checks `config.merge_mrs` — it never reads or forwards `release_strategy`. The `MergeConsolidationJob` model also lacks a `release_strategy` field. The queue-mode worker always uses the default `"merged"`. The config field and the architecture doc section describe a feature that isn't wired up.

## Issue: `test_consolidated_patches_consistent` contradicts the agent's purpose

This test (line 2508) asserts byte-equality between backport patches and consolidated patches: `assert consolidated_patches[patch_name] == patch_content`. But the agent's core job is to **adapt** overlapping patches — the instructions explicitly say to modify context lines when patches touch the same files. The LLM judge evaluator correctly says "NOT by whether it is byte-identical," but this deterministic test would fail for any test case where adaptation was needed. It currently skips for cached backports, so it may not fire yet, but it will produce false failures once live backport test cases exercise overlapping patches.

## Nit: Misleading `noqa` comment in `tasks.py`

```python
from ymir.common.merge_queue import (  # noqa: F401 — re-exported for backward compatibility
```

This is brand-new code — there is no backward compatibility concern. The comment should say something like "re-exported for use by unit tests and consolidation agent" or just drop the explanation.

## Nit: `MergeConsolidationJob` docstring has wrong Redis key name

The docstring says `merge_consolidation_jobs` but the actual key is `merge_consolidation_queue` (from `RedisQueues.MERGE_CONSOLIDATION_QUEUE.value`).

## Nit: `CloseMergeRequestTool` is registered but unused

`CloseMergeRequestTool` is added to the gateway (6 lines in `gateway.py`, 55 lines in `gitlab.py`), but the architecture doc explicitly states "Original MRs are labeled ymir_consolidated (not closed)" and the workflow only calls `add_merge_request_labels`. This is dead code in this PR. If it's intended for future use, it could be added in a separate PR when needed.

## Observation: `submit_merge_job` check-then-set is not atomic

The pending-exists check and subsequent `hset` in `submit_merge_job` are two separate Redis calls. A race between two concurrent backport agents could result in both seeing no pending key and both creating one (second overwrites first). Since auto-mode jobs are equivalent (the consolidation agent queries GitLab regardless), this is harmless for auto-mode. But for label-triggered mode with different `source_issues`, a race could lose the first job's issue keys. The risk is low given the fetcher runs as a single instance, but worth noting.

## Observation: `ListProjectMergeRequestsTool` has no pagination

The tool fetches at most 100 MRs (`per_page: "100"`) with no pagination loop. Unlikely to be a problem for a single package/branch, but worth documenting the limit.

TomasKorbar:
@nforro I addressed your Claudes suggestions :D all should be good now

ymir/agents/prompts/mr_consolidation/instructions.j2
```diff
@@ -0,0 +1,75 @@
+You are an expert on consolidating multiple merge requests for RPM packages in the RHEL ecosystem.
+
+You are given two merge request branches for package <PACKAGE> targeting dist-git branch
+<DIST_GIT_BRANCH>. Your task is to combine the changes from both MRs into a single
+coherent changeset on a new branch.
+
+To consolidate the two MRs, follow these steps:
+
+1. Examine both source branches to understand what each MR does:
```
lbarcziova, line 9: what about adding instruction to also pay attention to the triage summary in each MR (it might include additional context)

ymir/agents/tests/e2e/mr_consolidation/test_mr_consolidation.py
```diff
@@ -0,0 +1,431 @@
...
+
+    evaluator = ConsolidationEvaluator()
+    context = {
+        "package": test_case.backport_issues[0]["package"],
+        "issues": issues_context,
+    }
+
+    verdict = asyncio.run(evaluator.evaluate(artifacts, context))
+
+    assert verdict.passed, f"LLM judge FAILED for {test_case.name}:\n{verdict.reasoning}"
```
TomasTomecek, line 431: you're the best, even adding an e2e test case

ymir/common/constants.py
```diff
@@ -161,6 +164,11 @@ class JiraLabels(Enum):
 
     RETRY_NEEDED = "ymir_retry_needed"
 
+    MR_CONSOLIDATED = "ymir_consolidated"
```
lbarcziova, line 167: iiuc this is always just backports, right? Then I would maybe rename this to `ymir_consolidated_backports` or similar.

ymir/agents/tasks.py
```diff
@@ -610,3 +637,35 @@ async def clone_and_prep_sources(
...
+            "get_maintainer_rules",
+            package=package,
+            file_path="consolidation.json",
+            available_tools=available_tools,
+        )
+        if "not found" in raw.lower():
+            return PackageConsolidationConfig()
+        return PackageConsolidationConfig.model_validate_json(raw)
+    except Exception as e:
+        logger.warning("Failed to fetch consolidation config for %s: %s", package, e)
```
lbarcziova: for future this should be probably propagated somehow to user, if the config is wrong (not conforming to schema)

ymir/common/models.py
```diff
@@ -688,6 +682,88 @@ class CachedMRMetadata(BaseModel):
...
+    """Controls how the package release number is handled during consolidation."""
+
+    MERGED = "merged"
+    PER_COMMIT = "per_commit"
+
+
+class PackageConsolidationConfig(BaseModel):
+    """Machine-readable consolidation config from the per-package rules repo.
+
+    Fetched from ``gitlab.com/redhat/centos-stream/rules/<package>/consolidation.json``.
```
lbarcziova: I'm still not sure if we shouldn't go with something more generic, like ymir.json/ymir.yaml, for the case of future extensions, to not need additional files?

lbarcziova (dismissed):
amazing work, thank you so much! 🚀 

few remarks, but I don't want to block this so that we can check it out on prod ASAP. The most important is the one about config file name.

docs/mr_consolidation_architecture.md
```diff
@@ -0,0 +1,327 @@
...
+If branch info is pre-populated (e.g., in tests or direct invocation with
+`backport_branches`), this step is skipped entirely.
+
+### Step 2: `fork_and_prepare_dist_git`
+
+Forks the upstream dist-git repository (or reuses an existing fork), clones it locally,
+and checks out the target branch. Then fetches all candidate MR branches into the local
+clone.
+
+#### Stale MR Filtering
```
TomasTomecek, line 126: I understand this requirement but am also concerned that people would prefer to also include stale MRs, let's see in the next few months in the feedback how this works

ymir/common/constants.py
nforro: I don't suppose there would be two different rebases for a single release, but a rebase and a backport (or more) are IMHO quite possible (even if the backport is effectively obsoleted by the rebase).

docs/mr_consolidation_architecture.md
```diff
@@ -0,0 +1,327 @@
...
+found after HEAD-filtering. Each successful consolidation marks 2 MRs as
+`ymir_consolidated` (excluding them from future searches) and creates 1 new MR,
+reducing the eligible count by 1. The workflow only requeues when `count - 2 >= 1`,
+i.e., when 3+ MRs were present.
+
+## Per-Package Configuration
+
+Consolidation is enabled by default for all packages. The backport agent reads
+`consolidation.json` from the per-package rules repository
+(`gitlab.com/redhat/centos-stream/rules/<package>/consolidation.json`) before
```
TomasTomecek: I could totally bikeshed about naming here 😅 but I think this is good enough; since i't just 2 options, json is okay but yaml would be definitely a better long term solution

TomasTomecek (dismissed):
Let's go!!! this is amazing

ymir/common/models.py
nforro: Agreed, or at least add a `ymir_` prefix.

docs/mr_consolidation_architecture.md
nforro: +1 for YAML, if humans are supposed to interact with it.

ymir/agents/prompts/mr_consolidation/instructions.j2
TomasKorbar: I feel like the commits and their messages are enough for now, because the process did not have problem neither with Expat or Vim. Lets try it like this and add more info if we find a case of failure.

ymir/common/constants.py
TomasKorbar: And there even could be pure spec change that would have to be consolidated with backport, so perhaps ymir_consolidated is more appropriate 🤷‍♂️

ymir/agents/tasks.py
TomasKorbar: Fixed. A comment will now be posted in Jira.

ymir/common/models.py
TomasKorbar: Done. It will be ymir.yaml

docs/mr_consolidation_architecture.md
TomasKorbar: Done. It will be ymir.yaml

docs/mr_consolidation_architecture.md
TomasKorbar: Perhaps we will extend this if neccessary then.

nforro (approved):

PR #635, opened 2026-06-29
Pre-fetch internal RHEL build version for CVE triage

opohorel:
When a Z-stream CVE needs an internal RHEL fix, the triage agent checks CentOS Stream for the downstream version — but the internal branch may already carry a newer build (e.g. a security update). This causes the agent to propose unnecessary backports for already-fixed packages.

Query Brew for the latest candidate build on the internal RHEL branch before triage runs and inject the version into the prompt so the agent uses the correct downstream version.

### DIFF
--- a/ymir/agents/prompts/triage/prompt.j2
+++ b/ymir/agents/prompts/triage/prompt.j2
@@ -244,6 +244,15 @@ You must decide between one of the following actions. Follow these guidelines to
 
    2.4. Decide the Outcome
    {% if not is_older_zstream %}
+   {% if internal_build_version %}
+   * **IMPORTANT — Internal RHEL build version**: The latest build of this
+     package on the internal RHEL branch is version **{{ internal_build_version }}**,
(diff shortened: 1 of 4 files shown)


TomasTomecek (approved):
LGTM, definitely a big improvement so the triage agent now has a better context

lbarcziova (commented):
shouldn't this be covered by the applicability agent? It checks out the particular branch where the fix should go, so should have the needed data

https://github.com/packit/ai-workflows/pull/632 from this morning might be related to this as well

opohorel:
> shouldn't this be covered by the applicability agent? It checks out the particular branch where the fix should go, so should have the needed data

It is, but applicability then posts a comment where it says Not Affected/VEX, completely avoiding the fact that this was already fixed, which is a bit misleading. When triage finds this out first it shouldn't post a comment like this

opohorel:
Closing in favor of #637

PR #637, opened 2026-06-29
Use authenticated clone for internal RHEL branch triage

opohorel:
Use authenticated clone for internal RHEL branch triage

Previously the triage agent always inspected the CentOS Stream
branch (e.g. c10s), missing newer versions on the internal RHEL
branch and proposing unnecessary backports.

Pre-compute the internal target branch from fix_version and CVE
eligibility, pass it to the triage prompt, and instruct the agent
to use the authenticated clone_repository tool for the RHEL
namespace with a CentOS Stream fallback when the branch does not
exist.

Assisted-by: Cursor (Claude)

### DIFF
--- a/ymir/agents/prompts/triage/prompt.j2
+++ b/ymir/agents/prompts/triage/prompt.j2
@@ -37,9 +37,16 @@ Goal: Analyze the given issue to determine the correct course of action.
 
 2. Identify the package name that must be updated:
    * Determine the name of the package from the issue details (usually component name)
+   {% if needs_internal_fix and internal_target_branch %}
+   * Use the clone_repository tool to clone `https://gitlab.com/redhat/rhel/rpms/<package_name>`
+     with branch `{{ internal_target_branch }}`.
+     If the clone fails (branch or repo not found), fall back to cloning
+     `https://gitlab.com/redhat/centos-stream/rpms/<package_name>` using the same tool.
+   {% else %}
    * Confirm the package repository exists by running
      `GIT_TERMINAL_PROMPT=0 git ls-remote https://gitlab.com/redhat/centos-stream/rpms/<package_name>`
-   * A successful command (exit code 0) confirms the package exists
+     A successful command (exit code 0) confirms the package exists.
+   {% endif %}
    * If the package does not exist, re-examine the Jira issue
      for the correct package name and if it is not found,
      return error and explicitly state the reason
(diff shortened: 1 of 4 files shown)


ymir/agents/prompts/triage/prompt.j2
```diff
@@ -37,8 +37,15 @@ Goal: Analyze the given issue to determine the correct course of action.
 
 2. Identify the package name that must be updated:
    * Determine the name of the package from the issue details (usually component name)
+   {% if needs_internal_fix and internal_target_branch %}
    * Confirm the package repository exists by running
-     `GIT_TERMINAL_PROMPT=0 git ls-remote https://gitlab.com/redhat/centos-stream/rpms/<package_name>`
+     `GIT_TERMINAL_PROMPT=0 git ls-remote --exit-code https://gitlab.com/redhat/rhel/rpms/<package_name> {{ internal_target_branch }}`
```
lbarcziova: since this is private (as opposed to centos-stream)/needs authentication, have you tested if this work out of the box?

lbarcziova (commented):
makes a lot of sense for triage to know this 👍

ymir/agents/prompts/triage/prompt.j2
opohorel: that is a great point, and I'll have to investigate more, because that would fail consistently, and we would use the fallback all the time

TomasKorbar (commented):
+1 to Lauras point, otherwise LGTM.

ymir/agents/prompts/triage/prompt.j2
opohorel: I've reworked it so now it uses the privileged clone tool. Also ran the e2e triage tests and there is no regression found.

ymir/agents/prompts/triage/prompt.j2
```diff
@@ -37,9 +37,16 @@ Goal: Analyze the given issue to determine the correct course of action.
 
 2. Identify the package name that must be updated:
    * Determine the name of the package from the issue details (usually component name)
+   {% if needs_internal_fix and internal_target_branch %}
+   * Use the clone_repository tool to clone `https://gitlab.com/redhat/rhel/rpms/<package_name>`
```
lbarcziova, line 41: what will be the default behaviour here for cloning path? Should we instruct to use /tmp?

opohorel: that will be decided by the agent. I didn't look into it further, because now the agents are not instructed to clone anything and they are doing it anyway. I've only checked that it is not going to collide with applicability agent cloning the repo

lbarcziova: ok, let's try then, and if this leads to some unexpected issues, we could instruct for a particular location

lbarcziova (approved):

PR #638, opened 2026-06-30
fix(sentry): ignore LLM logging and use Sentry logs

mfocko:
Based on the changes in the later mentioned PR, the current logging outputs logged strings line-by-line which causes issues with the Sentry logging integration as, for example, the dumped JSONs produce entries even for closing tokens such as `}`.

Therefore adjust the setup to ignore the logging from the affected logger and log explicitly via Sentry SDK, so that it is formatted properly.

Related to #630

### DIFF
--- a/ymir/agents/utils.py
+++ b/ymir/agents/utils.py
@@ -241,5 +241,7 @@ def init_sentry() -> None:
             LiteLLMIntegration(),
         ],
     )
-    ignore_logger("agent.redis")
-    ignore_logger_for_sentry_logs("agent.redis")
+
+    for ignored_logger in ("agent.redis", "agent.task_loop", "agent.trajectory"):
+        ignore_logger(ignored_logger)
+        ignore_logger_for_sentry_logs(ignored_logger)


ymir/common/logging_setup.py
```diff
@@ -104,6 +106,7 @@ def write(self, s: str) -> int:
         if len(lines) > 1 and _buffered_handler is not None:
             _buffered_handler.begin_group()
         try:
+            sentry_sdk.logger.info(s)
```
nforro: I'm still not convinced we need this. Before my [PR](https://github.com/packit/ai-workflows/pull/630), this wasn't logged at all, and I think the `LiteLLMIntegration` already covers most of it.

nforro (dismissed):
Thanks!

ymir/common/logging_setup.py
mfocko: so are you more for skipping it altogether?

mfocko: > and I think the `LiteLLMIntegration` already covers most of it.

this is broken :/ I don't really know why, but if you check the “Conversations” in Sentry, there's nothing, even though I was able to find the chats from the spans :D I don't really get what's wrong with it, the only difference in between what is done in Ymir and what they suggest on the dashboard is the conversation ID:

```py
import sentry_sdk

# Call this at the start of each conversation
sentry_sdk.ai.set_conversation_id("my-conversation-123")
```

OTOH when you check the docs, they don't use that :D

I can probably check the versions if it's not because of that, but IDK.

nforro: > so are you more for skipping it altogether?

I'm fine either way, it could be beneficial to have more context if the "conversations" don't work.

mfocko: OK, I can disable for now, since it wasn't logged before and we can decide on the follow up?

Especially after looking at the failed CI, it needs the Sentry SDK too, so it's more annoying than I expected

lbarcziova (dismissed):

ymir/common/logging_setup.py
mfocko: Ignoring for now, will either open a follow-up issue or a PR right away.

nforro (approved):

PR #642, opened 2026-06-30
Copr API improvements to retries and diagnostics

antbob:
This patch revamps the retry mechanism when Copr API is failing (eg 5xx errors and such) and also adds proper diagnostic logging and error propagation so these types of errors can be seen clear in both the logs and the agent output. It also avoid token burn that is currently happening eg when Copr API returns 5xx errors the loop just keeps retrying up to 10 times and then the agent would retry up to 3 times on top of that. The later (3x retries) is not addressed in this patch as this needs to be tackled separately by rethinking how the agents handle infra failures. Another commit in this patch is somewhat related (tool_use/tool_result  consistency) but i kept it as separate commit as its not specific to Copr API.

### DIFF
--- a/ymir/agents/rebase_agent.py
+++ b/ymir/agents/rebase_agent.py
@@ -158,7 +158,6 @@ async def main() -> None:
             os.environ["MCP_GATEWAY_URL"], call_meta={"jira_issue": jira_issue}
         ) as gateway_tools:
             rebase_agent = create_rebase_agent(gateway_tools, local_tool_options)
-            build_agent = create_build_agent(gateway_tools, local_tool_options)
             log_agent = create_log_agent(gateway_tools, local_tool_options)
 
             workflow = Workflow(State, name="RebaseWorkflow")
@@ -225,6 +224,7 @@ async def main() -> None:
                 return "comment_in_jira"
 
             async def run_build_agent(state):
+                build_agent = create_build_agent(gateway_tools, local_tool_options)
                 response = await build_agent.run(
                     render_template(
                         get_build_prompt(),
@@ -243,6 +243,11 @@ async def main() -> None:
                 if build_result.is_timeout:
                     logger.info(f"Build timed out for {state.jira_issue}, proceeding")
                     return "update_release"
+                if build_result.is_infra_error:
+                    logger.error(f"Copr infrastructure error for {state.jira_issue}: {build_result.error}")
+                    state.rebase_result.success = False
+                    state.rebase_result.error = build_result.error or "Copr API infrastructure error"
+                    return "comment_in_jira"
                 state.attempts_remaining -= 1
                 if state.attempts_remaining <= 0:
                     state.rebase_result.success = False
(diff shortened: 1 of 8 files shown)


ymir/tools/privileged/copr.py
```diff
@@ -41,14 +43,44 @@
...
+            delay = COPR_API_RETRY_BACKOFF_BASE * (2**attempt) + random.uniform(0, 1)  # noqa: S311
+            logger.warning(
+                "Copr API error (%s), retrying in %.1fs (attempt %d/%d)",
+                _copr_error_detail(e),
+                delay,
+                attempt + 1,
+                COPR_API_MAX_RETRIES,
+            )
+            await asyncio.sleep(delay)
+    raise RuntimeError("unreachable")
```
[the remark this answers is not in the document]
antbob: no, this is intentional as our infra can have what you call "non-transient errors" as actually transient errors so its still worth re-trying.

ymir/tools/privileged/copr.py
```diff
@@ -41,14 +43,45 @@
...
+
+
 class BuildResult(BaseModel):
     success: bool = Field(description="Whether the build succeeded")
-    is_timeout: bool = Field(default=False, description="Whether the build failed due to a timeout")
```
nforro, line 51: Why was this removed?

TomasTomecek (dismissed):
LGTM, nicely done, though needs a test fix

ymir/agents/rebase_agent.py
```diff
@@ -158,7 +158,6 @@ async def run_workflow(
             os.environ["MCP_GATEWAY_URL"], call_meta={"jira_issue": jira_issue}
         ) as gateway_tools:
             rebase_agent = create_rebase_agent(gateway_tools, local_tool_options)
-            build_agent = create_build_agent(gateway_tools, local_tool_options)
```
nforro, line 161: Could you please explain this move?

antbob: otherwise it becomes stale. here is the full explanation (asked Claude to summarize):
The problem: when a ReasoningAgent is reused, self.memory.messages carries over from the previous run. So on the second build attempt, the build agent's memory still contained the full conversation from the first attempt — stale tool calls, old error messages, old build results. This contaminated the LLM's context and caused it to behave incorrectly (e.g., thinking it had already built when it hadn't).
Creating a fresh agent each time gives it a clean UnconstrainedMemory() with no prior history. This matches the pattern already used in the backport agent.

nforro: Ok, I suppose that makes sense, although the backport agent implementation is a mess. What about merge request agent though, doesn't it have the same issue?

ymir/tools/privileged/copr.py
antbob: good catch, this was a mistake! i'm merging this from earlier version so bunch of stuff doesnt match. there were some changes related to timeout that didnt match but this patch should not touch or change them.

ymir/agents/rebase_agent.py
antbob: yep, same problem apparently, lets fix that as well then

nforro (approved):
LGTM

PR #653, opened 2026-07-03
Smarter download_artifacts tool

jpodivin:
The tool has been rewritten to allow less space for errors. Instead of letting agent choose where to download files to, which is also a security hazard, the agent can only choose what to download. Upon invocation, the tool creates a temporary directory, and all artifacts are downloaded there.

Path to the temporary directory is returned upon successful tool run. If the tool fails, the temporary directory and all contents are removed.

The cleanup step is, arguably, an overkill. Name of the temporary dir is generated randomly, and unless the agent were to get into a particularly vicious loop, it is unlikely to run out possible names.

The tool output has been written with custom `get_text_content` method, to cut down number of tokens used. It is very much a question of taste, as we are saving handful of tokens per call at best.

Fixes

PACKIT-5157: Download artifacts failing due to nonexistent directory 

RELEASE NOTES BEGIN

The `download_artifacts` tool now uses temporary directories to store downloaded artifacts, and does not allow agent to choose where artifacts are downloaded.

RELEASE NOTES END

### DIFF
--- a/ymir/tools/privileged/copr.py
+++ b/ymir/tools/privileged/copr.py
@@ -2,8 +2,10 @@ import asyncio
 import gzip
 import logging
 import random
+import tempfile
 import time
 from pathlib import Path
+from shutil import rmtree
 from urllib.parse import urljoin, urlparse
 
 import aiohttp
@@ -12,7 +14,6 @@ from beeai_framework.context import RunContext
 from beeai_framework.emitter import Emitter
 from beeai_framework.tools import (
     JSONToolOutput,
-    StringToolOutput,
     ToolError,
     ToolRunOptions,
 )
@@ -324,13 +325,22 @@ class BuildPackageTool(Tool[BuildPackageToolInput, ToolRunOptions, BuildPackageT
 
 class DownloadArtifactsToolInput(BaseModel):
     artifacts_urls: list[str] = Field(description="URLs to build artifacts (logs and RPM files)")
-    target_path: AbsolutePath = Field(description="Absolute path where to download the artifacts")
 
 
-class DownloadArtifactsTool(Tool[DownloadArtifactsToolInput, ToolRunOptions, StringToolOutput]):
+class DownloadArtifactsResult(BaseModel):
+    target_path: Path = Field(description="Location of downloaded files")
+
+
+class DownloadArtifactsToolOutput(JSONToolOutput[DownloadArtifactsResult]):
+    def get_text_content(self) -> str:
+        """Return content with minimum tokens possible"""
+        return f"target_path: {self.result.target_path}"
+
+
+class DownloadArtifactsTool(Tool[DownloadArtifactsToolInput, ToolRunOptions, DownloadArtifactsToolOutput]):
     name = "download_artifacts"
     description = """
-    Downloads build artifacts to the specified location.
+    Downloads build artifacts to a temporary location.
     Any gzipped log files will be automatically decompressed,
     for example `http://example.com/builder-live.log.gz`
     will be downloaded as `builder-live.log`.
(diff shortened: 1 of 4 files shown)


ymir/tools/privileged/copr.py
```diff
@@ -369,7 +379,11 @@ async def _run(
                                     target = target.with_suffix("")
                             (target_path / target).write_bytes(content)
                         else:
+                            # Cleanup temporary dir
+                            rmtree(target_path)
                             raise ToolError(f"Failed to download {url}: {response.status} {response.reason}")
                 except (aiohttp.ClientError, TimeoutError) as e:
+                    # Cleanup temporary dir
+                    rmtree(target_path)
```
nforro, line 385: This should be in a `finally` block to cover unhandled exceptions and avoid duplication.

ymir/tools/privileged/copr.py
```diff
@@ -348,9 +358,9 @@ async def _run(
         tool_input: DownloadArtifactsToolInput,
         options: ToolRunOptions | None,
         context: RunContext,
-    ) -> StringToolOutput:
+    ) -> DownloadArtifactsToolOutput:
         artifacts_urls = tool_input.artifacts_urls
-        target_path = tool_input.target_path
+        target_path = tempfile.mkdtemp()
```
nforro:
```suggestion
        target_path = Path(tempfile.mkdtemp())
```

nforro (commented):
The prompt needs to be updated:
https://github.com/packit/ai-workflows/blob/505deccd5de5547040e770acad7eeb83fc247d8f/ymir/agents/prompts/build/instructions.j2#L4-L5

ymir/tools/privileged/copr.py
jpodivin: This is not really necessary. The `/` works just fine with string and `Path` objects. But I suppose it can improve readability.

ymir/tools/privileged/copr.py
jpodivin: But I want to keep the dir around if the operation was successful, so that other tools can read from it. I can expand the catch block so that it activates for all exceptions though.

ymir/tools/privileged/copr.py
nforro: I know, but Claude insisted on it being a bad practice. Though asking it now it can't explain why.

ymir/tools/privileged/copr.py
nforro: Sorry, somehow I didn't realize `finally` would be executed also on success :sweat_smile: 

> I can expand the catch block so that it activates for all exceptions though.

Thanks.

ymir/tools/privileged/copr.py
jpodivin: It's not a big deal. I'll adjust the result model.

ymir/tools/privileged/copr.py
```diff
@@ -369,7 +379,9 @@ async def _run(
                                     target = target.with_suffix("")
                             (target_path / target).write_bytes(content)
                         else:
-                            raise ToolError(f"Failed to download {url}: {response.status} {response.reason}")
-                except (aiohttp.ClientError, TimeoutError) as e:
+                            raise ValueError(f"{response.status} {response.reason}")
```
nforro, line 382: I would maybe use `RuntimeError` here, but that's just nitpicking.

nforro (commented):
What about the prompt? `to the current directory` should be removed.

ymir/tools/privileged/copr.py
jpodivin: It's going to get wrapped in tool error anyway.

jpodivin:
> What about the prompt? `to the current directory` should be removed.

That wasn't really true at any point. The agent was downloading the the artifact where it wanted, and not in the current directory. But sure, why not.

nforro (approved):

PR #656, opened 2026-07-06
Adapt rebuild agent to older z-streams

antbob:
We just started using the rebuild agent for older z-streams and these are the changes required to make it work for us. The inclusion of the side-tag is required so it can be propagated to MR and trigger pipeline automatic side-tag selection when building (the current requirement is that its lives on its own line in the MR description, clearly separated from everything else). The stage_changes logic requires adjustment because update_release changes need to be staged before run_log_agent makes further changes otherwise it makes a mess by updating the existing changelog entry (since its not aware of the update release changes otherwise).

### DIFF
--- a/ymir/agents/prompts/triage/output_format.j2
+++ b/ymir/agents/prompts/triage/output_format.j2
@@ -53,12 +53,15 @@ Rebuild resolution:
         "triage_summary": "Confirmed package vendors Go deps via spec BuildRequires.",
         "dependency_issue": "RHEL-67890",
         "dependency_component": "golang",
-        "fix_version": "rhel-X.Y.Z"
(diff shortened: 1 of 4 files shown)


nforro (approved):
LGTM, but it is not possible to specify a side-tag when running the rebuild agent in direct mode.

antbob:
@nforro yeah claude review flagged that as well but we dont use it and frankly i have no idea what its for. is that replacing agents as skills or its structly for testing ? i can easily add it there as well if needed of course.

nforro:
> is that replacing agents as skills or its structly for testing ?

No, it's been there since the beginning, and it's for local runs of individual agents/workflows. Mainly useful for debugging.

lbarcziova (approved):
thanks!

PR #659, opened 2026-07-08
Fix buildroot check for c8s z-stream rebuilds

opohorel:
c8s is EOL — builds always go to Brew, never CentOS Stream Koji. Skip the stale c8s-build tag check and only verify the Brew z-stream buildroot (rhel-8.10.0-z-build).

Assisted-by: Claude (Cursor)

### DIFF
--- a/ymir/common/utils.py
+++ b/ymir/common/utils.py
@@ -27,7 +27,11 @@ from specfile.utils import EVR
 from ymir.common.base_utils import is_cs_branch
 from ymir.common.constants import BREWHUB_URL, CENTOS_STREAM_KOJIHUB_URL
 from ymir.common.logging_setup import get_trajectory_writeable
-from ymir.common.version_utils import construct_internal_branch_name, parse_rhel_version
+from ymir.common.version_utils import (
+    construct_internal_branch_name,
+    get_maintenance_majors,
+    parse_rhel_version,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


ymir/common/utils.py
```diff
@@ -227,18 +227,24 @@ def _resolve_buildroot_checks(target_branch: str, fix_version: str) -> list[tupl
...
             rhel_branch = construct_internal_branch_name(major, minor)
-            checks.append((BREWHUB_URL, f"{rhel_branch}-z-build"))
-        return checks
+            if target_branch == "c8s":
+                return [(BREWHUB_URL, f"{rhel_branch}-z-build")]
+            return [
+                (CENTOS_STREAM_KOJIHUB_URL, f"{target_branch}-build"),
+                (BREWHUB_URL, f"{rhel_branch}-z-build"),
+            ]
+        return [(CENTOS_STREAM_KOJIHUB_URL, f"{target_branch}-build")]
```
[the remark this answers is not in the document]
opohorel: rhel 8 has no Y stream, so this shouldn't happen

ymir/common/utils.py
```diff
@@ -227,18 +227,24 @@ def _resolve_buildroot_checks(target_branch: str, fix_version: str) -> list[tupl
...
     is_zstream = fix_version.lower().endswith(".z")
 
     if is_cs_branch(target_branch):
-        checks = [(CENTOS_STREAM_KOJIHUB_URL, f"{target_branch}-build")]
         if is_zstream and (parsed := parse_rhel_version(fix_version)):
             major, minor, _ = parsed
             rhel_branch = construct_internal_branch_name(major, minor)
-            checks.append((BREWHUB_URL, f"{rhel_branch}-z-build"))
-        return checks
+            if target_branch == "c8s":
```
lbarcziova: I think we have already some logic for checking the maintenance version, could we use it here instead of hardcoding?

opohorel: good point, I've moved the `get_maintenance_majors()` from jira.py to version_utils.py so we can reuse the function

lbarcziova (approved):
LGTM, thank you!

PR #662, opened 2026-07-08
Add target::zstream label to MRs at creation time

lbarcziova:
Fixes: https://redhat.atlassian.net/browse/PACKIT-5088

Assisted-by: Claude Opus 4.6

TODO:

- [x] test the MR creation

### DIFF
--- a/ymir/agents/mr_consolidation_agent.py
+++ b/ymir/agents/mr_consolidation_agent.py
@@ -980,7 +980,12 @@ async def run_workflow(
                     available_tools=gateway_tools,
                 )
 
-                source_labels = {label for mr in state.all_open_mrs for label in (mr.get("labels") or [])}
+                source_labels = {
+                    label
+                    for mr in state.all_open_mrs
+                    if mr["url"] in state.mr_urls
+                    for label in (mr.get("labels") or [])
+                }
                 labels = ["ymir_backport"]
                 if ZSTREAM_TARGET_LABEL in source_labels:
                     labels.append(ZSTREAM_TARGET_LABEL)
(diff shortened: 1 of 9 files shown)


ymir/tools/privileged/gitlab.py
```diff
@@ -344,6 +349,25 @@ def _create_emitter(self) -> Emitter:
...
+        }
+        if labels:
+            parameters["labels"] = ",".join(labels)
+        if project.is_fork:
+            parameters["target_project_id"] = project.parent.gitlab_repo.attributes["id"]
+        target_project = project.parent if project.is_fork else project
+        try:
+            raw_mr = project.gitlab_repo.mergerequests.create(parameters)
+        except gitlab.GitlabError as ex:
+            raise GitlabAPIException() from ex
```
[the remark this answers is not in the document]
lbarcziova: GitlabAPIException.response_code is a property that reads from self.__cause__, which is set by the from ex clause. The 409 should work correctly through the exception chain

ymir/agents/mr_consolidation_agent.py
```diff
@@ -980,30 +980,25 @@ async def push_and_open_mr(state):
                     available_tools=gateway_tools,
                 )
 
+                source_labels = {label for mr in state.all_open_mrs for label in (mr.get("labels") or [])}
```
nforro: Only two oldest MRs are being consolidated, `source_labels` should come only from them.

ymir/agents/mr_consolidation_agent.py
```diff
@@ -980,30 +980,25 @@ async def push_and_open_mr(state):
...
-                if state.merge_request_url:
-                    try:
-                        await run_tool(
-                            "add_merge_request_labels",
-                            merge_request_url=state.merge_request_url,
-                            labels=["ymir_backport"],
-                            available_tools=gateway_tools,
-                        )
-                    except Exception as e:
-                        logger.warning("Failed to label consolidated MR: %s", e)
```
nforro, line 1006: From Claude:
The old post-creation `add_merge_request_labels` call is removed here, but `open_merge_request` doesn't update labels on the 409 (existing MR) path. The other agents are fine because `tasks.commit_push_and_open_mr` still has its own `add_merge_request_labels` fallback, but the consolidation agent calls `open_merge_request` directly.

nforro (approved):
LGTM

PR #663, opened 2026-07-08
Fix Z-Stream branch creation for EUS branches

nforro:
It's more than 2 lines, but the actual change boils down to:
https://github.com/packit/ai-workflows/blob/fa9a280730dc8415ebbc123938079807b8faf36b/ymir/tools/privileged/distgit.py#L229-L232


majamassarini (approved):
Thanks!

PR #664, opened 2026-07-08
Add MR cleanup cronjob to close stale bot-authored MRs

lbarcziova:
Daily job that scans open MRs by jotnar-bot and redhat-ymir-agent, checks if all referenced Jira issues (from Resolves: lines) are closed, and closes the MR with a comment. Adds ymir_cleaned_up label on closed MRs and skips MRs that already have it (prevents re-closing reopened MRs).

Includes DRY_RUN support (default true for local runs), TARGET_MR for single-MR testing, OpenShift manifests (4am UTC daily), CI build job, and make targets.

Fixes https://redhat.atlassian.net/browse/PACKIT-5167

Assisted-by: Claude Opus 4.6

Tested on https://gitlab.com/redhat/rhel/rpms/delve/-/merge_requests/3

### DIFF
--- a/templates/mr-cleanup.env
+++ b/templates/mr-cleanup.env
@@ -1,0 +1,9 @@
+# gitlab.com -> Edit profile -> Access tokens -> Add new token
+# Permissions: api (needed to close MRs and post notes)
+GITLAB_TOKEN=
+
+JIRA_URL=https://redhat.atlassian.net
+# Jira account email for authentication
+JIRA_EMAIL=
+# Jira API token - https://id.atlassian.com/manage-profile/security/api-tokens
+JIRA_TOKEN=
--- a/ymir/mr_cleanup/mr_cleanup.py
+++ b/ymir/mr_cleanup/mr_cleanup.py
@@ -1,0 +1,370 @@
+#!/usr/bin/env python3
+"""
+MR Cleanup Script
+
+Scans open GitLab merge requests authored by Ymir bots and closes those
+whose referenced Jira issues have all been closed.
+
+No Jira labels are modified — metrics dashboards depend on those labels
+remaining in place.
+"""
+
+import base64
+import enum
+import logging
+import os
+import re
+import sys
+import time
+from urllib.parse import quote as urlquote
+
+import backoff
+import requests
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(name)s %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S",
+    stream=sys.stdout,
+)
+logger = logging.getLogger(__name__)
+
+GITLAB_API_URL = "https://gitlab.com/api/v4"
+GITLAB_GROUPS = ["redhat/rhel/rpms", "redhat/centos-stream/rpms"]
+GITLAB_BOT_AUTHORS = ["jotnar-bot", "redhat-ymir-agent"]
+PER_PAGE = 100
+
+JIRA_CLOSED_STATUS = "Closed"
+JIRA_BATCH_SIZE = 100
+
+RATE_LIMIT_DELAY = 0.2
+REQUEST_TIMEOUT = 90
+
+CLOSE_NOTE_MARKER = "Closing this merge request"
+
+
+class Action(enum.Enum):
+    CLOSED = "closed"
+    SKIPPED_NO_JIRA = "skipped_no_jira"
+    SKIPPED_OPEN_JIRAS = "skipped_open_jiras"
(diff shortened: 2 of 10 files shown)


ymir/mr_cleanup/mr_cleanup.py
```diff
@@ -0,0 +1,365 @@
...
+        commits = self._gitlab_get(
+            f"projects/{project_id}/merge_requests/{iid}/commits",
+            params={"per_page": PER_PAGE},
+        )
+        keys: set[str] = set()
+        for commit in commits:
+            message = commit.get("message", "")
+            for line in message.split("\n"):
+                if "resolves:" in line.lower():
+                    keys.update(re.findall(r"RHEL-\d+", line))
```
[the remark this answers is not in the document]
lbarcziova: this is the correct Jiras format, shouldn't be problematic

ymir/mr_cleanup/mr_cleanup.py
```diff
@@ -0,0 +1,369 @@
...
+        return keys
```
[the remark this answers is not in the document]
lbarcziova: same as above

ymir/mr_cleanup/mr_cleanup.py
```diff
@@ -0,0 +1,369 @@
...
+                    all_mrs.extend(mrs)
+                    page += 1
+
+        if self.target_mr:
+            all_mrs = [mr for mr in all_mrs if mr["web_url"] == self.target_mr]
+            if not all_mrs:
+                logger.warning("TARGET_MR %s not found among open bot MRs", self.target_mr)
+
+        logger.info("Found %d open MRs total", len(all_mrs))
+        return all_mrs
```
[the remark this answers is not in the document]
lbarcziova: this is just for local testing, will not be used in production

templates/mr-cleanup.env
```diff
@@ -0,0 +1,9 @@
+# gitlab.com -> Edit profile -> Access tokens -> Add new token
+# Permissions: read_api, api (needed to close MRs and post notes)
```
mfocko: IIRC `read_api` is subset of `api`

ymir/mr_cleanup/mr_cleanup.py
```diff
@@ -0,0 +1,369 @@
...
+
+        closed_keys = {k for k in jira_keys if jira_statuses.get(k) == JIRA_CLOSED_STATUS}
+        open_keys = jira_keys - closed_keys
+
+        if not closed_keys:
+            logger.debug("All Jiras still open for MR %s (%s)", mr_url, ", ".join(sorted(jira_keys)))
+            return Action.SKIPPED_NONE_CLOSED
+
+        if open_keys:
+            return Action.SKIPPED_NONE_CLOSED
```
mfocko: If I get here, it means that some were closed right? is this intentional?

templates/mr-cleanup.env
lbarcziova: true, adjusted

ymir/mr_cleanup/mr_cleanup.py
lbarcziova: yes; I was originally thinking to post a comment about this also (without closing), but decided to leave this as a follow-up if needed, added a note

lbarcziova:
@antbob this is probably not needed for your deployment, given the opt-in approach on your side, wdyt? I am wondering how to make the deployment more safe so that you don't accidentally deploy this (as it has hardcoded our bot account names), iirc you don't use the https://github.com/packit/ai-workflows/blob/main/openshift/deploy.sh and have your own automation, is that correct?

antbob:
@lbarcziova thanks for headsup! it would get deployed if i'm not careful. we maintain a modified version of deploy so i would need to disable this new cronjob (or adapt it to our workflow later). i dunno how to make these things more safe other than removing the names from here upstream deployment files and using env vars that have to be explicitly set but that sounds like too much of a hassle for everyone.

ymir/mr_cleanup/mr_cleanup.py
lbarcziova: also renamed the enum value to be more clear

jpodivin (approved):

jpodivin:
As long as the API behaves at least somewhat compliant with spec this should work fine.

mfocko (approved):

lbarcziova:
I already ran this locally: https://gitlab.com/groups/redhat/-/merge_requests/?sort=updated_desc&state=closed&author_username=redhat-ymir-agent&label_name%5B%5D=ymir_cleaned_up&first_page_size=20

PR #668, opened 2026-07-09
chore(firewall): add the domains spotted in Sentry

mfocko:
I noticed these reoccurring errors in Sentry for fetching upstream patches. They represent majority of the captured events in the last 7 days.


nforro (approved):
Thanks!

PR #673, opened 2026-07-09
Fix Release bumping for RHEL in maintenance phase

nforro:
RHEL version in maintenance phase (currently RHEL 8.10) is being fixed via CentOS Stream, but the fixes still target an internal Z-Stream branch, so Release bumping must follow Z-Stream logic.

### DIFF
--- a/ymir/common/tests/unit/test_version_utils.py
+++ b/ymir/common/tests/unit/test_version_utils.py
@@ -1,6 +1,8 @@
 import pytest
+from flexmock import flexmock
 
 from ymir.common.version_utils import (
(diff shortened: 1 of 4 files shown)


TomasTomecek (dismissed):

nforro:
Gemini was right, even if for maintenance phase RHEL we don't need the base release at all in the majority of cases, there could be some when it is required to ensure a proper upgrade path, and deriving it from a CentOS Stream build is wrong, as that would correspond to the last Y-Stream build and ignore any follow-up Z-Stream builds.

TomasTomecek (approved):

PR #674, opened 2026-07-10
Rebase agent older z-stream support and restrictions on Sources

antbob:
There are 2 main changes here:

1) Add support for older z-stream re-bases. The leading z-stream, if available, is used as reference. Older z-streams now check out all the branches but that should not be a problem for disk space or processing time as those contain mostly patch files and such and not the actual source code.

2) More restrictive instructions on Source files and their URLs. Currently this is a problem because the agent can DuckDuckGo and bring in whatever junk it finds so whatever sources it ends up with are of an unknown provenance. Not only it can derail the re-base itself but it can also get malicious sources which it then uploads to look-aside cache, effectively poisoning it.

This patch does not solve the problem described in (2) completely, it just tightens up the instructions. The proper solution should be creating a deterministic tool (ie something that can count the r's in strawberry) for this problem that the agent can use to verify the provenance of Source files and URLs, including check-sums etc but it is outside the scope for this patch.


opohorel (approved):
LGTM as is. I would think about following this up with the self-review similar to what we have in the backport agent

PR #677, opened 2026-07-10
Fix multiple MR consolidation agent issues

TomasKorbar:
- Set working_directory on local_tool_options after clone so the LogAgent and consolidation LLM operate in the correct repo instead of defaulting to /home/beeai (caused missing changelog entries)
- Call download_sources after clone and again after base-branch cherry-pick so source tarballs added by MR branches are available for prep and SRPM builds (previously the LLM wasted ~15 min trying to download them manually)
- Add NFS cache wait after post-cherry-pick source download
- Handle spec-only MRs in per_commit_flow: detect branches with no new .patch files, use a dedicated adapt_spec.j2 prompt that feeds the spec diff to the LLM instead of the patch-oriented prompt that silently dropped spec-only changes
- Capture cherry-pick exit code and reset index on conflict
- Guard all run_subprocess .strip() calls against None returns
- Skip NFS wait when no branches were actually fetched via MCP
- Move _NFS_CACHE_WAIT constant to module level
- Add spec-only MR guidance to merged-strategy instructions.j2

### DIFF
--- a/ymir/agents/mr_consolidation_agent.py
+++ b/ymir/agents/mr_consolidation_agent.py
@@ -80,6 +80,8 @@ from ymir.tools.unprivileged.wicked_git import (
 logger = logging.getLogger(__file__)
 redis_logger = logging.getLogger("agent.redis")
 
+_NFS_CACHE_WAIT = 60
+
 
 async def create_consolidation_agent(
     mcp_tools_list: list[Tool],
@@ -357,6 +359,15 @@ async def run_workflow(
                 dist_git_branch=dist_git_branch,
                 available_tools=gateway_tools,
             )
+            local_tool_options["working_directory"] = state.local_clone
+
+            await run_tool(
+                "download_sources",
+                dist_git_path=str(state.local_clone),
+                package=package,
+                dist_git_branch=dist_git_branch,
+                available_tools=gateway_tools,
+            )
 
             namespace = "centos-stream" if is_cs_branch(dist_git_branch) else "rhel"
             repo_url = f"https://gitlab.com/redhat/{namespace}/rpms/{package}"
@@ -370,6 +381,7 @@ async def run_workflow(
                 state.mr_branches if state.mr_branches else [mr["source_branch"] for mr in state.all_open_mrs]
             )
 
+            any_fetched = False
             for branch_name in candidate_branches:
                 if branch_name == state.update_branch:
                     logger.warning(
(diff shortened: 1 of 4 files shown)


nforro (commented):
Findings from Claude:

### Misleading diff description in adapt_spec.j2

  The template says:
  `Here is the diff of the spec file from that commit against the {{ dist_git_branch }} base:`

  But the actual diff command is:
  `["git", "diff", f"{commit_sha}^", commit_sha, "--", f"{package}.spec"]`

  This shows changes introduced by commit_sha relative to its parent. For the first commit on a branch, the parent IS the dist-git base, so the description is accurate. For subsequent commits on a multi-commit branch, commit_sha^ is the previous commit on the branch, not the dist-git base. Something like "Here is the diff of the spec file introduced by that commit:" would be accurate in all cases.

### `git checkout -- {package}.spec` may silently fail on conflict (pre-existing, now more visible)

  After the cherry-pick:
```python
cp_exit, _, _ = await run_subprocess(["git", "cherry-pick", "--no-commit", commit_sha], ...)
await run_subprocess(["git", "checkout", "--", f"{package}.spec"], ...)
if cp_exit != 0:
    await run_subprocess(["git", "cherry-pick", "--quit"], ...)
    await run_subprocess(["git", "reset", "HEAD"], ...)
```

  If the cherry-pick conflicts on the spec file, git checkout -- {package}.spec fails silently (stages 1/2/3 exist, no stage 0 to restore). The subsequent git reset HEAD cleans the index but leaves conflict markers in the working tree. For the
  has_patches=False path, the LLM then works on a spec with conflict markers while adapt_spec.j2 says "Apply those changes to the current spec file" without mentioning possible conflict state.

  This is pre-existing (the checkout was there before), and the PR's new --quit/reset handling is strictly better than the old code which left the repo in CHERRY_PICKING state. But it might be worth passing a conflict=True flag to the template when cp_exit != 0.

nforro:
Unrelated to these changes, but could we please fix this?
https://github.com/packit/ai-workflows/blob/4ab461502b68ab6c6b460ed2c51ba10fbdd1697d/ymir/agents/mr_consolidation_agent.py#L470

The fallback should be `None`, not `package`.

TomasKorbar:
Updated the prompts to reflect the operations with git we did. Missed that earlier. All should be good now.

nforro (approved):
> Unrelated to these changes, but could we please fix this?
> 
> https://github.com/packit/ai-workflows/blob/4ab461502b68ab6c6b460ed2c51ba10fbdd1697d/ymir/agents/mr_consolidation_agent.py#L470
> 
> The fallback should be `None`, not `package`.

I'll open a new PR for this then.

PR #683, opened 2026-07-14
Add firewall rules for remote sources in RHEL/CS spec files

nforro:


opohorel (dismissed):
LGTM

openshift/tenant-egress.yml
```diff
@@ -125,6 +125,1775 @@ spec:
...
+    # packages: libtopology, powerpc-utils-papr
+    # domains: libtopology.ozlabs.org, powerpc-utils.ozlabs.org
+    - type: Allow
+      to:
+        dnsName: '*.ozlabs.org'
+    # packages: clufter, copy-jdk-configs, filesystem, fonts-rpm-macros, go-rpm-macros, golist, hardlink, hunspell-it, hyphen-it, kde-settings, libosinfo, lohit-tamil-fonts, mailcap, newt, osinfo-db-tools, pungi, python-rpmfluff, rust-packaging, rust-srpm-macros
+    # domains: pagure.io, releases.pagure.io
+    - type: Allow
+      to:
+        dnsName: '*.pagure.io'
```
TomasTomecek: hopefully we'll drop this one soon 😅

openshift/tenant-egress.yml
```diff
@@ -125,6 +125,1775 @@ spec:
...
+    # domains: cloud.github.com, codeload.github.com, download.github.com, github.com, raw.github.com
+    - type: Allow
+      to:
+        dnsName: '*.github.com'
```
TomasTomecek: just a headsup that github.com is already in the list; I wonder if the '*' notation works, I don't think I tried it

TomasTomecek (dismissed):

openshift/tenant-egress.yml
nforro: Ouch, I didn't think of that. It seems it indeed doesn't work, let me regenerate it.

nforro:
@TomasTomecek I've removed wildcards and duplicate entries, PTAL.

openshift/tenant-egress.yml
```diff
@@ -125,6 +125,2373 @@ spec:
...
+      to:
+        dnsName: ayera.dl.sourceforge.net
+    # edac-utils, kxml, libtnc, makebootfat
+    - type: Allow
+      to:
+        dnsName: dl.sourceforge.net
+    # hunspell-ss
+    - type: Allow
+      to:
+        dnsName: downlaods.sourceforge.net
```
opohorel: a typo here

opohorel (dismissed):
from Claude:
```
Defunct hosting services:

Several domains belong to services that have been shut down for years:

download.berlios.de -- BerliOS shut down in 2014
*.googlecode.com (12 entries) -- Google Code shut down in 2016
rubyforge.org -- shut down in 2014
fedorahosted.org -- migrated to Pagure years ago
```

but still LGTM

openshift/tenant-egress.yml
nforro: That's not a typo: https://pkgs.devel.redhat.com/cgit/rpms/hunspell-ss/tree/hunspell-ss.spec?h=rhel-11-main#n11

nforro: Actually, yes, it most likely is :sweat_smile: I thought it's intentional and the domain exists (or existed), but it's dead either way. I'll remove it, along with the other dead domains.

nforro: Done.

TomasTomecek (approved):
lgtm, thank you

PR #684, opened 2026-07-14
Expand find_base_commit tag patterns and soften tag-miss error

opohorel:
`find_base_commit` tried 7 tag patterns (`vX.Y.Z`, `X.Y.Z`, `release-X.Y.Z`, etc.) which missed many upstream conventions. From Jul 02-12 run data, 39 issues (110 error occurrences) hit tag-not-found — the single largest category of backport failure across curl, OpenSSH, PostgreSQL, gnutls, nginx, c-ares, pip, and others.
- Expand tag patterns from 7 to 21: infer package name from repo path, add underscore/uppercase variants (`curl-7_76_1`, `V_9_9_P1`, `REL_16_11`, `gnutls_3_6_2`), and strip trailing digits from RHEL versioned packages (`nodejs22` → `nodejs`)
- Soften tag-miss from `ToolError` to informational `StringToolOutput` — same retry guidance for the LLM, no error trace pollution
- Filter available tags by version components so the LLM sees relevant tags first

### DIFF
--- a/ymir/tools/unprivileged/tests/unit/test_upstream_tools.py
+++ b/ymir/tools/unprivileged/tests/unit/test_upstream_tools.py
@@ -449,6 +449,8 @@ class TestFindBaseCommitTool:
         repo = tmp_path / dir_name
         repo.mkdir()
         subprocess.run(["git", "init"], cwd=repo, check=True)
+        subprocess.run(["git", "config", "user.email", "test@example.com"], cwd=repo, check=True)
+        subprocess.run(["git", "config", "user.name", "Test"], cwd=repo, check=True)
         (repo / "file.c").write_text("int main() {}\n")
         subprocess.run(["git", "add", "."], cwd=repo, check=True)
         subprocess.run(["git", "commit", "-m", "Initial"], cwd=repo, check=True)
@@ -527,7 +529,8 @@ class TestFindBaseCommitTool:
         assert "base_tag_commit" not in tool.options
 
     @pytest.mark.asyncio
-    async def test_finds_curl_style_tag(self, tool, tmp_path):
+    async def test_finds_curl_style_tag(self, tmp_path):
+        tool = FindBaseCommitTool(options={"working_directory": None})
         repo = self._make_repo(tmp_path, "curl-upstream", ["curl-7_76_1"])
         result = await tool.run(
             input=FindBaseCommitToolInput(repo_path=str(repo), version="7.76.1")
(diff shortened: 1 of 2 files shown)


nforro (commented):
Nice, thanks!

Claude found some issues:

**1. Misleading test docstring — `test_finds_versioned_package_tag`**
`test_upstream_tools.py` line ~583

The docstring says:
> RHEL versioned packages like nodejs22 match upstream tags like node-v22.1.0.

But the test actually uses tag `nodejs-20_11_0` (package-underscore style, prefix `nodejs`), not `node-v22.1.0` (which would have prefix `node` and a `v`). The docstring describes a different scenario than what the test exercises. Should align the docstring to the actual test: something like "RHEL versioned packages like nodejs22 strip trailing digits to match upstream tags like nodejs-20_11_0."

**2. Soft failure changes the signal the agent receives**
`upstream_tools.py` line ~501

Changing from `ToolError` to `StringToolOutput` is a behavioral change beyond just suppressing traces. A `ToolError` is an unambiguous failure signal to the BeeAI framework — the agent knows the tool failed and must retry or try a different approach. A `StringToolOutput` is a "success" return that happens to contain failure text. The agent must parse the natural-language message to understand it needs to retry.

This is called out in the PR description as intentional ("same retry guidance for the LLM, no error trace pollution"), and the message text is clear ("Could not find tag... Retry with..."). But worth confirming this was tested in practice with the actual agent — LLMs can treat a non-error tool return as "done, move on" rather than "failed, retry." If there's data from runs showing the agent retries correctly on soft failures, this concern is resolved.

**3. `ver_upper` only handles digit→letter boundaries, not letter→digit**
`upstream_tools.py` line ~429

```python
ver_upper = re.sub(r"(\d)([a-zA-Z])", r"\1_\2", ver_under).upper()
```

This correctly transforms `9.9p1` → `9_9_P1` (openssh style) by splitting at the digit→letter boundary. However, it doesn't handle letter→digit boundaries. For a version like `1.0rc1`, the result is `1_0_RC1` rather than `1_0_RC_1`. If any upstream project uses tags with underscores at both boundaries (e.g., `V_1_0_RC_1`), this wouldn't match. This is a niche edge case and not a regression — just noting the asymmetry in case the data shows `rc`/`beta`/`alpha` suffixed versions failing.

**4. `pkg_base` regex may over-strip on packages ending in digits that aren't RHEL version suffixes**
`upstream_tools.py` line ~415

```python
pkg_base = re.sub(r"[-_]?\d+$", "", pkg_name) if pkg_name else None
```

For a package like `gtk3-upstream`, `pkg_name` is `gtk3`, `pkg_base` is `gtk`. GTK3's upstream tags are like `3.24.42` (just version) or `GNOME_BUILDTOOLS_...`, so adding `gtk-3_24_42` patterns wouldn't help but wouldn't hurt either (just extra git rev-parse calls that fail). No functional issue, just noting the over-eager stripping has low blast radius since the patterns are additive.

**5. Minor: test creates new `tool` instances inconsistently**
Several new tests (`test_finds_openssh_style_tag`, `test_finds_postgresql_style_tag`, etc.) create their own `tool = FindBaseCommitTool(...)` locally instead of using the `tool` fixture. This is because they also create their own repo via `_make_repo` and need `tmp_path` directly. Meanwhile `test_finds_curl_style_tag` does use the `tool` fixture. The inconsistency is harmless (both approaches work) but using the fixture everywhere or nowhere would be cleaner.

opohorel:
@nforro thanks for the review!

1. fixed the docstring
2. the change is intentional, so the trace server/phoenix doesn't flag this as errors. LLM should be able to recover even when we don't raise the ToolError
3. edge case which IMO isn't worth exploring
4. blast radius is small, and the LLM should recover, so I decided to not fix this one
5. fixed

nforro (approved):

PR #685, opened 2026-07-15
Raise pods memory

opohorel:
We OOM on mcp-gateway 3 times today, on rebase once and triage is dangerously close from time to time


mfocko (approved):
Does it fit the available resources though?

opohorel:
> Does it fit the available resources though?

we'll have 1Gi left

PR #686, opened 2026-07-15
Add GetPipelineJobLogTool for reading CI job logs

martinhoyer:
Adds MCP tool that fetches raw log output from a GitLab CI job via the /jobs/{id}/trace REST endpoint.
Complements GetFailedPipelineJobsFromMergeRequestTool - agents call that first to identify which jobs failed, then this tool to read the actual error output.

Not sure about the truncating to the last 500 lines. I'm thinking the ideal would be to fetch the whole logfile and let the agent search in it, wdyt?

TODO:

- [x] Write new tests or update the old ones to cover new functionality.
- [x] Update doc-strings where appropriate.

RELEASE NOTES BEGIN

New GetPipelineJobLogTool GitLab MCP tool.

RELEASE NOTES END


ymir/tools/privileged/gitlab.py
```diff
@@ -866,6 +866,62 @@ def get_latest_pipeline_jobs():
...
+MAX_LOG_LINES = 500
+
+
+class GetPipelineJobLogToolInput(BaseModel):
+    project_path: str = Field(description="GitLab project path (e.g. 'redhat/rhel/rpms/curl')")
+    job_id: str = Field(description="GitLab job ID (from get_failed_pipeline_jobs_from_merge_request output)")
+
+
+class GetPipelineJobLogTool(Tool[GetPipelineJobLogToolInput, ToolRunOptions, StringToolOutput]):
+    name = "get_pipeline_job_log"
```
lbarcziova, line 878: currently, this MCP tool isn't wired to any agent, is this supposed to be used by preliminary agent [here](https://github.com/packit/ai-workflows/blob/6b5c6388ec722cc03e4b6e3ea0128d1a3b06ff28/ymir/agents/preliminary_testing_agent.py#L94)?

ymir/tools/privileged/gitlab.py
```diff
@@ -866,6 +866,62 @@ def get_latest_pipeline_jobs():
...
+    description = (
+        "Fetches the raw log output from a specific GitLab CI job. "
+        "Use after get_failed_pipeline_jobs_from_merge_request to read WHY a job failed. "
+        "Returns the last 500 lines of log output."
```
lbarcziova, line 882: we can start with this, but could also consider the approach used for Copr build logs previously - download the full trace to disk and let the agent view the content iteratively with offset/limit

lbarcziova (commented):
thanks for the contribution!

ymir/tools/privileged/gitlab.py
martinhoyer: Yeah, alongside the `get_failed_pipeline_jobs_from_merge_request`, but it's not necessarily meant to be used by a specific agent.

PR #690, opened 2026-07-17
Fix zombie subprocess accumulation, oversized clones, and missing timeouts

opohorel:
Agent pods accumulate zombie processes and memory over time, leading to OOMKills after ~18 hours. Three root causes addressed:
- **Zombie processes**: Subprocesses not killed on timeout/cancellation — 27 zombies observed on one pod. Now all `proc.communicate()` and `proc.wait()` calls have try/finally cleanup that kills and reaps the child on interruption.
- **Oversized upstream clones**: `clone_upstream_repository` did full `git clone` of massive repos (vim ~2 GB, grafana, openjdk). Switched to `--filter=blob:none` (blobless partial clone) which preserves full commit history/tags but defers blob downloads to on-demand. Also adds 60-min timeout and removes partial directories on failure so retries aren't blocked.
- **No timeout on dist-git clone/fetch**: A hung `git clone` or `git fetch` in `CloneRepositoryTool` blocked the agent slot indefinitely. Added 60-minute `asyncio.wait_for` timeout with proper kill/reap on expiry.

e2e backport tests found no regression

### DIFF
--- a/ymir/common/base_utils.py
+++ b/ymir/common/base_utils.py
@@ -448,7 +448,12 @@ async def run_subprocess(
         if isinstance(cmd, str):
             cmd = shlex.split(cmd)
         proc = await asyncio.create_subprocess_exec(cmd[0], *cmd[1:], **kwargs)
-    stdout, stderr = await proc.communicate()
+    try:
+        stdout, stderr = await proc.communicate()
+    except BaseException:
+        proc.kill()
+        await proc.wait()
+        raise
     return (
         proc.returncode,
         stdout.decode() if stdout else None,
--- a/ymir/tools/privileged/gitlab.py
+++ b/ymir/tools/privileged/gitlab.py
@@ -25,6 +25,7 @@ from ogr.services.gitlab.project import GitlabProject
 from ogr.services.gitlab.pull_request import GitlabPullRequest
 from pydantic import BaseModel, Field
 
+from ymir.common.base_utils import run_subprocess
 from ymir.common.models import (
     CommentReply,
     FailedPipelineJob,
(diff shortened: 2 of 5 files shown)


ymir/tools/unprivileged/upstream_tools.py
```diff
@@ -319,11 +325,19 @@ async def _run(
...
-            # Clone the repository
-            cmd = ["git", "clone", tool_input.repo_url, str(clone_path)]
-            exit_code, _, stderr = await run_subprocess(cmd)
+            # Clone the repository (blobless: full commit graph but deferred blobs)
+            cmd = ["git", "clone", "--filter=blob:none", tool_input.repo_url, str(clone_path)]
+            try:
+                exit_code, _, stderr = await asyncio.wait_for(run_subprocess(cmd), timeout=3600)
+            except TimeoutError:
+                shutil.rmtree(clone_path, ignore_errors=True)
+                raise ToolError("Clone timed out after 60 minutes") from None
```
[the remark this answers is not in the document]
opohorel: Not applicable — the project requires Python >=3.13 (requires-python = ">=3.13" in pyproject.toml). Since Python 3.11, TimeoutError is asyncio.TimeoutError (they're the same class, asyncio.TimeoutError is just an alias). No compatibility concern here.

ymir/tools/unprivileged/tests/unit/test_upstream_tools.py
```diff
@@ -424,6 +424,53 @@ async def mock_run_subprocess(cmd, **kwargs):
...
+
+    @pytest.mark.asyncio
+    async def test_clone_timeout_removes_partial_directory(self, tool, tmp_path, monkeypatch):
+        clone_dir = tmp_path / "mypackage"
+        expected_path = tmp_path / "mypackage-upstream"
+
+        async def mock_wait_for(coro, timeout):
+            expected_path.mkdir(parents=True, exist_ok=True)
+            coro.close()
+            raise TimeoutError
```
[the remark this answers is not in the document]
opohorel: Same as above — on Python 3.13+ these are identical. The mock correctly raises TimeoutError which is what asyncio.wait_for raises.

ymir/common/base_utils.py
```diff
@@ -448,7 +448,12 @@ async def run_subprocess(
         if isinstance(cmd, str):
             cmd = shlex.split(cmd)
         proc = await asyncio.create_subprocess_exec(cmd[0], *cmd[1:], **kwargs)
-    stdout, stderr = await proc.communicate()
+    try:
+        stdout, stderr = await proc.communicate()
+    except BaseException:
+        proc.kill()
+        await proc.wait()
```
mfocko, line 455: What’s the benefit of waiting here? If the process hangs during termination, doesn’t that basically make the caller hang too?

ymir/tools/privileged/gitlab.py
```diff
@@ -502,24 +503,22 @@ async def _run(
         clone_path.mkdir(parents=True, exist_ok=True)
 
         if branch:
-            proc = await asyncio.create_subprocess_exec("git", "init", cwd=clone_path, env=git_env)
-            if await proc.wait():
+            returncode, _, _ = await run_subprocess(["git", "init"], cwd=clone_path, env=git_env)
```
mfocko, line 506: It would probably be a good idea to add customizable timeout to the `run_subprocess` in the future.

mfocko (approved):
lgtm

ymir/common/base_utils.py
opohorel: This came from the investigation. Reasoning is:

proc.kill() sends SIGKILL, which can’t be caught or blocked, so the process is terminated by the kernel in the normal case. await proc.wait() is required to reap it — without that, a killed but unreaped child still becomes a zombie, which is exactly what this PR fixes. After SIGKILL, wait() normally returns immediately; the only realistic hang would be an uninterruptible kernel wait (e.g. stuck NFS), which is outside what userspace can forcibly interrupt anyway.

cgwalters:
The more reliable fix here is deploying an init system as pid 1 like `ENTRYPOINT ["/usr/bin/catatonit"]`.

The other change I think Ymir should make here is not having containers persist by default.

PR #691, opened 2026-07-17
Switch database backend to postgresql

jpodivin:
This is obviously a substantial change. There is no easy way to migrate the trace data we have collected so far, but since we can't properly query it anyway, I don't see that as a much of an issue. That being said, we could export them, if we really want to.

- arize-phoenix is now installed with asyncpg
- Database credentials are set to defaults, it's going to run inside cluster with now outside facing API anyway
- New 150 GB volume for the database
- `phoenix-data` is now 2 GB, that should be enough for the runtime, maybe even an overkill 
- Shrinking volumes can be tricky, so we may have to delete the original and create a new one
- Uses quay.io/sclorg/postgresql-16-c9s container image

From local tests:
image

RELEASE NOTES BEGIN

Replace Phoenix SQLite backend with Postgresql.

RELEASE NOTES END

### DIFF
--- a/Containerfile.phoenix
+++ b/Containerfile.phoenix
@@ -2,7 +2,7 @@
 FROM fedora:44
 RUN dnf install -y python3-pip python3-numpy python3-scipy \
     && dnf clean all \
-    && pip3 install --no-cache-dir 'arize-phoenix<16' 'pydantic-ai-slim<2' \
+    && pip3 install --no-cache-dir 'arize-phoenix[pg]<16' 'pydantic-ai-slim<2' \
     && useradd -u 1001 -g 0 -m -s /sbin/nologin phoenix
 USER 1001
 EXPOSE 6006 4317
--- a/README-agents.md
+++ b/README-agents.md
@@ -166,7 +166,7 @@ Issues can be re-triggered through the workflow in two ways:
 
 ### Phoenix: Alembic migration failure after version rollback
 
-When Phoenix is rolled back to an older version (e.g. v16 → v15), the SQLite database
+When Phoenix is rolled back to an older version (e.g. v16 → v15), the database
 contains an Alembic revision from the newer version that the older code doesn't
 recognize. Phoenix crashes on startup with:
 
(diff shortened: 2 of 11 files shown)


TomasTomecek (commented):
Overall this looks really good, thanks for updating also the documents.

I'm wondering if we need to do some cleanup of the existing data once we deploy it.

jpodivin:
> Overall this looks really good, thanks for updating also the documents.
> 
> I'm wondering if we need to do some cleanup of the existing data once we deploy it.

That's the "best" part. There is no way to migrate the data to the new database. So the cleanup is kind of... given.

openshift/pvc-phoenix-data.yml
```diff
@@ -12,6 +12,6 @@ spec:
   - ReadWriteOnce
   resources:
     requests:
-      storage: 150Gi
```
mfocko, line 15: Did it actually utilize most of the 150Gi?

mfocko (commented):
LGTM

openshift/pvc-phoenix-data.yml
jpodivin: A valid question, I'm not sure about that.

opohorel:
```
oc exec deploy/phoenix -- df -h /mnt/data
Filesystem                                        Size  Used Avail Use% Mounted on
[redacted, so noone from prodsec comes after me]  150G  6.9G  144G   5% /mnt/data
```

opohorel: I would suggest going with 20Gi, so we have some headroom. we'll still save a huge portion of the allocation freed, and we hopefully don't have to worry about it too soon
edit: I meant this comment for db-data not the phoenix-data

jpodivin: I'll revise the PR then.

jpodivin: Done.

majamassarini (approved):
LGTM, thanks!

PR #695, opened 2026-07-20
Clean up all stale top-level git-repos directories

opohorel:
The RHEL-* name filter left arbitrarily named LLM clones (curl-c9s, openssh-upstream, etc.) accumulating on the shared PV. Delete any top-level directory older than the retention window.

Assisted-by: Cursor Grok 4.5


opohorel:
closing in favor of #694

PR #696, opened 2026-07-21
Accept "Done" resolution as shipped in Z-stream clone check

opohorel:
The Y-stream CVE postponement gate only recognized "Done-Errata" as a shipped resolution, causing clones closed with resolution "Done" to be silently dropped — neither counted as shipped nor listed as pending. This left Y-stream issues permanently postponed even after their Z-stream clones had shipped.

Assisted-by: Claude Opus 4.6 (Cursor)


lbarcziova (approved):
LGTM

PR #698, opened 2026-07-21
Fix show-queue targets to display all queued issues

majamassarini:
The jq filter used `.metadata.issue` but downstream agents (rebase, backport, rebuild) populate `.metadata.jira_issue` instead. Tasks from triage used `.issue`, making the filter silently drop every post-triage queue entry.

Check both fields so all items are visible regardless of which agent enqueued them.

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


nforro (approved):
LGTM

opohorel (approved):

PR #699, opened 2026-07-22
Rework trace server web UI

nforro:
- Replaced server-side rendering with a static single page application
- Added real-time trace detail view with live polling for in-progress traces
- Added recent traces and traces per issue views with auto-refresh
- Patched openinference-instrumentation-beeai for near real-time streaming


lbarcziova (approved):
I did a quick pass with help from Claude - this is a bit outside my comfort area, but the code looks solid. No blockers from me.

looking forward to see this live, thanks!

PR #700, opened 2026-07-22
Add Phase 2 to MR cleanup: reset Jira labels for closed bot MRs

lbarcziova:


lbarcziova:
@antbob I extended our auto-closing script to do the cleanup on Jiras as well which your team was interested in (see referenced Jira), I made it possible to skip the auto-closing and configure the Gitlab bot name(s) for which to run the clean-up. Let me know if this works for you!

ymir/mr_cleanup/mr_cleanup.py
```diff
@@ -303,52 +322,239 @@ def process_mr(self, mr: dict, jira_keys: set[str], jira_statuses: dict[str, str
...
+        if missing:
+            logger.warning("Could not fetch labels for Jira issues: %s", ", ".join(sorted(missing)))
+
+        return labels
+
+    JIRA_LABELS_TO_PRESERVE: typing.ClassVar[set[str]] = {
+        JIRA_MR_CLOSED_LABEL,
+        "ymir_todo",
+        "ymir_retry_needed",
+    }
```
opohorel: wouldn't it make sense to add "ymir_merged" here, so we can be sure it never gets removed?

antbob:
@lbarcziova thats great, thanks!

ymir/mr_cleanup/mr_cleanup.py
lbarcziova: makes sense, even though iirc `ymir_merged` is set from the QE agents which are not yet deployed.

But this made me realise a bigger problem: scenario where RHEL-100 has two bot MRs - MR-A merged, MR-B closed. Phase 2 picks up MR-B and wants to reset RHEL-100's labels. That doesn't look correct. I will think about what would be the best way to avoid this (first, but expensive solution would be to parse all the resolved Jiras from the merged MRs).

opohorel (dismissed):
LGTM, thanks!

lbarcziova:
@opohorel PTAL, the last commit changes it a bit, the  `ymir_mr_closed` label is now added to Jiras referenced by closed bot MRs but preserves existing automation labels (e.g. `ymir_backported`); I realised stripping them would break coverage metrics

opohorel (dismissed):
LGTM, thanks for looking into this!

PR #704, opened 2026-07-22
backport: add self-review criterion to catch unrelated changes

majamassarini:
Add Criterion 6 ("Changes match intent") to the build-fix agent's
self-review checklist in `prompt_fix_build_error.j2`. Instead of a
specific rule about dependencies, this is a general review step that
instructs the agent to go over each hunk and verify every change is
directly needed to fix the build error.

Changes that address pre-existing issues unrelated to the patch — such
as loosening or removing BuildRequires/Requires version constraints,
fixing unrelated compiler warnings, or reformatting untouched code — fail
the criterion. If the build failure is caused by something unrelated to
the patch (e.g. a dependency version not yet available in the buildroot),
the agent reports failure and explains the issue instead of working
around it.

Motivated by [python3.12 MR !108](https://gitlab.com/redhat/centos-stream/rpms/python3.12/-/merge_requests/108)
where the agent removed a `sqlite-devel` version constraint to work
around a Copr buildroot gap.

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/ymir/agents/prompts/backport/prompt_fix_build_error.j2
+++ b/ymir/agents/prompts/backport/prompt_fix_build_error.j2
@@ -108,11 +108,23 @@ SPECIAL CONSIDERATIONS FOR TEST FAILURES:
    Verify each patch filename matches the original name from the spec file.
    Do not rename patch files.
 
+   Criterion 6 — Changes match intent:
+   Review each hunk in the regenerated patch file(s). Every changed line must
+   be directly needed to fix the build error or to backport the upstream fix.
+   Changes that address pre-existing issues unrelated to your patch — such as
+   loosening or removing BuildRequires/Requires version constraints, fixing
+   unrelated compiler warnings, or reformatting code you did not touch — do
+   not pass. If the build failure is caused by something unrelated to the
+   patch (e.g. a dependency version not yet available in the buildroot),
+   report success=false and explain the issue instead of working around it.
+
    If ALL criteria pass, report success as normal.
 
    If ANY criterion fails, attempt to fix the problem before reporting failure:
-   - Criterion 1 or 3 (bad patch content or disabled tests): return to step 3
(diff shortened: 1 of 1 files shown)


ymir/agents/prompts/backport/instructions.j2
```diff
@@ -294,6 +294,12 @@ General instructions:
   errors in place — do not abandon to git-am and do not restart from step 1.
 - If necessary, you can run `git checkout -- <FILE>` to revert any changes done to <FILE>.
 - Never change anything in the spec file changelog.
+- Never modify existing BuildRequires, Requires, or other dependency fields in the
```
cgwalters: I know this is a larger architectural discussion, but in my experience things like this are better as a "review checklist" that could be a skill that an agent runs through.

I've also framed it more generally like "Go over each hunk of the commit and ensure that the change is needed and matches the intent" etc.

nforro (commented):
I wonder if we need something similar in the rebase prompt 🤔

opohorel:
> I wonder if we need something similar in the rebase prompt 🤔

IMO in rebase it is more common that you need to adjust BuildRequires

opohorel (changes requested):
I believe that this is not the right place for this prompt change. we are running this prompt when we are backporting. there the agent has no need to adjust BuildRequires. the need comes from the build errors, which are handled by a prompt in `prompt_fix_build_error.j2`. I would suggest adding the changes there instead

opohorel (dismissed):
LGTM

nforro:
> IMO in rebase it is more common that you need to adjust BuildRequires

Yes, but I was wondering if we are able to distinguish between what's part of the rebase and what's a workaround for a failing build - and if we can turn that into a prompt.

opohorel (approved):

PR #717, opened 2026-07-26
Triage CLI

jpodivin:
Triage can now be executed using CLI command `ymir`. The command verifies credentials, spins up containers, and starts the workflow. Once it's done, the containers are stopped. The sole exception is the trace-server, so that developers can access the collected trace information and see how the agent behaved. The dotenv package was introduced to handle env vars, as most of the credentials are managed this way, and they have to be verified and synced across containers and host.

When the workflow is executed, the given JIRA is marked with `ymir_manual_trigger` label. This blocks the the JIRA from being picked up by the fetcher. The intent here is to clearly mark this workflow as somewhat irregular. The CLI is not intended to be a primary way for anyone to use Ymir. This way we can also see how often people invoke Ymir manually, by running as simple query in Jira.

This PR is also a base for possible expansion of CLI to other parts of the workflow.

### Trace server

I didn't want to run more containers than necessary, so to skip `otel-collector`, I've implemented protobuf parsing in the trace-server. This does require two new dependencies, `opentelemetry-proto` and  `protobuf`. The image size is comfortable 265 MB.

But I understand that this is a matter of opinion.

### Changes to ymir/common/pyproject.toml

This code was developed in parallel to [ymir/common/pyproject.toml ](https://github.com/packit/ai-workflows/pull/709) so I had put the `merge_queue.py` on the explicit list to get the package working. Depending on the merge order, this part can be dropped.

### RHEL_CONFIG_PATH env var

To make sure that all workflows load the config pointed to by `--secrets` the `RHEL_CONFIG_PATH` is set by the CLI, and later used by the `load_rhel_config` function.

RELEASE NOTES BEGIN

Triage workflow can now be triggered using CLI.

RELEASE NOTES END

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -87,7 +87,7 @@ x-supervisor: &supervisor
   restart: unless-stopped
 
 services:
-  mcp-gateway:
+  mcp-gateway: &mcp-gateway
     image: mcp-gateway
     build:
       context: .
@@ -108,10 +108,10 @@ services:
       - DRY_RUN=${DRY_RUN:-false}
       - MOCK_JIRA=${MOCK_JIRA:-false}
       - JIRA_DRY_RUN=${JIRA_DRY_RUN:-false}
-      - GIT_REPO_BASEPATH=/git-repos
+      - GIT_REPO_BASEPATH=${GIT_REPO_BASEPATH:-/git-repos}
       - JIRA_MOCK_FILES=ymir/tools/privileged/tests/data
       # e2e tests write insteadOf rewrites here; ignored when the file is absent
-      - GIT_CONFIG_GLOBAL=/git-repos/.mock_gitconfig
+      - GIT_CONFIG_GLOBAL=${GIT_REPO_BASEPATH:-/git-repos}/.mock_gitconfig
     env_file:
       - .secrets/mcp-gateway.env
     volumes:
@@ -119,11 +119,25 @@ services:
       - ${JIRA_MOCK_FILES_HOST:-./ymir/tools/privileged/tests/data}:/home/mcp/ymir/tools/privileged/tests/data:rw,z
       - ./ymir/common:/home/mcp/ymir/common:ro,z
       - .secrets/keytab:/home/mcp/keytab:ro,z,U
-      - git-repos:/git-repos
+      - ${GIT_REPOS_HOST:-git-repos}:${GIT_REPO_BASEPATH:-/git-repos}:z
       - .secrets/rhel-config.json:/home/mcp/rhel-config.json:ro,z,U
     command: ["python", "-m", "ymir.tools.privileged.gateway"]
     restart: unless-stopped
     profiles: ["agents", "supervisor", "e2e-test"]
+
+  mcp-gateway-cli:
+    <<: *mcp-gateway
+    # keep-id maps the host UID into the container so git clones written to
+    # the bind-mounted GIT_REPOS_HOST directory are owned by the calling user.
+    userns_mode: keep-id
+    ports:
+      - "127.0.0.1:8000:8000"
+    restart: "no"
+    profiles: ["cli"]
+    networks:
+      default:
+        aliases:
+          - mcp-gateway
 
   valkey:
     image: docker.io/valkey/valkey:8
@@ -179,7 +193,7 @@ services:
       - ./trace_server:/app:ro,z
       - trace-server-data:/data:U
     restart: unless-stopped
-    profiles: ["agents", "supervisor", "e2e-test"]
+    profiles: ["agents", "supervisor", "e2e-test", "cli"]
 
   otel-collector:
     image: docker.io/otel/opentelemetry-collector-contrib
@@ -212,6 +226,23 @@ services:
       <<: *beeai-env
     command: ["python", "-m", "ymir.agents.triage_agent"]
     profiles: ["agents"]
+
+  triage-cli:
+    <<: *beeai-agent-c10s
+    depends_on: []
+    userns_mode: keep-id
+    environment:
+      <<: *beeai-env
+      COLLECTOR_ENDPOINT: http://trace-server:8080/v1/traces
+      GOOGLE_APPLICATION_CREDENTIALS: /home/beeai/jotnar-vertex-dev.json
+    volumes:
+      - ./ymir:/home/beeai/ymir:ro,z
+      - ${GIT_REPOS_HOST:-git-repos}:${GIT_REPO_BASEPATH:-/git-repos}:rw,z
+      - .secrets/rhel-config.json:/home/beeai/rhel-config.json:ro,z
+      - .secrets/jotnar-vertex-dev.json:/home/beeai/jotnar-vertex-dev.json:ro,z
+    command: ["python", "-m", "ymir.agents.triage_agent"]
+    restart: "no"
+    profiles: ["cli"]
 
   preliminary-testing-agent:
     <<: *beeai-agent-c10s
--- a/jira_label_workflow_routing.md
+++ b/jira_label_workflow_routing.md
@@ -98,6 +98,7 @@ flowchart TD
 | `ymir_todo` | Maintainer-facing trigger for an e2e run | Fetcher swaps it for `ymir_triage_in_progress` on enqueue; only honored when the changelog shows the label was added by a member of the `Red Hat Employee` Jira group (verified per-issue, not via JQL). The triage run posts an ack comment and a result comment so the requester gets feedback. Default is silent — without `ymir_todo`, no comments are posted. |
 | `ymir_consolidate_base` | Mark a backport MR for consolidation (base) | Paired with `ymir_consolidate_next` on another issue for the same package/branch. The fetcher matches the pair, submits a targeted consolidation job, removes both labels, and posts comments. |
 | `ymir_consolidate_next` | Mark a backport MR for consolidation (next) | Must be on a different issue than `ymir_consolidate_base`, for the same package/branch. |
+| `ymir_cli_triage` | Usage-tracking marker for CLI runs | Set at the start of a user-triggered CLI triage run. Prevents the fetcher from picking up the same issue in parallel. Persists on success as a usage-tracking marker; removed on failure so the fetcher can re-process the issue. Covered by the fetcher's stale-label recovery (IN_FLIGHT_LABELS) to handle SIGKILL/OOM. |
 
 ### GitLab MR Labels
 
--- a/ymir/agents/triage_agent.py
+++ b/ymir/agents/triage_agent.py
@@ -996,21 +996,100 @@ async def main() -> None:
     force_cve_triage = os.getenv("FORCE_CVE_TRIAGE", "false").lower() == "true"
 
     if jira_issue := os.getenv("JIRA_ISSUE", None):
+        user_triggered = os.getenv("USER_TRIGGERED", "false").lower() == "true"
         logger.info("Running in direct mode with environment variable")
+        # Direct-mode label lifecycle (user_triggered only):
+        #   Start:   add ymir_cli_triage to prevent concurrent service runs.
+        #   Crash:   remove ymir_cli_triage so the fetcher can re-process.
+        #   Success: keep ymir_cli_triage — it doubles as a usage-tracking
+        #            marker for CLI-triggered triage runs.
+        if user_triggered and not dry_run:
+            try:
+                await tasks.set_jira_labels(
+                    jira_issue=jira_issue,
+                    labels_to_add=[JiraLabels.CLI_TRIAGE.value],
+                    user_triggered=True,
+                    dry_run=dry_run,
+                )
+            except Exception:
+                logger.error(
+                    "Failed to set manual-trigger label on %s — aborting to avoid unprotected run", jira_issue
+                )
+                raise
         with span_processor.start_transaction(jira_issue, workflow="TriageWorkflow"):
             agent_factory = build_agent_factory_with_mock_repos(create_triage_agent, jira_issue)
-            state = await run_workflow(
-                jira_issue,
-                dry_run,
-                agent_factory,
-                auto_chain=auto_chain,
-                force_cve_triage=force_cve_triage,
-            )
-            logger.info(f"Direct run completed: {state.triage_result.model_dump_json(indent=4)}")
+            try:
+                state = await run_workflow(
+                    jira_issue,
+                    dry_run,
+                    agent_factory,
+                    auto_chain=auto_chain,
+                    force_cve_triage=force_cve_triage,
+                    user_triggered=user_triggered,
+                )
+            except Exception:
+                if user_triggered and not dry_run:
+                    try:
+                        await tasks.set_jira_labels(
+                            jira_issue=jira_issue,
+                            labels_to_remove=[JiraLabels.CLI_TRIAGE.value],
+                            user_triggered=True,
+                            dry_run=dry_run,
+                        )
+                    except Exception as label_err:
+                        logger.warning(
+                            "Failed to remove manual-trigger label from %s on error: %s",
+                            jira_issue,
+                            label_err,
+                        )
+                raise
+            output = state.triage_result
+            logger.info(f"Direct run completed: {output.model_dump_json(indent=4)}")
             if state.cve_eligibility_result:
                 logger.info(f"CVE eligibility result: {state.cve_eligibility_result}")
             if state.target_branch:
                 logger.info(f"Target branch: {state.target_branch}")
+
+            result_dir = Path(os.environ.get("GIT_REPO_BASEPATH", "/git-repos")) / jira_issue
+            result_dir.mkdir(parents=True, exist_ok=True)
+            (result_dir / "triage_result.json").write_text(output.model_dump_json(indent=2), encoding="utf-8")
+
+            if user_triggered and not dry_run:
+                resolution_label = _RESOLUTION_TO_LABEL.get(output.resolution)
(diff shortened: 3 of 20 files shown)


compose.yaml
```diff
@@ -108,29 +108,46 @@ services:
...
   valkey:
     image: docker.io/valkey/valkey:8
+    ports:
+      # Exposed for ymir CLI — localhost only, does not affect inter-container traffic
+      - "127.0.0.1:6379:6379"
     volumes:
       - valkey-data:/data
     restart: unless-stopped
-    profiles: ["agents", "supervisor", "e2e-test"]
+    profiles: ["agents", "supervisor", "e2e-test", "cli"]
```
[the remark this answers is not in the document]
jpodivin: This is not really relevant. Why would we even have our own valkey image?

ymir/cli/compose.py
```diff
@@ -0,0 +1,95 @@
...
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            pass
+
+        if standalone := shutil.which(f"{runtime}-compose"):
+            return [standalone]
+
+    raise RuntimeError(
+        "No compose tool found. Install one of: "
+        "podman compose, podman-compose, docker compose, docker-compose"
+    )
```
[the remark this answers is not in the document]
jpodivin: Yes, this is a podman only feature. We are mostly targeting podman anyway. I only included handling for docker to keep parity with Makefile.

ymir/agents/triage_agent.py
```diff
@@ -863,21 +863,89 @@ async def main() -> None:
...
+            output = state.triage_result
+            logger.info(f"Direct run completed: {output.model_dump_json(indent=4)}")
             if state.cve_eligibility_result:
                 logger.info(f"CVE eligibility result: {state.cve_eligibility_result}")
             if state.target_branch:
                 logger.info(f"Target branch: {state.target_branch}")
+
+            result_dir = Path(os.environ.get("GIT_REPO_BASEPATH", "/git-repos")) / jira_issue
+            result_dir.mkdir(parents=True, exist_ok=True)
+            (result_dir / "triage_result.json").write_text(output.model_dump_json(indent=2), encoding="utf-8")
```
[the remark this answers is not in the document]
jpodivin: Technically true, but to exploit this, you would need an access to the host, and have permissions to access the file. And why would you then try to exfiltrate it in this convoluted way.

jira_label_workflow_routing.md
```diff
@@ -95,6 +95,7 @@ flowchart TD
 | `ymir_todo` | Maintainer-facing trigger for an e2e run | Fetcher swaps it for `ymir_triage_in_progress` on enqueue; only honored when the changelog shows the label was added by a member of the `Red Hat Employee` Jira group (verified per-issue, not via JQL). The triage run posts an ack comment and a result comment so the requester gets feedback. Default is silent — without `ymir_todo`, no comments are posted. |
 | `ymir_consolidate_base` | Mark a backport MR for consolidation (base) | Paired with `ymir_consolidate_next` on another issue for the same package/branch. The fetcher matches the pair, submits a targeted consolidation job, removes both labels, and posts comments. |
 | `ymir_consolidate_next` | Mark a backport MR for consolidation (next) | Must be on a different issue than `ymir_consolidate_base`, for the same package/branch. |
+| `ymir_manual_trigger` | Dedup anchor for CLI-triggered runs | Set by the CLI at the start of a user-triggered triage run, before the workflow executes. Prevents the fetcher from picking up the same issue in parallel. Persists after completion alongside the terminal resolution label. |
```
TomasTomecek: I really like this that we'd be able to track how many people will use the CLI

ymir/cli/main.py
```diff
@@ -0,0 +1,180 @@
...
+_INFERENCE_CREDENTIALS: dict[str, tuple[str, str]] = {
+    "gemini:": ("GEMINI_API_KEY", "Google Gemini API key"),
+    "anthropic:": ("ANTHROPIC_API_KEY", "Anthropic API key"),
+    "vertexai:": ("GOOGLE_APPLICATION_CREDENTIALS", "Path to Google Vertex AI service account JSON"),
+}
+
+_ENV_KEYS = ("RHEL_CONFIG_PATH", "MOCK_JIRA", "JIRA_MOCK_FILES_HOST", "GIT_REPOS_HOST")
+
+
+def check_credentials() -> None:
```
TomasTomecek: this function brings a ton of value and should help the adoption a lot

ymir/agents/triage_agent.py
```diff
@@ -996,21 +863,89 @@ async def main() -> None:
     force_cve_triage = os.getenv("FORCE_CVE_TRIAGE", "false").lower() == "true"
 
     if jira_issue := os.getenv("JIRA_ISSUE", None):
+        user_triggered = os.getenv("USER_TRIGGERED", "false").lower() == "true"
```
lbarcziova, line 999: I think with this we might be mixing 2 things, the triage triggered via CLI and the Jira processing triggered by `ymir_todo` lablling in the service, which is designed for posting error comments to Jira and adding intermediate failure labels (_errored suffixes) that don't make sense for CLI (the user sees errors in their terminal). These two concepts should not be conflated (also applies to the code below). 

On a similar note, `ymir_manual_trigger` might be confusing as well because of that, what do you think about something like `ymir_cli` or `ymir_cli_triage` ?

ymir/agents/triage_agent.py
```diff
@@ -996,21 +863,89 @@ async def main() -> None:
...
+            if user_triggered and not dry_run:
+                resolution_label = _RESOLUTION_TO_LABEL.get(output.resolution)
+                if resolution_label and output.resolution != Resolution.ERROR:
+                    stale_labels = [
+                        label
+                        for label in current_labels
+                        if label.startswith("ymir_") and label != JiraLabels.MANUAL_TRIGGER.value
+                    ]
+                    try:
+                        await tasks.set_jira_labels(
```
lbarcziova, line 1061: I think we should try to minimise labelling of Jiras during the CLI triggered runs. The labels are mainly for coordinating what's happening in the service (dedup, routing, status tracking), they don't add much value when triggered locally where the user sees what's happening in their terminal. A lot of service logic relies on these and this could cause some unexpected behaviours (e.g. if the CLI run fails, the label set before the workflow stays and blocks the fetcher from picking up the issue in the service). Maybe we could stick to adding the label to be able to see issues that were invoked from CLI *after* the run finishes successfully?

nforro (dismissed):
LGTM, but Claude has some remarks (nothing major):

#### 1. Compose changes to `mcp-gateway` affect all profiles, not just CLI

The diff parameterizes `GIT_REPO_BASEPATH`, `GIT_CONFIG_GLOBAL`, and the git-repos volume mount in the `*mcp-gateway` anchor:

```yaml
# Before (hardcoded)
- GIT_REPO_BASEPATH=/git-repos
- git-repos:/git-repos

# After (env-var-driven)
- GIT_REPO_BASEPATH=${GIT_REPO_BASEPATH:-/git-repos}
- ${GIT_REPOS_HOST:-git-repos}:${GIT_REPO_BASEPATH:-/git-repos}:z
```

Defaults preserve existing behavior, but these apply to `agents`, `supervisor`, and `e2e-test` profiles too (via the anchor). If someone accidentally exports `GIT_REPOS_HOST` or `GIT_REPO_BASEPATH` in a prod-like environment, the gateway's volume mount silently changes. The `:z` SELinux label addition is actually a consistency fix (other volumes already had it), but it's still a behavioral change for non-CLI profiles.

Consider documenting as intentional scope-widening, or limiting the override to `mcp-gateway-cli` only.

#### 2. `Resolution.ERROR` leaves the issue stuck with only `ymir_manual_trigger`

The label lifecycle:

| Outcome | `ymir_manual_trigger` | Resolution label |
|---|---|---|
| Success | stays | added |
| Crash (exception) | removed | — |
| `Resolution.ERROR` | stays | **skipped** |

On `Resolution.ERROR`, the issue keeps `ymir_manual_trigger` (blocking the fetcher) but gets no resolution label. The issue is effectively claimed with no indication of what happened — diagnosis requires checking container logs. The crash path removes `ymir_manual_trigger` (allowing fetcher retry), suggesting transient failures should be retryable — `Resolution.ERROR` is arguably closer to that category.

Consider writing `TRIAGE_ERRORED` even in direct mode, or documenting this stuck state in the CLI help.

#### 3. `stale_labels` list is a snapshot from before the workflow runs

```python
current_labels, _ = await tasks.get_jira_issue_metadata(jira_issue)
# ...workflow runs (can take minutes)...
stale_labels = [
    label for label in current_labels
    if label.startswith("ymir_") and label != JiraLabels.MANUAL_TRIGGER.value
]
```

If another process adds or removes `ymir_*` labels while the workflow runs, `stale_labels` would miss them or try to remove labels that are already gone. Low risk in practice (CLI issues shouldn't be fetcher-processed concurrently), but worth documenting the assumption.

#### 4. `_normalize_protobuf_ids` doesn't handle link IDs
The function only normalizes span-level `traceId`, `spanId`, `parentSpanId`. Span `links` also carry these fields in base64 after `MessageToDict`. The trace server doesn't currently process link data, so no runtime impact — but if it ever does, base64 IDs containing `/` would break URL routing.

#### 5. f-string in `logger.warning` — style inconsistency

`trace_server/server.py` line 429:
```python
logger.warning(f"POST /v1/traces rejected: invalid protobuf: {e}")
```

Other logging calls in the same file use `%s`-style lazy formatting. The f-string eagerly evaluates `str(e)` even when WARNING is disabled.

#### 6. `functools.cache` on `_compose_base_cmd` never cleared in tests

The cache persists for the entire test process. Each test gets a unique `tmp_path` so there are no cache hits across tests, but within a single test that calls both `start_services` and `stop_services` on the same path (e.g. `TestTriageContainerLifecycle`), the second call hits the cache and skips the mocked `detect_compose_cmd()`. Works by accident since the mock returns the same value — but the cache masks the second call.

#### 7. `load_dotenv` side effects leak between tests

`triage()` calls `load_dotenv(override=False)` which sets env vars like `CHAT_MODEL`. These aren't in `_ENV_KEYS` and aren't restored in the `finally` block. Tests that don't use `patch.dict(os.environ, clear=True)` (e.g. `TestEnvVarRestoration`) leak these values to subsequent tests. Not currently causing failures, but a latent ordering dependency.

compose.yaml
```diff
@@ -108,29 +108,46 @@ services:
...
+    networks:
+      default:
+        aliases:
+          - mcp-gateway
+
   valkey:
     image: docker.io/valkey/valkey:8
+    ports:
+      # Exposed for ymir CLI — localhost only, does not affect inter-container traffic
+      - "127.0.0.1:6379:6379"
```
lbarcziova: is the intent for the CLI to go through queues at some point? Currently direct mode doesn't put anything to queues (that's only in queue mode), so valkey isn't used in CLI runs. If that's the expected behavior, it could be dropped from the cli profile, and `--no-auto-chain` could be removed since it doesn't do anything in direct mode. For follow-up actions users can trigger backport/rebase from CLI as well when that's added.

jpodivin:
> LGTM, but Claude has some remarks (nothing major):
> #### 1. Compose changes to `mcp-gateway` affect all profiles, not just CLI
> 
> The diff parameterizes `GIT_REPO_BASEPATH`, `GIT_CONFIG_GLOBAL`, and the git-repos volume mount in the `*mcp-gateway` anchor:
> 
> ```yaml
> # Before (hardcoded)
> - GIT_REPO_BASEPATH=/git-repos
> - git-repos:/git-repos
> 
> # After (env-var-driven)
> - GIT_REPO_BASEPATH=${GIT_REPO_BASEPATH:-/git-repos}
> - ${GIT_REPOS_HOST:-git-repos}:${GIT_REPO_BASEPATH:-/git-repos}:z
> ```
> 
> Defaults preserve existing behavior, but these apply to `agents`, `supervisor`, and `e2e-test` profiles too (via the anchor). If someone accidentally exports `GIT_REPOS_HOST` or `GIT_REPO_BASEPATH` in a prod-like environment, the gateway's volume mount silently changes. The `:z` SELinux label addition is actually a consistency fix (other volumes already had it), but it's still a behavioral change for non-CLI profiles.
> 
> Consider documenting as intentional scope-widening, or limiting the override to `mcp-gateway-cli` only.
> #### 2. `Resolution.ERROR` leaves the issue stuck with only `ymir_manual_trigger`
> 
> The label lifecycle:
> Outcome 	`ymir_manual_trigger` 	Resolution label
> Success 	stays 	added
> Crash (exception) 	removed 	—
> `Resolution.ERROR` 	stays 	**skipped**
> 
> On `Resolution.ERROR`, the issue keeps `ymir_manual_trigger` (blocking the fetcher) but gets no resolution label. The issue is effectively claimed with no indication of what happened — diagnosis requires checking container logs. The crash path removes `ymir_manual_trigger` (allowing fetcher retry), suggesting transient failures should be retryable — `Resolution.ERROR` is arguably closer to that category.
> 
> Consider writing `TRIAGE_ERRORED` even in direct mode, or documenting this stuck state in the CLI help.
> #### 3. `stale_labels` list is a snapshot from before the workflow runs
> 
> ```python
> current_labels, _ = await tasks.get_jira_issue_metadata(jira_issue)
> # ...workflow runs (can take minutes)...
> stale_labels = [
>     label for label in current_labels
>     if label.startswith("ymir_") and label != JiraLabels.MANUAL_TRIGGER.value
> ]
> ```
> 
> If another process adds or removes `ymir_*` labels while the workflow runs, `stale_labels` would miss them or try to remove labels that are already gone. Low risk in practice (CLI issues shouldn't be fetcher-processed concurrently), but worth documenting the assumption.
> #### 4. `_normalize_protobuf_ids` doesn't handle link IDs
> 
> The function only normalizes span-level `traceId`, `spanId`, `parentSpanId`. Span `links` also carry these fields in base64 after `MessageToDict`. The trace server doesn't currently process link data, so no runtime impact — but if it ever does, base64 IDs containing `/` would break URL routing.
> #### 5. f-string in `logger.warning` — style inconsistency
> 
> `trace_server/server.py` line 429:
> 
> ```python
> logger.warning(f"POST /v1/traces rejected: invalid protobuf: {e}")
> ```
> 
> Other logging calls in the same file use `%s`-style lazy formatting. The f-string eagerly evaluates `str(e)` even when WARNING is disabled.
> #### 6. `functools.cache` on `_compose_base_cmd` never cleared in tests
> 
> The cache persists for the entire test process. Each test gets a unique `tmp_path` so there are no cache hits across tests, but within a single test that calls both `start_services` and `stop_services` on the same path (e.g. `TestTriageContainerLifecycle`), the second call hits the cache and skips the mocked `detect_compose_cmd()`. Works by accident since the mock returns the same value — but the cache masks the second call.
> #### 7. `load_dotenv` side effects leak between tests
> 
> `triage()` calls `load_dotenv(override=False)` which sets env vars like `CHAT_MODEL`. These aren't in `_ENV_KEYS` and aren't restored in the `finally` block. Tests that don't use `patch.dict(os.environ, clear=True)` (e.g. `TestEnvVarRestoration`) leak these values to subsequent tests. Not currently causing failures, but a latent ordering dependency.

Most of these seem valid. And implementing adjustments should be very easy.

compose.yaml
jpodivin: Good point. It's a leftover from an earlier prototype.

ymir/agents/triage_agent.py
jpodivin: I don't really care about the label name, so `ymir_cli_triage` is fine with me.

jpodivin:
Looks like something is wrong with the gateway.

nforro:
> Looks like something is wrong with the gateway.

Should be fixed by https://github.com/packit/ai-workflows/pull/721/changes/f96cd9608f3b2e9980e086dce6d06a3d4bdd33e0.

ymir/agents/triage_agent.py
jpodivin: Blocking the fetcher is actually what I was going for. Running Ymir through CLI is a valid, but a fundamentally irregular operation. I'd rather avoid possible race between fetcher and CLI. If user wants to triage something outside of the regular workflow, we should trust them that they don't want the regular workflow to touch it, whatever reasons they may have.

That being said. I do understand why do you want the label removed on a failed run.

lbarcziova: Thanks for clarification, blocking the fetcher with the label makes sense, agreed. 

My concern was also about error comments to be posted to Jira (because of the user_triggered=True). For CLI, the user sees errors in their terminal, so posting them to Jira too feels unnecessary. This behavior was designed for `ymir_todo` where the user only has Jira as feedback channel.

ymir/cli/main.py
```diff
@@ -0,0 +1,185 @@
...
+@app.command()
+def triage(
+    issue: str = typer.Argument(help="Jira issue key (e.g. RHEL-12345)"),
+    dry_run: bool = typer.Option(False, "--dry-run", help="Skip Jira writes and downstream queues"),
+    force_cve_triage: bool = typer.Option(
+        False, "--force-cve-triage", help="Force triage even when CVE eligibility check says to skip"
+    ),
+    no_auto_chain: bool = typer.Option(
+        False, "--no-auto-chain", help="Do not push results to downstream agent queues"
+    ),
```
lbarcziova: I think this whole functionality can be removed here, as we don't have the valkey queue and propagation to downstream agents/queues

lbarcziova (dismissed):
just one note, otherwise LGTM, thank you

ymir/cli/main.py
jpodivin: Done

lbarcziova (approved):
LGTM

PR #722, opened 2026-07-29
Show queue length in show-*-queue targets

lbarcziova:


majamassarini (approved):
LGTM. I think we can live even without length output.

PR #725, opened 2026-07-30
Add QE agents overview and proposed integration design

lbarcziova:


lbarcziova:
@martinky82 could you please go through the document and provide feedback on the summary and next steps?

martinky82:
Hi,, I went through the document and overall, it a very good description of status quo and where were are heading. 
My comments:

* in the state diagram, there should be the same decision and waiting loop in issue-verification  agent as in preliminay-testing agent. The work in a very similar way, just the source of test results and the resulting action if tests pass are different.
* the ownership checks in the supervisor agents were because of how the Jotnar pilot was designed - the jotnar bot assigned the issues to itself, hence all the agents checked if the ownership is right. That's not needed anymore.

PR #727, opened 2026-08-02
Enable enhancing tool exception context

DaliborKr:
## Summary

- Introduce additional error context for tool failures: a brief and clear error message for the LLM (to not provide unnecessarily bloated or sensitive information), and additional debugging context visible only in logs and observability tools (Phoenix, Trace Server)
- Add `tool_error_context` context manager that automatically captures exception type and message alongside tool-specific context (URLs, parameters, etc.)
- Propagate error context through MCP boundary via `beeai-framework-error-context.patch` (pending [upstream PR](https://github.com/i-am-bee/beeai-framework/commit/718048cd) - by @nforro)
- Filter `additional_context` from LLM error messages in `_runner.py` to prevent context pollution
- Refactor first batch of privileged tools to use `tool_error_context`: jira, errata, distgit, testing_farm, maintainer_rules, zstream_search

This is the **first batch** of tool refactoring — remaining tools will be migrated based on feedback. The context fields provided per tool are not definitive and can be extended or reduced over time as we learn what's most useful for debugging.

### How it works

The first argument is a brief error message visible to the LLM. The keyword arguments are additional context visible only in logs and observability tools. See this dummy example in 'GetJiraDetailsTool':

error_handling_example

Trace server example (@nforro plans to highlight the additional context in the Trace Server later):

trace_server_example

Phoenix example:

phoenix_example


jpodivin (changes requested):
Please rebase on top of current state.

ymir/tools/privileged/tests/unit/test_distgit.py
```diff
@@ -151,7 +151,7 @@ async def mock_get_latest_candidate_build(package, dist_git_branch):
 
     monkeypatch.setenv("GITLAB_TOKEN", "<TOKEN>")
 
-    with pytest.raises(ToolError, match="Push rejected"):
```
lbarcziova, line 154: this change here is intentional, right?

lbarcziova (commented):
the approach looks solid, thanks! 🚀 

One concern from my Claude code session that I'm not sure is intentional:

`Exception scope widens silently when migrating. Several jira.py handlers originally caught only (aiohttp.ClientError, TimeoutError) — network errors. Now tool_error_context catches all Exception. This
  means programming bugs (e.g., KeyError from an unexpected API response shape) get wrapped as tool errors instead of surfacing as unhandled exceptions. When migrating remaining tools, it's worth checking
  whether narrow exception handling was intentional — some tools might want bugs to propagate rather than be silently wrapped as "Failed to do X."`

ymir/tools/privileged/tests/unit/test_distgit.py
DaliborKr: Yes, originally it was intentional. The `ToolError("Push to dist-git was rejected")` is outside the inner `tool_error_context("Failed to push branch to dist-git", ...)`, so it was caught by the outer one with `"Failed to create Z-Stream branch"` message as a part of generalizing messages for LLM (see [changes in `distgit.py` file](https://github.com/DaliborKr/ai-workflows/blob/bb384206e5243632f75fe9cb41a72545502af2d0/ymir/tools/privileged/distgit.py#L202-L252). 

But you're right, as I am looking into this, it should just be moved inside the inner `tool_error_context` block, so the behavior matches the original implementation and provides the specific information `"Failed to push branch to dist-git"`. I will fix this, thanks!

beeai-framework-error-context.patch
```diff
@@ -0,0 +1,50 @@
...
+-        output: ToolOutput = await cloned_tool.run(kwargs)
++        try:
++            output: ToolOutput = await cloned_tool.run(kwargs)
++        except Exception as e:
++            error_context = getattr(e, "context", None) or {}
++            meta = {"error_context": error_context} if error_context else None
++            return MCPCallToolResult(
++                content=[MCPTextContent(type="text", text=f"Error executing tool {tool.name}: {e}")],
++                isError=True,
++                _meta=meta,
```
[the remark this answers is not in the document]
nforro: I'd say it's the caller's responsibility not to raise exception instances that render sensitive data when stringified.

ymir/agents/reasoning_agent/_runner.py
```diff
@@ -477,9 +477,20 @@ async def _invoke_tool_calls(
...
+                    filtered_context = {
+                        k: v for k, v in original_context.items() if k != "additional_context"
+                    }
+                    tool_call.error.context = filtered_context
+                    try:
+                        reason = tool_call.error.explain()
+                    finally:
+                        tool_call.error.context = original_context
+                else:
+                    reason = tool_call.error.explain()
```
nforro, line 491:
```suggestion
                try:
                    if original_context and "additional_context" in original_context:
                        tool_call.error.context = {k: v for k, v in original_context.items() if k != "additional_context"}
                    reason = tool_call.error.explain()
                finally:
                    tool_call.error.context = original_context
```

ymir/tools/errors.py
```diff
@@ -0,0 +1,28 @@
...
+        context: dict[str, Any] | None = None,
+        additional_context: dict[str, Any] | None = None,
+    ) -> None:
+        # Merge additional_context into context under special key for observability
+        merged_context = context.copy() if context else {}
+        if additional_context:
+            merged_context["additional_context"] = additional_context
+
+        super().__init__(message, cause=cause, context=merged_context)
+        self.additional_context = additional_context or {}
```
nforro, line 28: Is the attribute necessary? If yes, wouldn't it make more sense to make it a property that extracts the `additional_context` key from `context`?

ymir/tools/gateway_utils.py
```diff
@@ -62,6 +62,9 @@ def on_tool_error(data: Any, meta: Any):
         error = getattr(data, "error", None)
         if error is not None:
             logger.error(f"Tool {meta.creator} traceback:", exc_info=error)
+            additional_context = getattr(error, "additional_context", None)
+            if additional_context:
```
nforro, line 66:
```suggestion
            if additional_context := getattr(error, "additional_context", None):
```

ymir/agents/observability.py
```diff
@@ -84,6 +90,28 @@ def force_flush(self, timeout_millis: int = 30000) -> bool:
...
+        if error is None:
+            return
+        # Unprivileged tools: error is ToolErrorWithContext with .additional_context
+        additional_context = getattr(error, "additional_context", None)
+        if not additional_context:
+            # Privileged tools: error crossed MCP, context is in ToolError.context dict
+            error_context = getattr(error, "context", None)
+            if isinstance(error_context, dict):
+                additional_context = error_context.get("additional_context")
+        if additional_context:
```
nforro, line 105: See the note about property above. Then you can do just:
```suggestion
        if additional_context := getattr(error, "additional_context", None):
```

DaliborKr:
@lbarcziova 

> the approach looks solid, thanks! 🚀
> 
> One concern from my Claude code session that I'm not sure is intentional:
> 
> `Exception scope widens silently when migrating. Several jira.py handlers originally caught only (aiohttp.ClientError, TimeoutError) — network errors. Now tool_error_context catches all Exception. This means programming bugs (e.g., KeyError from an unexpected API response shape) get wrapped as tool errors instead of surfacing as unhandled exceptions. When migrating remaining tools, it's worth checking whether narrow exception handling was intentional — some tools might want bugs to propagate rather than be silently wrapped as "Failed to do X."`

The widening should be harmless here. The original narrow catches all did the same thing. They did `raise ToolError(f"Failed to do X: {e}")` with no different behavior per exception type. And the beeai framework already catches all `Exception` at the tool level anyway, so unexpected errors never propagated unhandled (they just got a worse error message without any context). 

In this approach what matters is the `error_message` we pass to `tool_error_context`, not which exceptions it catches.

ymir/agents/observability.py
DaliborKr: Yes this approach is much cleaner, but if I understand it correctly this would work only for unprivileged tools where the error is a `ToolErrorWithContext` instance. For privileged tools, the error crosses the MCP boundary and gets reconstructed as a plain `ToolError`, so it doesn't have `.additional_context` as a property, and the data has to be retrieved as `ToolError.context["additional_context"]`, which is why I thought the "fallback" path is needed. Correct me if I am wrong.

ymir/agents/reasoning_agent/_runner.py
DaliborKr: This simplification makes sense, thanks.

ymir/tools/errors.py
DaliborKr: Yes, `additional_context` is used in `ymir/agents/observability.py` (as you pointed out there). Anyway, you are right, I replaced the attribute with a property.

ymir/agents/observability.py
nforro: Ok, then there are two options:
- drop the attribute/property and otherwise leave this as it is - there is no reason to do things differently for privileged and unprivileged tools
- instantiate `ToolErrorWithContext` from `ToolError` - the only difference between those classes would be the property (and a constructor argument)

DaliborKr: I went with option 1. I dropped the property and both `observability.py` and `gateway_utils.py` now read the additional context from `error.context.get("additional_context")`.

ymir/agents/observability.py
```diff
@@ -84,6 +90,28 @@ def force_flush(self, timeout_millis: int = 30000) -> bool:
...
+            logger.error(f"Tool {meta.creator} additional context: {additional_context}")
+            span = trace_api.get_current_span()
+            if span.is_recording():
+                for key, value in additional_context.items():
+                    span.set_attribute(f"metadata.{key}", str(value))
```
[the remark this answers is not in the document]
DaliborKr: The `additional_context` is always set as a `dict` by our code in `ToolErrorWithContext.__init__`, and the `getattr` default handles the missing-attribute case. No change needed here.

jpodivin (commented):
I don't see any glaring issues. But the merge conflict has to be resolved first.

DaliborKr:
@jpodivin

I am continuously updating the branch to keep it up to date and pushing it along with the review changes. In any case, the plan was to implement this approach for all tools (privileged and unprivileged) in subsequent PRs and then merge them all together later.

ymir/tools/privileged/jira.py
```diff
@@ -197,7 +198,7 @@ async def _run(
         logger.info(f"Connecting to JIRA API to get issue details: {jira_url}")
 
         async with aiohttpClientSession(timeout=AIOHTTP_TIMEOUT) as session:
-            try:
+            with tool_error_context(f"Failed to get details about issue {issue_key}", url=jira_url):
                 async with aiohttp_get_with_retries(
                     session,
                     jira_url,
```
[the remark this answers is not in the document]
DaliborKr: Added unit tests for `tool_error_context` to `ymir/tools/privileged/tests/unit/test_base.py` covering exception wrapping, inner `ToolErrorWithContext` passthrough, and credential redaction. The jira tools use the same pattern so this covers them too.

ymir/tools/privileged/jira.py
```diff
@@ -197,7 +198,7 @@ async def _run(
         logger.info(f"Connecting to JIRA API to get issue details: {jira_url}")
 
         async with aiohttpClientSession(timeout=AIOHTTP_TIMEOUT) as session:
-            try:
+            with tool_error_context(f"Failed to get details about issue {issue_key}", url=jira_url):
```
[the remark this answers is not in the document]
DaliborKr: Error handling is not covered in `jira_data_flow.md`. It documents 
data flow and architecture only. This change does not affect either.

ymir/agents/observability.py
```diff
@@ -84,6 +90,23 @@ def force_flush(self, timeout_millis: int = 30000) -> bool:
...
 
 
+def _setup_tool_error_logging() -> None:
+    def on_tool_error(data: Any, meta: Any) -> None:
+        error = getattr(data, "error", None)
+        if error is None:
+            return
+        additional_context = getattr(error, "context", {}).get("additional_context")
+        if not additional_context:
+            return
```
nforro:
```suggestion
        if (error := getattr(data, "error", None)) is None:
            return
        if not (additional_context := getattr(error, "context", {}).get("additional_context")):
            return
```

nforro (dismissed):

ymir/agents/observability.py
DaliborKr: Applied!

ymir/tools/base.py
```diff
@@ -1,11 +1,32 @@
...
+
+
+@contextmanager
+def tool_error_context(error_message: str, **additional_context):
+    try:
+        yield
+    except ToolErrorWithContext:
+        raise
+    except Exception as e:
+        additional_context["exception"] = f"{type(e).__name__}: {e}"
```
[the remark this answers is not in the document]
nforro: This is a valid observation, but it actually uncovers a discrepancy between privileged and unprivileged tools - a privileged tool raises `AbortError` on timeout, but it is captured by MCP and surfaces as a regular `ToolError`, while an unprivileged tool raises `AbortError` directly in the agent context and aborts the workflow. We should decide which behavior is the correct one and make it consistent across all tools.

jpodivin (approved):

lbarcziova (approved):
nicely done, thank you!

TomasTomecek (approved):

ymir/tools/base.py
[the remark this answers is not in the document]
nforro: I've added an arch topic. As this PR touches only privileged tools, this is not really a concern.

nforro (approved):

PR #737, opened 2026-08-05
Add bare bones LangFuse integration

antbob:
This adds export to LangFuse Cloud OTEL endpoint and Session ID markers to aggregate Sessions by JIRA.

I'm keeping this as DRAFT for now as enabling it by default and without on-boarded LangFuse account  and egress rule/s would likely spam the logs with trace export failures so feel free to slice and dice it as needed.

### DIFF
--- a/openshift/configmap-otel-collector-config.yml
+++ b/openshift/configmap-otel-collector-config.yml
@@ -17,6 +17,11 @@ data:
       otlphttp/trace-server:
         endpoint: http://localhost:8080
         encoding: json
+      otlphttp/langfuse:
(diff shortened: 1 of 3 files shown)


TomasTomecek (commented):
looks solid, thank you!

we'll definitely need to figure out internal docs how to set up credentials correctly

antbob:
@TomasTomecek you will need to create API Keys in LangFuse and it will give you LANGFUSE_SECRET_KEY and LANGFUSE_PUBLIC_KEY which you then set as "pk:sk" form secret in OCP and thats it.

PR #740, opened 2026-08-05
The resolution is not an optional argument

jpodivin:
This bothered me a bit. The `resolution` argument is supplied in all call locations. It also doesn't make sense to keep it with `None` as a default.


jpodivin:
Duplicate of https://github.com/packit/ai-workflows/pull/719

PR #741, opened 2026-08-06
Trace server improvements

nforro:
- **Use enum values of supported OTEL types directly**
  This fixes span rendering in trace-server.
- **Make spans linkable**
  With these changes clicking on a span updates the URL and when accessing an URL with a span ID the corresponding span is scrolled to and focused. Sidebar navigation has been updated to use span IDs as well,
for consistency.


lbarcziova (approved):
LGTM, don't see any issues, thanks!

PR #744, opened 2026-08-07
Use the same z-stream branch creation logic as rhpkg implementation

antbob:
The rhpkg is the reference implementation for z-stream branch creation logic. As we're still seeing a great deal of z-stream branch creation failures i've decided doing exactly what rhpkg does is the best way to address those issues. The main changes to the current logic are:

- mirrors rhpkg --allow-pending behavior
- tries -z-pending first then falls back to -z
- added commit verification
- added post-push ls-remote check
- restricted NVR walk to main branches only (note that rhpkg does not do the NVR walk)

Also some changes related to logging so when these kinda errors occur the MCP logs actually have some useful data/output to troubleshoot.

### DIFF
--- a/ymir/common/utils.py
+++ b/ymir/common/utils.py
@@ -4,6 +4,7 @@ Common utility functions shared across the BeeAI system.
 
 import asyncio
 import logging
+import os
 from collections.abc import AsyncGenerator, Callable
 from contextlib import asynccontextmanager
 from datetime import timedelta
@@ -401,3 +402,39 @@ def extract_text_from_adf(adf_body) -> str:
     if isinstance(adf_body, list):
         return " ".join(extract_text_from_adf(item) for item in adf_body)
     return ""
+
+
+def init_sentry() -> None:
+    """Initialize Sentry, if the DSN is set."""
+    if not (dsn := os.getenv("SENTRY_DSN")):
+        # no DSN, no reporting
+        return
+
+    import sentry_sdk
+    from sentry_sdk.integrations.asyncio import AsyncioIntegration
+    from sentry_sdk.integrations.litellm import LiteLLMIntegration
+    from sentry_sdk.integrations.logging import (
+        ignore_logger,
+        ignore_logger_for_sentry_logs,
+    )
+
+    sentry_sdk.init(
+        dsn=dsn,
+        environment=os.getenv("SENTRY_ENVIRONMENT"),
+        enable_logs=True,
+        # Set traces_sample_rate to 1.0 to capture 100%
+        # of transactions for tracing.
+        traces_sample_rate=1.0,
+        # Add data like inputs and responses;
+        # see https://docs.sentry.io/platforms/python/data-management/data-collected/ for more info
+        stream_gen_ai_spans=True,
+        send_default_pii=True,
+        integrations=[
+            AsyncioIntegration(),
+            LiteLLMIntegration(),
+        ],
+    )
+
+    for ignored_logger in ("agent.redis", "agent.task_loop", "agent.trajectory"):
+        ignore_logger(ignored_logger)
+        ignore_logger_for_sentry_logs(ignored_logger)
--- a/ymir/common/utils.py
+++ b/ymir/common/utils.py
@@ -240,31 +240,6 @@ async def get_latest_z_pending_build(package: str, dist_git_branch: str) -> tupl
     )
 
 
-async def get_latest_z_build(package: str, dist_git_branch: str) -> tuple[EVR, str]:
-    """Look up the latest build for z-stream branch creation.
-
-    Tries ``{branch}-z-pending`` first (builds approved for release but not
-    yet shipped — equivalent to rhpkg's ``--allow-pending``), then falls back
-    to ``{branch}-z`` (released builds).  Both queries use Koji
-    ``inherit=True`` so they also see parent tags.
-    """
-    pending_tag = f"{dist_git_branch}-z-pending"
-    released_tag = f"{dist_git_branch}-z"
-    build = await asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, pending_tag, package)
-    tag_used = pending_tag
-    if build is None:
-        build = await asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, released_tag, package)
-        tag_used = released_tag
-    if build is None:
-        raise RuntimeError(f"There are no builds of {package} in {pending_tag} or {released_tag}")
-    evr = _evr_from_build(build)
-    session = koji.ClientSession(BREWHUB_URL)
-    metadata = await asyncio.to_thread(session.getBuild, build["build_id"], strict=True)
-    source_ref = metadata["source"].split("#")[-1]
-    logger.info(f"Found {package} build {build['nvr']} in {tag_used} (commit {source_ref[:12]})")
-    return evr, source_ref
-
-
 def _resolve_buildroot_checks(
     target_branch: str, fix_version: str, rhel_config: dict | None = None
 ) -> list[tuple[str, str]]:
--- a/ymir/common/utils.py
+++ b/ymir/common/utils.py
@@ -240,6 +240,31 @@ async def get_latest_z_pending_build(package: str, dist_git_branch: str) -> tupl
     )
 
 
+async def get_latest_z_build(package: str, dist_git_branch: str) -> tuple[EVR, str]:
+    """Look up the latest build for z-stream branch creation.
+
+    Tries ``{branch}-z-pending`` first (builds approved for release but not
+    yet shipped — equivalent to rhpkg's ``--allow-pending``), then falls back
+    to ``{branch}-z`` (released builds).  Both queries use Koji
+    ``inherit=True`` so they also see parent tags.
+    """
+    pending_tag = f"{dist_git_branch}-z-pending"
+    released_tag = f"{dist_git_branch}-z"
+    build = await asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, pending_tag, package)
+    tag_used = pending_tag
+    if build is None:
+        build = await asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, released_tag, package)
+        tag_used = released_tag
+    if build is None:
+        raise RuntimeError(f"There are no builds of {package} in {pending_tag} or {released_tag}")
+    evr = _evr_from_build(build)
+    session = koji.ClientSession(BREWHUB_URL)
+    metadata = await asyncio.to_thread(session.getBuild, build["build_id"], strict=True)
+    source_ref = metadata["source"].split("#")[-1]
+    logger.info(f"Found {package} build {build['nvr']} in {tag_used} (commit {source_ref[:12]})")
+    return evr, source_ref
+
+
 def _resolve_buildroot_checks(
     target_branch: str, fix_version: str, rhel_config: dict | None = None
 ) -> list[tuple[str, str]]:
--- a/ymir/tools/privileged/distgit.py
+++ b/ymir/tools/privileged/distgit.py
@@ -16,8 +16,8 @@ from pydantic import BaseModel, Field
 from specfile import Specfile
 
 from ymir.common.base_utils import KerberosError, init_kerberos_ticket
-from ymir.common.utils import get_latest_z_build
-from ymir.common.version_utils import parse_zstream_branch_name
+from ymir.common.utils import get_latest_candidate_build, get_latest_z_pending_build
+from ymir.common.version_utils import is_older_zstream, parse_zstream_branch_name
 from ymir.tools.base import CloneableTool as Tool
 from ymir.tools.privileged.utils import sanitize_url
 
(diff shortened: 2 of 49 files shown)


ymir/common/utils.py
```diff
@@ -240,6 +240,31 @@ async def get_latest_z_pending_build(package: str, dist_git_branch: str) -> tupl
...
+        build = await asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, released_tag, package)
+        tag_used = released_tag
+    if build is None:
+        raise RuntimeError(f"There are no builds of {package} in {pending_tag} or {released_tag}")
+    evr = _evr_from_build(build)
+    session = koji.ClientSession(BREWHUB_URL)
+    metadata = await asyncio.to_thread(session.getBuild, build["build_id"], strict=True)
+    source_ref = metadata["source"].split("#")[-1]
+    logger.info(f"Found {package} build {build['nvr']} in {tag_used} (commit {source_ref[:12]})")
+    return evr, source_ref
```
sgallagher: You are doing double the work here. It's safe to just check `{dist_git_branch}-z-pending` because it inherits from `{dist_git_branch}-z`. Thus, if nothing is specifically pending, it will already fall back to the released Z-stream.

ymir/common/utils.py
```diff
@@ -240,6 +240,31 @@ async def get_latest_z_pending_build(package: str, dist_git_branch: str) -> tupl
...
+    build = await asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, pending_tag, package)
+    tag_used = pending_tag
+    if build is None:
+        build = await asyncio.to_thread(_get_latest_koji_build, BREWHUB_URL, released_tag, package)
+        tag_used = released_tag
+    if build is None:
+        raise RuntimeError(f"There are no builds of {package} in {pending_tag} or {released_tag}")
+    evr = _evr_from_build(build)
+    session = koji.ClientSession(BREWHUB_URL)
+    metadata = await asyncio.to_thread(session.getBuild, build["build_id"], strict=True)
```
sgallagher: Something I've done in ELNBuildSync that might be handy here:

Replace the pattern of "create session and then call in a thread" with a function `call_koji()` that is decorated with:

```
@tenacity.retry(
    wait=wait_exponential(),
    stop=stop_after_delay(60),
    reraise=True,
)
def call_koji (method, *args, **kwargs):
```

It gives you a single place to update if you end up needing to add authentication (such as `session.gssapi_login()`) as well as using Tenacity to automatically retry. (In EBS, we have [additional logic](https://github.com/fedora-eln/elnbuildsync/blob/main/elnbuildsync/kojihelpers/connection.py#L367) to retry only on transient errors).

ymir/tools/privileged/distgit.py
```diff
@@ -222,24 +221,30 @@ async def _clone():
                         "skipping push and waiting for mirror sync"
                     )
                 else:
-                    if await is_older_zstream(branch):
-                        _, ref = await get_latest_z_pending_build(package, branch)
-                    else:
-                        _, ref = await get_latest_candidate_build(package, branch)
```
sgallagher, line 228: You replace these with `get_latest_z_build()`, but you didn't remove the actual implementations. Are they used elsewhere? And if so, are they incorrect there too?

ymir/common/utils.py
```diff
@@ -240,6 +240,31 @@ async def get_latest_z_pending_build(package: str, dist_git_branch: str) -> tupl
     )
 
 
+async def get_latest_z_build(package: str, dist_git_branch: str) -> tuple[EVR, str]:
+    """Look up the latest build for z-stream branch creation.
+
+    Tries ``{branch}-z-pending`` first (builds approved for release but not
+    yet shipped — equivalent to rhpkg's ``--allow-pending``), then falls back
+    to ``{branch}-z`` (released builds).  Both queries use Koji
+    ``inherit=True`` so they also see parent tags.
```
sgallagher: This doesn't actually return a "build", but a tuple of (EVR, commit_ID). Maybe rename it to `get_latest_z_build_commit()`?

ymir/tools/privileged/distgit.py
```diff
@@ -222,24 +221,30 @@ async def _clone():
...
-                    if source_branch := self._find_source_branch(repo, branch):
+                    _, ref = await get_latest_z_build(package, branch)
+                    source_branch = self._find_source_branch(repo, branch)
+                    if source_branch and source_branch.endswith("-main"):
                         ref = await self._find_latest_same_nvr_ref(
                             repo,
                             package,
                             ref,
                             source_branch,
                         )
```
sgallagher, line 235: I'm confused why you do this. NVRs are guaranteed unique in Koji, so the `_find_latest_same_nvr_ref()` is always going to return the same value that you get from `get_latest_z_build()`, except that it will take longer and require more calls into Koji.

ymir/tools/privileged/distgit.py
```diff
@@ -222,24 +221,30 @@ async def _clone():
...
+                        ) from None
+                    await _retry_transient(
+                        lambda: asyncio.to_thread(repo.git.push, "origin", f"{ref}:refs/heads/{branch}"),
                         f"push {branch} to dist-git",
                     )
-                    for info in push_infos:
-                        if info.flags & git.remote.PushInfo.ERROR:
-                            raise RuntimeError(f"Push rejected: {info.summary.strip()}")
+                    if not await asyncio.to_thread(repo.git.ls_remote, "--heads", "origin", branch):
+                        raise RuntimeError(
```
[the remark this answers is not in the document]
sgallagher: The AI review here is correct, but it's also possible (depending on the server implementation) that there could also be a race condition here where the `ls_remote` ends up talking to a different, load-balanced git server that hasn't yet picked up the replication. A retry with backoff would make sense here.

ymir/tools/privileged/distgit.py
```diff
@@ -222,24 +221,30 @@ async def _clone():
...
                     )
                 else:
-                    if await is_older_zstream(branch):
-                        _, ref = await get_latest_z_pending_build(package, branch)
-                    else:
-                        _, ref = await get_latest_candidate_build(package, branch)
-                    if source_branch := self._find_source_branch(repo, branch):
+                    _, ref = await get_latest_z_build(package, branch)
+                    source_branch = self._find_source_branch(repo, branch)
+                    if source_branch and source_branch.endswith("-main"):
```
sgallagher, line 229: I don't understand why you care about the branch name _at all_. From `get_latest_z_build()`, you have the exact git ref ID that you need to create the new branch. Wherever it came from is irrelevant.

sgallagher (changes requested):

ymir/tools/privileged/distgit.py
```diff
@@ -222,24 +221,30 @@ async def _clone():
                         "skipping push and waiting for mirror sync"
                     )
                 else:
-                    if await is_older_zstream(branch):
-                        _, ref = await get_latest_z_pending_build(package, branch)
-                    else:
-                        _, ref = await get_latest_candidate_build(package, branch)
-                    if source_branch := self._find_source_branch(repo, branch):
+                    _, ref = await get_latest_z_build(package, branch)
```
nforro: This would create a regression for leading Z-Streams. We don't want to ignore builds that passed gating (i.e. are in a candidate tag) and create a Z-Stream branch from an older commit.

ymir/tools/privileged/distgit.py
nforro: No, it will return a more recent commit that didn't change NVR, if any. The purpose is to include fixups that don't require a rebuild, e.g. test changes required for the build to pass gating.

ymir/tools/privileged/distgit.py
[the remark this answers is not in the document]
antbob: lets fix this

ymir/common/utils.py
antbob: this as well

ymir/common/utils.py
antbob: perhaps a good suggestion but out of scope for this patch, i dont wanna introduce anything else here besides

ymir/tools/privileged/distgit.py
antbob: they are used elsewhere and it would introduce inconsistency (see nforro comments as well) so i think what needs to be done here is to relax rhpkg pending logic for candidate tags for leading z-streams and strictly follow rhpkg (pending) logic for older z-streams

ymir/common/utils.py
antbob: existing sibling functions follow the same pattern

ymir/tools/privileged/distgit.py
antbob: good point about ls_remote. wrt branch we do care because of NVR walk (which doesnt exist in rhpkg logic) as we dont wanna do it for older z-streams

ymir/tools/privileged/distgit.py
antbob: let me address this by redoing it so its only applies to older z-streams and keep your existing logic as is

antbob:
this reverts previous changes and instead applies only to older z-streams.  it now behaves identically to rhpkg --create-z-stream-branch --allow-pending for older z-streams only, for leading z-streams the behavior stays as it was before so this should address most (all?) the concerns related to the original patch.

ymir/common/utils.py
sgallagher: Agreed, but looking at the code in general, you have a LOT of retry logic all over the place. I highly recommend consolidating around a simple decorator so you don't have to fix it in dozens of places in the future. For a future PR, certainly.

ymir/common/utils.py
sgallagher: For the sake of future maintainers, please use names that do not mislead.

ymir/tools/privileged/distgit.py
sgallagher: Now that I understand the purpose of that NVR walk, we actually probably want to incorporate it into `rhpkg` as well. I can see it being frustrating if we create a branch that lacks test fixes. I'm not sure how that will play once we use RHEL-on-GitLab for everything, but at least right now we need to accommodate that. I'll open a ticket with `rhpkg`.

ymir/tools/privileged/distgit.py
sgallagher: > This would create a regression for leading Z-Streams. We don't want to ignore builds that passed gating (i.e. are in a candidate tag) and create a Z-Stream branch from an older commit.

Given the automation we have today, doesn't a lack of the `-pending` mean that the build is NOT part of an erratum? In that case, can't we make the reasonable assumption that something (human or machine) determined that it was not suitable for delivery to customers? Why would we want to branch from there?

ymir/tools/privileged/distgit.py
sgallagher: I was thinking about this some more this morning, and there's still a problem here. The NVR walk can still break things in the situation where a maintainer adds a patch _only intended for the Y-stream_ but forgets to bump the release field, you will pull in that Y-stream-only patch incorrectly. Because of this, I think we'll probably not make this change in `rhpkg`. It's much easier for the maintainer to pull in additional commits after the branch than it is to rewrite the branch to remove unintentionally-included ones (or force the maintainer to put a revert commit into the history).

Also, it looks like you determine this by doing a full parse of the Specfile. This has known security issues, since the parser is Turing-complete. Please consider switching to https://github.com/praiskup/norpm which is specifically designed to avoid side effects from the parser.

antbob:
added z-stream branch creation info propagation to MR notes so the branching point is clear to reviewer/s.

ymir/tools/privileged/distgit.py
```diff
@@ -230,31 +241,47 @@ async def _clone():
...
                         _, ref = await get_latest_candidate_build(package, branch)
-                if source_branch := self._find_source_branch(repo, branch):
+                source_branch = self._find_source_branch(repo, branch)
+                if source_branch and source_branch.endswith("-main"):
                     ref = await self._find_latest_same_nvr_ref(
                         repo,
                         package,
                         ref,
                         source_branch,
                     )
```
nforro: It was established that `-z-pending` is correct also for leading Z-Streams, so we can get rid of the condition and `get_latest_candidate_build()`. And it would probably make sense to ditch the NVR walk (`_find_latest_same_nvr_ref()`) as well, @TomasKorbar WDYT?

nforro: @antbob It seems Tomáš is on PTO, so let's leave the NVR walk in for now, I'll open a follow-up PR if needed.

antbob: @nforro thanks for not leaving me hanging! the get_latest_z_pending_build is now used for all z-streams and i left the nvr walk for you guys to figure out later. please have a look and approve if everything is in order.

nforro (approved):
Just one remark - the branch creation note won't make it into a consolidated MR, right? Would it make sense to implement that?

antbob:
> Just one remark - the branch creation note won't make it into a consolidated MR, right? Would it make sense to implement that?

it would but it will be less obvious / user friendly than with a standalone MR. whichever standalone MR has created the branch would include a Note in its description which in turn is included in consolidated MR so the info will be there just less visible. not ideal but i just wanted to keep it simple for now.

nforro (commented):
Actually, could you please get rid of those merge commits?

antbob:
sure, i will use GH "squash and merge" for this.

PR #745, opened 2026-08-10
Update log instructions to exclude unrelated Jira references

antbob:
When backporting to older z-streams the current log agent instructions allow bleeding unrelated Jira issue references (related to the original Jira issue/s only) in the Changelog and Commit messages. This can become problematic in some cases as there are Jira issue checks in the standard CI pipeline now so what we wanna do here is to instruct the agent to remove any unrelated Jira references and only keep those that are in the actual backport scope. There is no good way to address this via code changes bc simply pattern matching and removing unrelated refs can render the rest of the messaging illegible so we have to trust the agent to adjust according to each specific context when removing those unrelated refs.


nforro (approved):
LGTM

PR #746, opened 2026-08-10
Phase 1# of postponed issues reprocessing 

jpodivin:
Implementing design from [research](https://github.com/packit/research/blob/main/research/postponed-issues-reprocessing/index.md) by @majamassarini .

Postponed issues are now split into several categories, depending on the reason for postponement. In each of these cases, a comment, in a well defined format, is posted under the given JIRA item. This comment is later processed by the `comment_parser` and used to determine if the postponement label can be removed, and processing resumed.

Sweep is executed using one of 4 strategies:

-  Y-stream checks if dependencies have been shipped in Z stream, using references in the comment and `check_build_in_buildroot`.
- PR Pending uses `ymir.supervisor.gitlab_utils.gitlab_api_get` to check state of MRs. If they are merged, the issue is pushed back into triage queue.
- No Patch simply sends all issues marked with `ymir_postponed_no_patch` back into triage queue, subject to limit of 20 per run.
- Dependency checks if the build is in the Y-stream, again using `check_build_in_buildroot`

All with their own cronjobs.

## Notes:

- Sweep jobs make use of code already developed for supervisor.
- Unit tests, as well as minimal integration tests were implemented.
- Minor modification was made to `gitlab_api_get` function, so it can be used by sweep jobs.
- The `ymir_abandoned` label has been implemented as part of this PR, although it is going to be used only in phase 2#.
- Container image is very similar to that for supervisor, we can discuss building of a base image for both perhaps. But I would rather leave that for a follow up.


Containerfile.sweep
```diff
@@ -0,0 +1,55 @@
+FROM quay.io/centos/centos:stream10
```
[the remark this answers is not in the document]
jpodivin: The `Containerfile.supervisor` uses the same base image.

ymir/sweep/base.py
```diff
@@ -0,0 +1,199 @@
...
+        """Push issue to triage queue, then remove the postponement label.
+
+        Redis is written first so that a subsequent Jira API failure leaves
+        the issue labelled (retried on the next sweep) rather than
+        unlabelled but missing from the queue.  The comment is posted
+        atomically with the label removal when provided.
+        """
+        task = Task.from_issue(issue_key)
+        redis_conn.lpush(RedisQueues.TRIAGE_QUEUE.value, task.to_json())
+        remove_issue_label(issue_key, self.label.value, comment=comment)
```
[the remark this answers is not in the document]
jpodivin: Good point.

ymir/sweep/pr_pending.py
```diff
@@ -0,0 +1,108 @@
...
+                issue_key=issue_key,
+                action="error",
+                detail=(f"Blocker reference is not a recognisable GitLab MR URL for {issue_key}: {mr_url!r}"),
+            )
+
+        host, project_path, mr_iid = m.group(1), m.group(2), m.group(3)
+        path = f"projects/{urlquote(project_path, safe='')}/merge_requests/{mr_iid}"
+
+        try:
+            mr_data = gitlab_api_get(path, gitlab_url=f"https://{host}")
```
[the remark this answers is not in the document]
jpodivin: I think this is a good idea, but it should be fixed in supervisor, as a general hardening change.

ymir/sweep/comment_parser.py
```diff
@@ -0,0 +1,77 @@
...
+        ``CommentData`` with blocker_reference, pending_issues, summary, and
+        comment_id.  Returns ``None`` if no Ymir comment is found.  Fields
+        that are absent from the comment are ``None``/empty; each sweep
+        strategy validates the fields it needs.
+    """
+    ymir_comment = None
+    for comment in reversed(issue.comments):
+        if _YMIR_COMMENT_MARKER in comment.body:
+            ymir_comment = comment
+            break
```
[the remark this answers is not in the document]
jpodivin: True, let`s limit it to triage comments.

ymir/common/constants.py
```diff
@@ -159,7 +159,16 @@ class JiraLabels(Enum):
...
 
+    # Deprecated general postponed label,
+    # replaced by labels with reason for postponement
     TRIAGED_POSTPONED = "ymir_triaged_postponed"
+
+    # Reasons for postponing implementation
+    YMIR_POSTPONED_DEPENDENCY = "ymir_postponed_dependency"
+    YMIR_POSTPONED_Y_STREAM = "ymir_postponed_y_stream"
+    YMIR_POSTPONED_NO_PATCH = "ymir_postponed_no_patch"
+    YMIR_POSTPONED_PR_PENDING = "ymir_postponed_pr_pending"
```
lbarcziova, line 205: these should be added in https://gitlab.cee.redhat.com/jotnar-project/cve-scope/-/blob/main/check_filter.py?ref_type=heads#L39 to be excluded from the fetcher filter additionally?

ymir/sweep/y_stream.py
```diff
@@ -0,0 +1,139 @@
...
+
+class YStreamSweep(SweepStrategy):
+    """Checks whether Z-stream dependencies have shipped for Y-stream issues.
+
+    For each issue the strategy:
+
+    1. Reads the blocker Jira issue key(s) from the comment.
+    2. For each blocker: fetches the issue via ``jira_utils.get_issue()``
+       and reads its ``fixed_in_build`` field.
+    3. If ANY blocker has no ``fixed_in_build`` → still blocked.
```
lbarcziova: I think the logic is that it is enough if *any of the blockers is resolved*, we can proceed, not the other way. Can be taken from the cve eligibility check function.

ymir/sweep/y_stream.py
```diff
@@ -0,0 +1,139 @@
...
+    4. Derives the Y-stream ``target_branch`` from the issue's
+       ``fix_versions`` (e.g. ``rhel-9.6.0`` → ``c9s``).
+    5. Reads ``dep_component`` from each blocker's ``components`` list.
+    6. Calls ``check_build_in_buildroot()`` for each blocker.
+    7. Unblocks only when ALL blockers are resolved.
```
lbarcziova: this logic doesn't seem correct, it should be mirroring what we already do in initial triage, iirc we need to be checking errata for this case

ymir/sweep/pr_pending.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class PRPendingSweep(SweepStrategy):
+    """Checks whether the upstream GitLab merge request has been merged.
+
+    For each issue the strategy:
+
+    1. Reads the MR URL from the comment's ``*Blocker*:`` line.
+    2. Parses ``(host, project_path, mr_iid)`` from the URL.
+    3. Fetches the MR via the GitLab REST API.
```
lbarcziova: upstreams can be on other forges too, I'm wondering if we can cover everything here by code (get inspired from ogr), or we will need LLM to cover some cases? Or cover in this branch just the most common one, i.e. Github, Gitlab including custom domains?

ymir/sweep/no_patch.py
```diff
@@ -0,0 +1,114 @@
...
+
+_RECHECK_COMMENT = "Sweep: no-patch recheck — pushing to triage queue for fresh evaluation."
+
+
+class NoPatchSweep(SweepStrategy):
+    """Pushes no-patch issues back to the triage queue for re-evaluation.
+
+    Overrides ``run()`` because the base class's comment-parsing loop
+    does not fit this strategy: there is no blocking condition to check
+    per issue; instead the strategy unconditionally re-queues eligible
```
lbarcziova, line 37: if these issues get re-triaged on every sweep, does it mean also duplicating comments on each run?

ymir/agents/triage_agent.py
```diff
@@ -496,17 +500,30 @@ async def check_cve_eligibility(state):
...
+                    )
+                    return "comment_in_jira"
                 logger.info(
                     f"Issue {state.jira_issue} postponed — waiting for "
                     f"{len(pending)} Z-stream issue(s): {pending}. "
                     f"Reason: {state.cve_eligibility_result.reason}"
                 )
                 state.triage_result = OutputSchema(
-                    resolution=Resolution.POSTPONED,
+                    resolution=Resolution.POSTPONED_DEPENDENCY,
```
lbarcziova: this should be the pending z-stream case

ymir/agents/triage_agent.py
```diff
@@ -496,17 +500,30 @@ async def check_cve_eligibility(state):
...
                     data=PostponedData(
                         summary=state.cve_eligibility_result.reason,
                         pending_issues=pending,
                         jira_issue=state.jira_issue,
+                        blocker_reference=pending[0] if len(pending) == 1 else None,
```
lbarcziova: why are we picking here just the first one?

ymir/agents/prompts/triage/prompt.j2
lbarcziova, line 1: I don't see any guidance for the agent for categorising the other types of postponed scenarios, e.g. the pending MR, no patch etc. Have you tested any of these? I think we should add explicit guidance about each of them.

ymir/agents/prompts/triage/output_format.j2
```diff
@@ -67,11 +67,12 @@ mentions a build side-tag. Do not set it otherwise.
 Postponed resolution (rebuild waiting for dependency):
 ```json
 {
-    "resolution": "postponed",
+    "resolution": "postponed_dependency",
```
lbarcziova, line 70: similar as for the triage prompt generally, I think we should give more examples (probably one for each postponed category) in here for the agent

ymir/sweep/comment_parser.py
```diff
@@ -0,0 +1,82 @@
...
+from dataclasses import dataclass
+
+from ymir.supervisor.supervisor_types import FullIssue
+
+_YMIR_TRIAGE_AGENT_COMMENT_MARKER = "Output from Ymir Triage Agent"
+
+_BLOCKER_RE = re.compile(r"^\*Blocker\*:\s*(.+)$", re.MULTILINE)
+_PENDING_ISSUE_RE = re.compile(r"^\* ([A-Z]+-\d+)$", re.MULTILINE)
+_SUMMARY_RE = re.compile(r"^\*Summary\*:\s*(.+)$", re.MULTILINE)
+
```
lbarcziova: this seems a bit fragile, I don't have an idea right now how to make it more robust, will think about it

lbarcziova (commented):
thanks for working on this, this will be great once deployed in production. I like that for most of the re-checks it's just code, so won't cost us too much.

I have few initial questions regarding some implementation details.

ymir/common/models.py
```diff
@@ -588,10 +613,16 @@ def format_for_comment(self, auto_chain: bool = False) -> str:
                     heading = "*Waiting for*:"
                 else:
                     heading = "*Waiting for at least one of*:"
+                blocker_line = (
+                    f"\n*Blocker*: {self.data.blocker_reference}" if self.data.blocker_reference else ""
```
lbarcziova: if we are relying on this for sweeps, we would need to also adjust the existing comments if we do the relabeling

ymir/sweep/y_stream.py
lbarcziova: thinking about it, ideally we should have this logic only on one place to avoid duplication/diverging

ymir/common/constants.py
jpodivin: Yes.

ymir/agents/prompts/triage/output_format.j2
jpodivin: Agreed,

ymir/sweep/y_stream.py
jpodivin: So should this be routed directly to triage instead?

lbarcziova: I don't think so, I would just extract the code logic and use it on both places. @majamassarini wdyt?

jpodivin: I'm going by the original research doc https://github.com/packit/research/blob/main/research/postponed-issues-reprocessing/index.md 

Although, it is self contradictory when it describes the y-stream sweep.

ymir/sweep/pr_pending.py
jpodivin: Github is easy enough. But for anything else we are going to need an agent.

ymir/agents/triage_agent.py
jpodivin: Right.

ymir/sweep/no_patch.py
jpodivin: Yes. Although I would argue that it is reasonable. We should make it clear that triage did happen, and what blockers were found.

ymir/sweep/comment_parser.py
jpodivin: I have considered couple of options, all the way to attaching json with metadata to the comment. None of them were exactly elegant. This way, it's at least something maintainers can read.

ymir/sweep/y_stream.py
jpodivin: As an alternative, I'll use the `CheckCveTriageEligibilityTool` directly.

ymir/common/models.py
jpodivin: I don't know if I've said it before, but I'd like to perform a mass check on all postponed issues as a followup.

ymir/agents/prompts/triage/prompt.j2
```diff
@@ -350,11 +371,17 @@ You must decide between one of the following actions. Follow these guidelines to
...
      Set summary to explain that rebuild is waiting for the dependency to ship,
      and set pending_issues to the dependency issue key.
+     Set blocker_reference to the dependency issue key.
      Also set package, fix_version, cve_id, dependency_issue, and dependency_component
      (same values as you would for a rebuild resolution).
+   * Note: "postponed_y_stream" is typically set automatically by the CVE
+     eligibility system when a Y-stream issue depends on Z-stream fixes.
+     You should NOT set this resolution manually in most cases — the
+     eligibility check handles it. Only use it if you explicitly identify
+     that the issue is Y-stream and has Z-stream blockers.
```
lbarcziova: I think maybe we should even remove this, and just forbid it generally

ymir/sweep/no_patch.py
lbarcziova: but this is the case of no upstream patch, and if we keep retrying every day 20 times, this means 20 almost identical comments. To me this seems really spammy. How complex would be just reusing the one already added comment? Or start with not adding the comment if the state is the same? 

On a related note, we could maybe add a small headsup in the comments about postponed_* category, something along the lines `This issue will be monitored by Ymir and picked up once it gets unblocked`, potentially even saying how often we recheck it - but this might be just added to the docs page too. Wdyt?

lbarcziova (commented):
I see some great improvements here, I really like the prompt changes; just few (hopefully last) notes, and something raised by Claude:

```
  1. GitLab Bearer token sent to any gitlab.* host (security, low-severity)

  _GITLAB_MR_RE = re.compile(r"https://(gitlab\.[^/]+)/...") still accepts any gitlab.* domain. The extracted host is passed to gitlab_api_get(gitlab_url=f"https://{host}"), which appends the Bearer token to
  the request. A crafted blocker_reference pointing to a rogue gitlab.attacker.com host would exfiltrate the token. jpodivin deflected to the supervisor layer but the risk sits in pr_pending.py — the regex
  is where the host is extracted and trusted. Since blocker_reference values come from Jira comments written by the LLM (not from a curated allowlist), this is exploitable if a Jira comment is compromised or
  the LLM hallucinates a hostile URL.

```

jpodivin:
> I see some great improvements here, I really like the prompt changes; just few (hopefully last) notes, and something raised by Claude:
> 
> ```
>   1. GitLab Bearer token sent to any gitlab.* host (security, low-severity)
> 
>   _GITLAB_MR_RE = re.compile(r"https://(gitlab\.[^/]+)/...") still accepts any gitlab.* domain. The extracted host is passed to gitlab_api_get(gitlab_url=f"https://{host}"), which appends the Bearer token to
>   the request. A crafted blocker_reference pointing to a rogue gitlab.attacker.com host would exfiltrate the token. jpodivin deflected to the supervisor layer but the risk sits in pr_pending.py — the regex
>   is where the host is extracted and trusted. Since blocker_reference values come from Jira comments written by the LLM (not from a curated allowlist), this is exploitable if a Jira comment is compromised or
>   the LLM hallucinates a hostile URL.
> ```

We could have a whitelist for this. But it will add to the maintenance burden.

ymir/sweep/no_patch.py
jpodivin: I'm fine with removing the recheck comment here. And I'll put a mention of monitoring in comments made for postponement.

ymir/agents/triage_agent.py
jpodivin: You are right, that doesn't make sense.

templates/sweep.env
```diff
@@ -0,0 +1,9 @@
+# Required: Jira instance URL
```
majamassarini, line 1: I am wondering if, instead of creating this new secret file, we should reuse the `jira-issue-fetcher.env` and potentially rename it - if clearer. This would limit the places that need to be updated.

majamassarini (commented):
Beside the comment below and this small finding from Claude, this LGTM.

```
PostponedData.blocker_references semantically required but schema-optional (ymir/common/models.py:1167)

  Field is list[str] | None with default=None, but PRPendingSweep returns an error if None for postponed_pr_pending issues. The schema doesn't enforce the semantic requirement.

  Fix: Either make the field required in the schema or handle None gracefully in PRPendingSweep.
```

jpodivin:
> Beside the comment below and this small finding from Claude, this LGTM.
> 
> ```
> PostponedData.blocker_references semantically required but schema-optional (ymir/common/models.py:1167)
> 
>   Field is list[str] | None with default=None, but PRPendingSweep returns an error if None for postponed_pr_pending issues. The schema doesn't enforce the semantic requirement.
> 
>   Fix: Either make the field required in the schema or handle None gracefully in PRPendingSweep.
> ```

That is not entirely correct. The `PostponedData` is used for all cases of postponement, and not all of them need references for blockers. So we can't make it mandatory. It is true, that model can hypothetically forget to fill it, if it sets resolution to `POSTPONED_PR_PENDING`. Even though that would mean ignoring the prompt. I can put a validator in, but turning this into a mandatory field wouldn't work.

ymir/agents/prompts/triage/prompt.j2
jpodivin: Done

templates/sweep.env
jpodivin: But some of the vars in that file, like `QUERY`, have nothing to do in sweep jobs. 

Using a single file for both, would lead to confusion regarding what the cronjobs need. Not just for people, agents too.

Having a single file for two services, would also mean that variables for issue fetcher, would get in sweep containers. Finally, while there is already divergence today, there is no guarantee that it won't be even greater in the future. And then we would have to create a new file anyway.

lbarcziova (commented):
I think we are close to merging this, thanks for all the changes! Few last things to consider:                                                                                                                                                                               
- The sweep JQL only filters by label, I think we should we add a status gate to skip issues that are already Closed too (to prevent scenarios where maintainer fixes the Jira before our sweep gets to it)
- Can we adjust the triage comment for postponed resolutions to mention that the issue will be picked up for reprocessing periodically by the sweep? (we can also communicate this on rhel-devel after tested in prod)       
- Do you still plan to re-label and adjust the comments on existing postponed issues with ymir_triaged_postponed? What's the timeline for this and merging this PR                                                                                         
-  the no_patch path will still post a new Jira comment on every sweep cycle iiuc when the agent still finds no patch. Could we add a guard to check if there is the Ymir comment with the same resolution already and skip commenting for that case? 
- Task.from_issue(issue_key) in on_unblock() always uses the default user_triggered=False. Is this intentional? If a maintainer originally triggered triage via `ymir_todo` and the issue ends up postponed and later re-queued by the sweep, they won't get a comment update when triage completes. Seems probably minor, but we could eventually store the `user_triggered`

jpodivin:
> I think we are close to merging this, thanks for all the changes! Few last things to consider:
> 
>     * The sweep JQL only filters by label, I think we should we add a status gate to skip issues that are already Closed too (to prevent scenarios where maintainer fixes the Jira before our sweep gets to it)
> 
>     * Can we adjust the triage comment for postponed resolutions to mention that the issue will be picked up for reprocessing periodically by the sweep? (we can also communicate this on rhel-devel after tested in prod)
> 
>     * Do you still plan to re-label and adjust the comments on existing postponed issues with ymir_triaged_postponed? What's the timeline for this and merging this PR
> 
>     * the no_patch path will still post a new Jira comment on every sweep cycle iiuc when the agent still finds no patch. Could we add a guard to check if there is the Ymir comment with the same resolution already and skip commenting for that case?
> 
>     * Task.from_issue(issue_key) in on_unblock() always uses the default user_triggered=False. Is this intentional? If a maintainer originally triggered triage via `ymir_todo` and the issue ends up postponed and later re-queued by the sweep, they won't get a comment update when triage completes. Seems probably minor, but we could eventually store the `user_triggered`

In order:
1. Why not.
2. The comment already says that: `This issue will be monitored by Ymir and processed once it is unblocked` but sure, I can say that the check is done periodically.
3. Yes. It has to be done after this PR is merged, in a follow up, because it is out of scope of this feature. Some of those issues are bound to be resolved in the meantime. And for the rest, it won't make much difference compared to the previous state of affairs.
4. I'll put the guard in the triage agent.
5. Yes it is intentional. The task in this case is triggered by the sweep. Maybe the user was indirectly involved at a some point. But it's not an important distinction.

lbarcziova:
thanks, sounds good, as for 2, I overlooked that, the message looks sufficient

ymir/agents/triage_agent.py
```diff
@@ -1165,6 +1190,23 @@ async def comment_in_jira(state):
                     f"(resolution={state.triage_result.resolution.value}, not user-triggered)"
                 )
                 return Workflow.END
+            # skip re-posting a comment when this issue already carries
+            # a Ymir comment for the same postponement resolution.
+            # User-triggered runs always get fresh confirmation, so they bypass this guard.
```
lbarcziova, line 1221: haven't we discussed yesterday that the user_triggered will actually not be stored/will be set to False?

lbarcziova (approved):
LGTM! 🚀 

Given the scope of this change, we should coordinate the deployment and monitoring with Skald. Do you think it's safe to deploy this week, given the long weekend - will we have time to catch and fix any follow-up issues? If you see meaningful risk here, I'd lean toward deploying Monday instead.

ymir/agents/triage_agent.py
jpodivin: Yes. The sweep has it set to `False` for now, because it is conceived as a cronjob. But it is conceivable that we would want to see how postponed issues are handled on demand. This way we can bypass the check and see what happens, otherwise we wouldn't have that option.

jpodivin:
> LGTM! 🚀
> 
> Given the scope of this change, we should coordinate the deployment and monitoring with Skald. Do you think it's safe to deploy this week, given the long weekend - will we have time to catch and fix any follow-up issues? If you see meaningful risk here, I'd lean toward deploying Monday instead.

If I am to err, I would like to err on the side of caution. The Monday deployment is preferable, in my opinion.

lbarcziova:
@jpodivin hi, can we merge this, or does this wait on something else?

jpodivin:
> @jpodivin hi, can we merge this, or does this wait on something else?

Honestly, I thought we already did.

lbarcziova (approved):
can you just please check the failing CI?

jpodivin:
> can you just please check the failing CI?

The problem is in the centos:stream10 base image. All of our images based on that are failing in the build.

PR #750, opened 2026-08-12
Reasoning agent context management

TomasKorbar:
This PR includes:

1. Dynamic context management implementation
2. Making Reasoning Agent note about doing things sequentially optional
3. Fix for parallel tool calling

### DIFF
--- a/ymir/agents/reasoning_agent/context_management.py
+++ b/ymir/agents/reasoning_agent/context_management.py
@@ -1,0 +1,237 @@
+"""Agent-driven conversation context compaction for ReasoningAgent.
+
+The agent authors a durable summary via ``manage_context`` in the same inference
+turn as other tools. The tool only schedules compaction; the runner applies it
+after ``asyncio.gather`` so parallel tool execution cannot race on memory.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Self
+
+from beeai_framework.backend import (
+    AnyMessage,
+    AssistantMessage,
+    MessageToolCallContent,
+    ToolMessage,
+    UserMessage,
+)
+from beeai_framework.context import RunContext
+from beeai_framework.emitter import Emitter
+from beeai_framework.tools import StringToolOutput, Tool, ToolRunOptions
+from pydantic import BaseModel, Field
+
+if TYPE_CHECKING:
+    from ymir.agents.reasoning_agent.types import ReasoningAgentRunState
+
+YMIR_PROTECTED_META_KEY = "ymir_protected"
+YMIR_ROLE_META_KEY = "ymir_role"
+YMIR_CONTEXT_SUMMARY_META_KEY = "ymir_context_summary"
+MANAGE_CONTEXT_TOOL_NAME = "manage_context"
+
+
+class ManageContextSchema(BaseModel):
+    durable_summary: str = Field(
+        description=(
+            "Concise durable facts still needed for further work: paths, IDs, "
+            "current hypothesis, TF/MR identifiers. "
+            "You MUST include every unsuccessful approach already tried — what was "
+            "attempted, why it failed, and that it must not be retried. "
+            "Do not restate the task or system instructions."
+        )
+    )
+    keep_recent_exchanges: int = Field(
+        default=1,
+        ge=1,
+        description=(
+            "Number of most recent assistant+tool exchanges to keep verbatim "
+            "(including the current turn). Older exchanges are replaced by durable_summary."
+        ),
+    )
+
+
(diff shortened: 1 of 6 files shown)


ymir/agents/reasoning_agent/context_management.py
```diff
@@ -0,0 +1,234 @@
...
+    )
+
+
+class ManageContextTool(Tool[ManageContextSchema, ToolRunOptions, StringToolOutput]):
+    name = MANAGE_CONTEXT_TOOL_NAME
+    description = (
+        "Compact conversation memory by replacing older tool exchanges with a durable "
+        "summary you provide. Call this in the SAME turn as another useful tool when "
+        "history contains dead ends, large obsolete dumps, or failed approaches. "
+        "Do not call it alone. Never summarize away the task or system instructions."
```
jpodivin, line 61: I think we ought to have stronger emphasis on keeping the record of unsuccessful approaches taken. Just to make sure that the agent doesn't try them again.

ymir/agents/reasoning_agent/context_management.py
```diff
@@ -0,0 +1,234 @@
...
+
+    protected, exchanges = partition_exchanges(list(state.memory.messages))
+    keep_n = pending.keep_recent_exchanges
+
+    if keep_n >= len(exchanges):
+        to_drop: list[list[AnyMessage]] = []
+        to_keep = exchanges
+    else:
+        to_drop = exchanges[:-keep_n]
+        to_keep = exchanges[-keep_n:]
```
jpodivin: I think we can simplify this.

jpodivin (changes requested):

ymir/agents/reasoning_agent/prompts.py
```diff
@@ -94,7 +95,9 @@ class ReasoningAgentSystemPromptInput(BaseModel):
 - Use markdown syntax to format code snippets, links, JSON, tables, images, and files.
 - If the provided task is unclear, ask the user for clarification.
 - Do not refer to tools or tool outputs by name when responding.
-- Always take it one step at a time. Don't try to do multiple things at once.
+{{#sequential_tool_calls}}- Always take it one step at a time. Don't try to do multiple things at once.
+{{/sequential_tool_calls}}{{^sequential_tool_calls}}- You may submit multiple independent tool calls in a single turn to save time. Only keep calls sequential when one depends on the result of another.
+{{/sequential_tool_calls}}
```
nforro: Why such weird formatting?
```suggestion
{{#sequential_tool_calls}}
- Always take it one step at a time. Don't try to do multiple things at once.
{{/sequential_tool_calls}}
{{^sequential_tool_calls}}
- You may submit multiple independent tool calls in a single turn to save time.
  Only keep calls sequential when one depends on the result of another.
{{/sequential_tool_calls}}
```

ymir/agents/reasoning_agent/context_management.py
TomasKorbar: Good point. Addressed.

ymir/agents/reasoning_agent/context_management.py
TomasKorbar: Done

ymir/agents/reasoning_agent/context_management.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def _assistant_has_provider_visible_output(msg: AssistantMessage) -> bool:
+    """Return True if the message has text or tool_calls (not reasoning-only).
+
+    Anthropic/Vertex reject assistant messages whose final block is ``thinking``.
+    BeeAI sends thinking via ``meta['thinking_blocks']`` and ignores
+    ``MessageReasoningContent`` on the wire, so a message with only reasoning
+    content (or empty content + thinking_blocks) becomes an invalid
+    thinking-only assistant turn.
```
nforro: BeeAI sends `meta['thinking_blocks']` because Anthropic models require it. I don't really understand the point of this docstring, but it seems to blame BeeAI for something I've implemented and contributed only because of Anthropic quirks.

jpodivin (dismissed):

ymir/agents/reasoning_agent/context_management.py
TomasKorbar: We are stripping the manage_context tool call to save tokens and if it is the only thing in the AssistantMessage then there is no visible output but the meta['thinking_blocks'] remain which vertex rejects with 400. This processing ensures that AssistantMessage which has also different stuff than manage_context tool call will remain with the other tool calls and thinking blocks, but the one where only thinking blocks remain is removed.

jpodivin (approved):

PR #754, opened 2026-08-12
Improve reporting during consolidation

jpodivin:
Unfortunately, in order to have agent post comments on jiras consistently, I had to slightly modify the workflow. Alternatives involved more convoluted rewrites that I liked even less. 

The `MRConsolidationOutputSchema` now includes fields `status` and `status_detail`. Valid values of the `status` are limited
to `ConsolidationStatus` literal.

The `handle_failure` step now routes to `update_jira_issues` step, only logging error and optionally filling `jira_issues_collected` if it is not set.

The `update_jira_issues` returns `Workflow.END` if the status field of `MRConsolidationOutputSchema` is set to "failed", "nothing_to_consolidate" or "error".

There is a bit of defensive programming between lines 1671 and 1680, in cases the status is not recognized.
I'm open to removing it, since such a case should trigger a validation error. But, I would also understand keeping it in.

### DIFF
--- a/ymir/agents/mr_consolidation_agent.py
+++ b/ymir/agents/mr_consolidation_agent.py
@@ -188,6 +188,19 @@ def _extract_cves_from_cve_footer_lines(text: str) -> list[str]:
     return sorted(set(cves))
 
 
+def _extract_jira_from_mr_descriptions(mrs: list[dict]) -> list[str]:
+    """Extract Jira issue keys from MR description Resolves:/Related: footers.
+
+    Used in auto mode to seed jira_issues_collected before the git clone is
+    available, so early-exit paths can still post Jira comments.
+    """
+    issues: list[str] = []
+    for mr in mrs:
+        desc = mr.get("description") or ""
+        issues.extend(_extract_jira_issues_from_resolves_footer_lines(desc))
+    return sorted(set(issues))
+
+
 def _extract_jira_issues_from_resolves_footer_lines(text: str) -> list[str]:
     """Extract RHEL keys only from ``Resolves:`` / ``Related:`` lines.
 
@@ -349,10 +362,14 @@ async def _resolve_source_issues(
             )
             state.consolidation_result = MRConsolidationOutputSchema(
                 success=False,
-                status=f"Could not find an open MR for {issue_key}",
+                status="mr_not_found",
+                status_detail=f"Could not find an open MR for {issue_key}",
                 error=f"No open MR matching {issue_key} in {project_path}",
             )
-            return Workflow.END
+            # When have no collected issues post updates under those from issue_keys
+            if not state.jira_issues_collected:
+                state.jira_issues_collected = issue_keys
+            return "update_jira_issues"
 
         matched_mrs.append(mr)
         logger.info(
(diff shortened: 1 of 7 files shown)


ymir/agents/tasks.py
```diff
@@ -965,5 +973,26 @@ async def try_submit_consolidation_job(
     )
     if submitted:
         logger.info("Submitted consolidation job for %s/%s", package, dist_git_branch)
+
+        # Post a Jira comment notifying that consolidation has been triggered
```
lbarcziova: iiuc the code here, this is run before the actual eligibility check (list_open_mrs checking >= 2 MRs). So users could often see:
 > "Your MR has been queued... a consolidated MR will be created automatically"

and later on:

> "Fewer than 2 MRs; nothing to do."

I think we might want to skip this and comment about it just when we are 100% sure, to avoid spammy behaviour

lbarcziova: instead, I would prefer if we had a dedicated comment once the pre-checks are run and satisfied with something like "Consolidation is starting", wdyt?

ymir/agents/mr_consolidation_agent.py
```diff
@@ -375,9 +392,13 @@ async def _resolve_source_issues(
         logger.info("Fewer than 2 unique MRs resolved, nothing to consolidate")
         state.consolidation_result = MRConsolidationOutputSchema(
             success=True,
-            status="Fewer than 2 unique MRs resolved; nothing to do.",
+            status="nothing_to_consolidate",
+            status_detail="Fewer than 2 unique MRs resolved; nothing to do.",
```
lbarcziova, line 395: I feel this might be spammy as well (basically commenting additionally for all cases of just 1 PR iiuc?) -- if we have instead commenting on consolidation starting, I would suppress this

lbarcziova (commented):
thank you, there are some good improvements here. I commented about things I would change to make this less spammy. @TomasKorbar could you have a look too?

ymir/agents/tasks.py
jpodivin: My ideal state, would be that no maintainer has to ask: "What is Ymir waiting for?" Obviously we can't control that, but we should at least be able to tell them: "Check Jira comments." 

So I'm open to moving the triggered message to the point when all conditions are cleared, but we should notify on all cases when the consolidation is impossible.

lbarcziova: > My ideal state, would be that no maintainer has to ask: "What is Ymir waiting for?" Obviously we can't control that, but we should at least be able to tell them: "Check Jira comments."

I totally agree on this, but this needs to be also aligned with not spamming them, which commenting if there is just 1 MR seems like to me. Wondering if we could maybe utilise labels as well?

jpodivin: I think labels may be worse. There is already a lot of them, and we should make people memorize them all. How about modifying the comment instead? That wouldn't create a new comment, and it would provide all the information.

lbarcziova: that makes sense, but regarding the 1 MR scenario, I would still avoid commenting on that (what I described in https://github.com/packit/ai-workflows/pull/754#discussion_r3783148533)

TomasKorbar: I would be too for removing this and the 1MR comment. Proper fix for non-transparency during job submission is to make information how do the queues work accessible in trace server.

ymir/agents/mr_consolidation_agent.py
TomasKorbar, line 845: Could you please set state.consolidation_result appropriately for these build failures as well? I noticed that there is no indication about the error when i was looking through your changes.

TomasKorbar (changes requested):

ymir/agents/tasks.py
jpodivin: Ok. No comments on the consolidation trigger.

jpodivin: I've removed both, if it is fine like this, I'll squash those patches so we have clean history for the merge.

ymir/agents/mr_consolidation_agent.py
jpodivin: Done

PR #756, opened 2026-08-15
Fix modular duplicate detection to compare streams, not just boolean

fila43:
## Summary

- **Stream-aware duplicate detection**: `_check_duplicate_tracker()` previously used `is_modular()` which only returned a boolean (modular vs non-modular). This caused false duplicate matches between issues with different module streams (e.g. `postgresql:16` vs `postgresql:18`). Replaced with `parse_module_stream()` which compares the full `(module, stream)` tuple.
- **Generalized modular summary regex**: `MODULAR_SUMMARY_PREFIX` had a CVE-specific optional prefix `(?:CVE-\d{4}-\d+\s+)?` that would fail on summaries with multiple CVEs or non-CVE prefixes. Replaced with generic `(?:\S+\s+)*` that skips any tokens before the `module:stream/` pattern.
- **Shared `parse_module_stream()` utility**: Extracted stream-parsing logic into `version_utils.py` so both `jira.py` and `triage_agent.py` use the same function.

Supersedes the stream-comparison gap left by #743.

## Test plan

- [x] Unit tests: 822 passed across all three test suites (privileged tools, common, agents) in containers
- [x] New tests for stream-aware dedup: same stream = dup, different stream = not dup, modular vs non-modular, mixed streams
- [x] New tests for `parse_module_stream()` including multi-CVE prefix
- [x] Manual dry-run against real modular ticket RHEL-241451 (`postgresql:16/postgis`) completed successfully

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ymir/agents/tests/unit/test_triage_agent.py
+++ b/ymir/agents/tests/unit/test_triage_agent.py
@@ -7,7 +7,6 @@ from ymir.agents.triage_agent import (
     TriageState,
     _build_reproducer_input,
     _map_version_to_module_branch,
-    _parse_module_summary,
     _should_update_jira,
     determine_target_branch,
 )
@@ -23,7 +22,7 @@ from ymir.common.models import (
     TriageEligibility,
     TriageOutputSchema,
 )
-from ymir.common.version_utils import is_modular
+from ymir.common.version_utils import is_modular, parse_module_stream
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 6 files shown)


jpodivin (approved):
I like this. Although couple more test cases for strings that should result in `None` return from `parse_module_stream` would be nice.

lbarcziova (approved):
LGTM, can you please rebase?

fila43:
rebased

PR #757, opened 2026-08-17
Shared rules implementation

lbarcziova:


ymir/tools/privileged/shared_rules.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def _cache_and_return(self, registry: dict[str, list[str]] | None) -> dict[str, list[str]]:
+        self._registry_cache = registry
+        self._registry_fetched_at = time.monotonic()
+        return registry or {}
+
+    async def _fetch_registry(self) -> dict[str, list[str]]:
+        if (
+            self._registry_fetched_at
+            and (time.monotonic() - self._registry_fetched_at) < REGISTRY_TTL_SECONDS
```
TomasTomecek, line 62: this definitely makes a lot of sense, we don't need to fetch the same file multiple times per minute

I do wonder though if we could just check if the file changed using the respective HTTP header

TomasTomecek (dismissed):
nicely done

ymir/tools/privileged/shared_rules.py
lbarcziova: do you mean just for the periodic recheck? I didn't consider it since the yaml should be rather small, but if this needs to scale up later, we can definitely add it, good point

nforro:
@lbarcziova It seems Qodo doesn't like you.

nforro (dismissed):
LGTM

nforro (approved):

PR #759, opened 2026-08-18
Freeze pre-commit repo references and update secrets baseline

jpodivin:
By pinning repos to hashes instead of tag we prevent potential supply chain attack on developer machines. This also counts as an upgrade of the ruff pre commit. 

Since the hashes do look somewhat like secrets, they have to be included in the baseline. New `.secrects.baseline` was generated with `detect-secrets` version 1.5.0, which is the same one as used by the pre-commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v6.0.0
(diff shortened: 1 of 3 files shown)


jpodivin:
Pre-commit is failing due to non-compliant union. This should be fixed by: https://github.com/packit/ai-workflows/pull/758

majamassarini (approved):
LGTM

PR #763, opened 2026-08-18
Replace `get_jira_labels` use with `get_jira_issue_metadata`

jpodivin:
The function was removed but the call remained. Bulk of this PR are tests. They are a bit tricky, since the `process_task` function is closure inside main.


TomasKorbar (approved):
LGTM

PR #764, opened 2026-08-18
Reproducer fixes

TomasKorbar:
- Fix wrong retrieval of jira labels in reproducer_agent
- Fix context compaction when manage_context is called alone
- Wait for GitLab fork provisioning before returning fork URL

### DIFF
--- a/ymir/tools/privileged/gitlab.py
+++ b/ymir/tools/privileged/gitlab.py
@@ -98,6 +98,56 @@ async def _run_git_cmd(
 # Maintainer (40), Owner (50)
 DEVELOPER_ACCESS_LEVEL = 30
 
+_FORK_READY_IMPORT_STATUSES = frozenset({"finished", "none"})
+_FORK_READY_POLL_INTERVAL_SEC = 2.0
+_FORK_READY_TIMEOUT_SEC = 110  # leave margin under fork_repository tool timeout
+
+
+def _fork_api_project(fork: GitlabProject):
+    """Return a python-gitlab Project object suitable for import_status polling.
+
+    ``forks.create()`` returns a ``ProjectFork`` without ``refresh()``; always
+    fetch the full project by path for status polling.
+    """
+    return fork.service.gitlab_instance.projects.get(f"{fork.namespace}/{fork.repo}")
+
+
+def _wait_for_fork_ready(fork: GitlabProject) -> None:
+    """Block until GitLab finishes provisioning a fork's git repository.
+
+    Fork creation is asynchronous: the API returns before the repository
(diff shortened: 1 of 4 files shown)


ymir/tools/privileged/gitlab.py
```diff
@@ -98,6 +98,47 @@ async def _run_git_cmd(
...
+_FORK_READY_IMPORT_STATUSES = frozenset({"finished", "none"})
+_FORK_READY_POLL_INTERVAL_SEC = 2.0
+_FORK_READY_TIMEOUT_SEC = 110  # leave margin under fork_repository tool timeout
+
+
+def _wait_for_fork_ready(fork: GitlabProject) -> None:
+    """Block until GitLab finishes provisioning a fork's git repository.
+
+    Fork creation is asynchronous: the API returns before the repository
+    accepts git pushes. Poll ``import_status`` until the fork is ready.
```
lbarcziova, line 119: any ideas how is that we haven't hit this before for the dist-git repos?

lbarcziova (dismissed):
LGTM, thanks!

ymir/tools/privileged/gitlab.py
```diff
@@ -361,6 +402,7 @@ def get_fork():
             return None
 
         if fork := await asyncio.to_thread(get_fork):
+            await asyncio.to_thread(_wait_for_fork_ready, fork)
```
lbarcziova: actually, in case the fork existed before, do we still to call the wait method?

ymir/tools/privileged/gitlab.py
TomasKorbar: Reproducer hit this because the fork is done on the end of workflow and immediately after that the test is pushed, so it managed to hit this time between API return and fork operation finish

ymir/tools/privileged/gitlab.py
TomasKorbar: I added this as a precaution. Do you think i should remove it?

ymir/tools/privileged/gitlab.py
lbarcziova: aah I see, makes sense

ymir/tools/privileged/gitlab.py
lbarcziova: I would remove it as it's additional API call, we can add it if we bumped into this issue

TomasKorbar: okay

lbarcziova (dismissed):

lbarcziova (approved):

PR #765, opened 2026-08-19
Update filter for daily fetcher

lbarcziova:
We do not need to explicitly support early adopters anymore, as we process whole RHEL.


majamassarini (approved):

nforro (approved):

TomasTomecek (approved):
LGTM, tried in jira and worked well

PR #766, opened 2026-08-19
Fix: Skip buildroot check for Y-stream issues when dependency shipped from older Z-stream

majamassarini:
## Summary

Fixes incorrect postponement of Y-stream rebuild issues when their dependency was already shipped from an older Z-stream.

## Problem

When a Y-stream issue (e.g., rhel-10.3) depends on a package that was fixed in an older Z-stream (e.g., rhel-10.0.z) and is already Done-Errata, the triage agent would:

1. ✅ Correctly reason that the dependency is shipped and flows from CentOS Stream
2. ❌ But then check if the Z-stream build NVR exists in the Y-stream buildroot
3. ❌ Find it's not there (different NVR: `.el10_0` vs `.el10`)
4. ❌ Incorrectly postpone the issue

**Example:** RHEL-189822 (go-fdo-client, rhel-10.3) waiting for RHEL-189749 (golang, rhel-10.0.z, Done-Errata with golang-1.26.5-1.el10_0 shipped Aug 3).

## Solution

Added logic in `verify_rebuild_buildroot()` to:
- Check if dependency issue is from an older Z-stream
- Check if dependency is Done/Closed status (already shipped)
- If both true, skip the buildroot check entirely
- Proceed directly to rebuild consolidation

## Why This Works

When a Z-stream fix is Done-Errata, it has already been merged to CentOS Stream. Y-stream releases build from CentOS Stream, so they automatically get the fix. The buildroot will have the CS version (`.el10`), not the Z-stream version (`.el10_0`), but that's expected and correct.

## Testing

Tested with RHEL-189822:
- **Before:** Incorrectly postponed waiting for golang-1.26.5-1.el10_0
- **After:** Correctly resolved as rebuild (go-fdo-client needs to be rebuilt with fixed golang from c10s)

See test output showing correct resolution: https://github.com/packit/ai-workflows/issues/[issue-number]

### DIFF
--- a/ymir/agents/triage_agent.py
+++ b/ymir/agents/triage_agent.py
@@ -970,7 +970,7 @@ async def run_workflow(
             # flowed to CentOS Stream (and thus to newer Y-streams) — skip buildroot check.
             # Example: dependency targets rhel-10.0.z (Done-Errata), current issue targets
             # rhel-10.3 — the golang fix is already in c10s, no need to wait.
-            dep_status = dep_details.get("fields", {}).get("status", {}).get("name", "")
+            dep_status = (dep_details.get("fields", {}).get("status") or {}).get("name", "")
             dep_fix_versions = dep_details.get("fields", {}).get("fixVersions", [])
             dep_fix_version = dep_fix_versions[0].get("name", "") if dep_fix_versions else ""
 
--- a/ymir/agents/triage_agent.py
+++ b/ymir/agents/triage_agent.py
@@ -971,16 +971,26 @@ async def run_workflow(
             # Example: dependency targets rhel-10.0.z (Done-Errata), current issue targets
             # rhel-10.3 — the golang fix is already in c10s, no need to wait.
             dep_status = (dep_details.get("fields", {}).get("status") or {}).get("name", "")
+            dep_resolution = (dep_details.get("fields", {}).get("resolution") or {}).get("name", "")
             dep_fix_versions = dep_details.get("fields", {}).get("fixVersions", [])
             dep_fix_version = dep_fix_versions[0].get("name", "") if dep_fix_versions else ""
 
+            # Dependency is considered shipped if:
+            # 1. Status is "Done" (RHEL-only status indicating shipped), OR
+            # 2. Status is "Closed" AND resolution is "Done" or "Done-Errata"
+            # Closed with other resolutions (WONTFIX, NOTABUG, etc.) means NOT shipped.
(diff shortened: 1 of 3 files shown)


ymir/agents/triage_agent.py
```diff
@@ -966,6 +966,36 @@ async def verify_rebuild_buildroot(state):
...
+            dep_fix_version = dep_fix_versions[0].get("name", "") if dep_fix_versions else ""
+
+            # Dependency is considered shipped if:
+            # 1. Status is "Done" (RHEL-only status indicating shipped), OR
+            # 2. Status is "Closed" AND resolution is "Done" or "Done-Errata"
+            # Closed with other resolutions (WONTFIX, NOTABUG, etc.) means NOT shipped.
+            _SHIPPING_RESOLUTIONS = frozenset({"Done", "Done-Errata"})
+            dep_is_shipped = dep_status == "Done" or (
+                dep_status == "Closed" and dep_resolution in _SHIPPING_RESOLUTIONS
+            )
```
TomasTomecek, line 985: this part is definitely triggering me, why can't we have a single way in our Jira of telling if an issue is done and shipped `</rant>`

I'm really looking forward once we have the verification check implemented that verifies the dependency is installed during the build via root.log

TomasTomecek (approved):
LGTM, nicely done, my comment is just a rant 😅

nforro:
> * ✅ Correctly reason that the dependency is shipped and flows from CentOS Stream
> * ❌ But then check if the Z-stream build NVR exists in the Y-stream buildroot
> * ❌ Find it's not there (different NVR: `.el10_0` vs `.el10`)

Hm, isn't the problem rather here then?

> * ❌ Incorrectly postpone the issue

majamassarini:
> > * ✅ Correctly reason that the dependency is shipped and flows from CentOS Stream
> > * ❌ But then check if the Z-stream build NVR exists in the Y-stream buildroot
> > * ❌ Find it's not there (different NVR: `.el10_0` vs `.el10`)
> 
> Hm, isn't the problem rather here then?

Yes I think this is the actual problem, and it solves it not looking for the build at all since its already shipped. I don't know if it is the best solution. But it works. I tested it on the failing issue.

nforro:
> But it works. I tested it on the failing issue.

I'm not saying it doesn't work, but wouldn't it be more robust to adjust the NVR comparison so that it allows variance in dist tag? It could be a simpler fix at the same time. Is there a reason not to go that way?

PR #769, opened 2026-08-20
Enforce fixVersion validation when searching for rebuild dependencies

lbarcziova:
The triage and sibling rebuild agents search for dependency CVE trackers by CVE ID and component, but never validated that the found tracker targets the same RHEL version stream. With multiple z-streams now active (10.0.z, 10.1.z, 10.2.z, 10.3), the LLM could pick a tracker from a different stream, causing the buildroot check to use the wrong NVR and produce nonsensical postpone messages (e.g. "waiting for golang (golang-1.26.5-1.el10_0)" on a rhel-10.3 ticket).

Changes:
- Instruct triage agent to validate that dependency issues have a matching fixVersion (same major.minor) and never use trackers from different streams
- Add rebase/update ticket fallback when no CVE tracker exists for the target stream (with requirement to verify the rebase fixes the CVE)
- Fall back to clarification-needed instead of using a wrong-stream tracker
- Apply same fixVersion validation to sibling rebuild analysis

Assisted-by: Claude Opus 4.6

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -35,4 +35,6 @@ ignore = [
 "**/tests/**" = ["S101", "S105", "S106", "S107", "S108", "S603", "S607"]
 # Helper scripts shell out to known binaries by name.
 "scripts/**" = ["S603", "S607"]
+# LLM prompt strings contain JQL (SQL-like syntax) but are not SQL queries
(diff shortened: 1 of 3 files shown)


majamassarini (approved):
Thanks!

TomasKorbar (approved):
LGTM

PR #770, opened 2026-08-20
Reproducer improvements

TomasKorbar:


lbarcziova (commented):
overall LGTM, just these from Claude might be relevant to check:

1. Multi-CVE bracket format breaks MR matching (confirmed bug)

  _build_mr_title emits [CVE-A, CVE-B] for multiple CVEs, but _REPRODUCER_MR_BRACKET_CVE = re.compile(r"\[(CVE-\d{4}-\d+)\]") only matches a single CVE inside brackets — findall returns [] on a multi-CVE
  title. So _reproducer_mr_title_tags → _match_open_reproducer_mr silently fails for multi-CVE packages, and orchestration creates a duplicate MR instead of updating the existing one. No unit test covers the
  roundtrip.

  ---
  
  2. Dead code in _resolve_reproducer_mr_target (confirmed)

  reproducer_agent.py lines 351–354: the if result.adapted_existing and result.existing_mr_url: guard and the unconditional return below it produce the exact same tuple — fallback_branch is computed once and
  nothing between the two statements mutates result.existing_mr_url. The conditional is unreachable dead code. The practical consequence is that when matched is None but the agent reported an
  existing_mr_url, the caller gets matched_mr=None, so _build_mr_title can't accumulate existing JIRA tags into a regression MR title.

TomasKorbar:
@lbarcziova fixed.

lbarcziova (approved):
🙏

PR #772, opened 2026-08-24
Rename is_dependency_rebuild to include_in_rebuild in sibling analysis

lbarcziova:
The old name is_dependency_rebuild sounds like a question about the nature of the issue ("is this structurally a dependency rebuild?"), causing the agent to set it true for in-progress dependencies and then walk it back repeatedly.

The flag actually drives a single decision: whether the sibling gets included in the current rebuild MR. Renaming to include_in_rebuild makes that explicit. The field description now encodes both required conditions (dependency rebuild type AND fix shipped), and step 2 of the prompt exits early with include_in_rebuild=false for issues that are not dependency rebuilds at all, removing the ambiguity that caused the reasoning loop.

Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/ymir/agents/rebuild_consolidation.py
+++ b/ymir/agents/rebuild_consolidation.py
@@ -51,9 +51,11 @@ def build_rebuild_siblings_jql(
 class SiblingRebuildAnalysis(BaseModel):
     """LLM output schema for analyzing whether a sibling issue is a dependency rebuild."""
 
-    is_dependency_rebuild: bool = Field(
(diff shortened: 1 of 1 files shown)


TomasKorbar (approved):
LGTM

nforro (approved):

nforro:
> /agentic_review

What happened to Qodo?

lbarcziova:
> What happened to Qodo? 

no idea :(

PR #773, opened 2026-08-24
Add SKIP_MODULAR flag to enable modular issue triage

fila43:
## Summary

- Adds `SKIP_MODULAR` env var to gate the fetcher's modular-component filter. Code defaults to `"true"` (skip modular, safe for local runs); the cluster ConfigMap sets `"false"` to enqueue modular issues for triage.
- Logs the resolved `SKIP_MODULAR` value on every startup for easy cluster vs local debugging.
- Documents the new knob in the ConfigMap (with rationale comment) and in `openshift/README.md`.
- Fixes modular `downstream_component` parsing: `customfield_10669` stores `module:stream/package` (e.g. `postgresql:16/postgis`) but `is_modular()` / `parse_module_stream()` expect just the package name. Adds `extract_downstream_package()` utility to normalize the value so modular issues get the correct stream-specific branch (e.g. `stream-postgresql-16-rhel-9.8.0`) instead of falling through to `rhel-9.8.0`.

Resolves: PACKIT-5260

Depends on [#756](https://github.com/packit/ai-workflows/pull/756) for stream-aware duplicate detection — without it, modular issues with different streams would be incorrectly flagged as duplicates.

## Test plan

- [x] Existing `test_push_issues_to_queue_skip_modular` still passes (default skip behavior)
- [x] New `test_push_issues_to_queue_enqueue_modular` verifies `SKIP_MODULAR=false` enqueues modular issues
- [x] `test_init` asserts `skip_modular is True` by default
- [x] `test_extract_downstream_package` covers modular, non-modular, None, and empty inputs
- [x] `test_is_modular_requires_package_not_full_modular_string` proves the raw value fails, extracted value works
- [x] `test_map_version_to_module_branch_extracts_package_from_raw_field` end-to-end proof
- [ ] Deploy to staging with `SKIP_MODULAR=false` and verify modular issues get correct stream branches

### DIFF
--- a/openshift/configmap-jira-issue-fetcher-env.yml
+++ b/openshift/configmap-jira-issue-fetcher-env.yml
@@ -7,6 +7,12 @@ data:
   # jira_issue_fetcher.py. Shared by both cronjobs (daily + todo) since it's a
   # property of agent task duration, not of which query triggered the sweep.
   STALE_LABEL_THRESHOLD_HOURS: "24"
+  # Controls the fetcher-level filter for modular issues (Downstream Component
+  # matching module:stream/pkg, e.g. postgresql:16/postgresql). When "true"
+  # (default), the fetcher skips these issues before they reach the triage
+  # queue. When "false", modular issues pass through and are enqueued for
+  # triage like any other issue.
+  SKIP_MODULAR: "false"
 immutable: false
 kind: ConfigMap
 metadata:
--- a/openshift/README.md
+++ b/openshift/README.md
@@ -126,7 +126,7 @@ Two CronJobs run the fetcher with different JQL queries:
 | `jira-issue-fetcher` | `0 8 * * *` (daily, 8am UTC) | Main CVE batch — processes up to `MAX_ISSUES` issues from the filter in `jira-issue-fetcher-filter-env` | `jira-issue-fetcher-filter-env` |
 | `jira-issue-fetcher-todo` | `*/5 * * * *` | `labels = "ymir_todo"` OR consolidation labels (`ymir_consolidate_base`, `ymir_consolidate_next`) | `jira-issue-fetcher-todo-env` |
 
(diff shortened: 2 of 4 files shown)


openshift/configmap-jira-issue-fetcher-env.yml
```diff
@@ -7,6 +7,10 @@ data:
   # jira_issue_fetcher.py. Shared by both cronjobs (daily + todo) since it's a
   # property of agent task duration, not of which query triggered the sweep.
   STALE_LABEL_THRESHOLD_HOURS: "24"
+  # When "false", modular issues (Downstream Component matching module:stream/pkg)
+  # are enqueued for triage instead of being silently dropped. Code default is
+  # "true" (skip) so local runs without this ConfigMap preserve the old behavior.
```
lbarcziova: can you elaborate on this please? Not sure I follow, iiuc the filtering happens only on the fetcher level

lbarcziova (dismissed):
this LGTM, let's merge it after #756

openshift/configmap-jira-issue-fetcher-env.yml
fila43: Yes, the filtering happens on the fetcher level, but this setting controls whether modular issues are skipped there or passed further to the triage workflow. When it is set to true, modular issues are filtered out by the fetcher. When set to false, they are no longer skipped and can be picked up for triage. 

I can rephrase the comment to make this more precise.

fila43: @lbarcziova I hope the wording is better now

lbarcziova: looks good, thanks! Can you please rebase, and after I will merge this

lbarcziova (approved):

openshift/configmap-jira-issue-fetcher-env.yml
fila43: rebased

PR #775, opened 2026-08-25
Extend error context refactoring to remaining privileged tools

DaliborKr:
## Summary

Follow-up to #727 that extends the `tool_error_context` approach to the remaining privileged tools (`copr`, `gitlab`, `lookaside`) and enhances the base infrastructure.

- Extend `tool_error_context()` with `include_exception_message_for` parameter for selective exception message forwarding to the LLM
- Add `additional_context` arguments to existing `tool_error_context()` calls in tools already refactored in #727
- Replace manual `try/except ToolError` patterns with `tool_error_context` in `copr`, `gitlab` and `lookaside` tools

---
### Feedback welcome on `include_exception_message_for`:

This PR adds an optional `include_exception_message_for` parameter to `tool_error_context()`. It allows selectively appending exception messages of specified types to the LLM-facing `error_message`, while all other exceptions remain hidden behind the generic message (with details going only to observability).

Here are two useful example scenarios from the code:

**1. `ymir/tools/privileged/gitlab.py`**
- Forwarding intentional `ToolError` messages from within a wrapped block. For example, in `ForkRepositoryTool`, multiple validation checks raise `ToolError` with specific messages the LLM should see:

```python
with tool_error_context(
    "Failed to fork repository",
    include_exception_message_for=(ToolError,),
    repository=repository,
):
    project = await asyncio.to_thread(get_project, ...)
    if not project:
        raise ToolError("Failed to get the specified repository")
    if urlparse(project.service.instance_url).hostname != "gitlab.com":
        raise ToolError("Unexpected git forge, expected gitlab.com/redhat")
```

- The LLM sees `"Failed to fork repository: Unexpected git forge, expected gitlab.com/redhat"` instead of just `"Failed to fork repository"`, while unexpected exceptions (e.g. from the `get_project` call) still produce only the generic message.

**2. `ymir/tools/privileged/copr.py`**
- Surfacing specific error details constructed within the tool logic. For example, in `DownloadArtifactsTool`, a `ValueError` is raised with HTTP status info that helps the LLM understand what went wrong:

```python
with tool_error_context(
    "Failed to download build artifact",
    include_exception_message_for=(ValueError,),
    artifacts_url=url,
):
    async with aiohttp_get_with_retries(session, url) as response:
        if response.status < 400:
            # ... process content ...
        else:
            raise ValueError(f"{response.status} {response.reason}")
```

- The LLM sees `"Failed to download build artifact: 404 Not Found"`. Again, the status detail is useful context, while any other unexpected exception stays hidden behind the generic message.

I realize this adds complexity to what is otherwise a straightforward wrapper. The alternative would be catching and re-raising as `ToolErrorWithContext` manually at each call site, but that defeats the purpose of the context manager. If you see a cleaner approach, I'm open to suggestions.


ymir/tools/base.py
```diff
@@ -15,19 +15,49 @@
...
+            which passes through unchanged) and adds provided additional context
+            for observability.
+    """
     try:
         yield
     except ToolErrorWithContext:
         raise
     except Exception as e:
+        if isinstance(e, include_exception_message_for):
+            error_message = f"{error_message}: {e}"
```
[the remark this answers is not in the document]
DaliborKr: Right, I have wrapped the exception message with `redact_credentials()`.

nforro:
> **2. `ymir/tools/privileged/copr.py`**
> 
> * Surfacing specific error details constructed within the tool logic. For example, in `DownloadArtifactsTool`, a `ValueError` is raised with HTTP status info that helps the LLM understand what went wrong:
> 
> ```python
> with tool_error_context(
>     "Failed to download build artifact",
>     include_exception_message_for=(ValueError,),
>     artifacts_url=url,
> ):
>     async with aiohttp_get_with_retries(session, url) as response:
>         if response.status < 400:
>             # ... process content ...
>         else:
>             raise ValueError(f"{response.status} {response.reason}")
> ```
> 
> * The LLM sees `"Failed to download build artifact: 404 Not Found"`. Again, the status detail is useful context, while any other unexpected exception stays hidden behind the generic message.

I think this second example demonstrates the downside of this approach - `ValueError` is generic enough to be raised from standard library or third-party functions, where you don't control the message. It's probably mostly harmless, but I think it's possible to come up with a better solution.

DaliborKr:
@nforro 
> 
> I think this second example demonstrates the downside of this approach - `ValueError` is generic enough to be raised from standard library or third-party functions, where you don't control the message. It's probably mostly harmless, but I think it's possible to come up with a better solution.

Agree, I hadn't considered this risk. I have reviewed the current usage, and I realized that `include_exception_message_for` is only used for `ValueError` and `ToolError` exception types. 

I can address this by replacing the `include_exception_message_for` parameter with a simpler boolean like `forward_tool_errors`. This would automatically forward only `ToolError` messages when set to `True` (only when necessary). This should be safe since `ToolError` is unlikely to be raised by third-party libraries, right?

For the `ValueError` cases, I will need to find a safer approach to propagate those messages to the LLM without the scoping issues you mentioned.

nforro:
> This should be safe since `ToolError` is unlikely to be raised by third-party libraries, right?

Yes, but is there a reason not to include all `ToolError` messages?

Perhaps it would be best to be explicit and introduce e.g. a `DetailedToolError` that would be always included, while any other exception wouldn't?

DaliborKr:
@nforro 
> Yes, but is there a reason not to include all ToolError messages?
>
>Perhaps it would be best to be explicit and introduce e.g. a DetailedToolError that would be always included, while any other exception wouldn't?

Yes, the main reason is that in many cases the `error_message` parameter and the `ToolError` message are similar, so forwarding the `ToolError` would not add any meaningful information to the LLM. That is why I tried to forward it only in specific cases where the exception message provides details beyond what the generic error already says.

But you're right. Your suggestion of `DetailedToolError` would make this intent explicit at the raise site and make the `tool_error_context` wrapper straightforward to use by removing the optional parameter entirely.

ymir/tools/privileged/zstream_search.py
```diff
@@ -263,6 +265,8 @@ def _not_found():
         with tool_error_context(
             f"Failed to search Jira for related z-stream issues"
             f" for {tool_input.component}/{tool_input.fix_version}",
+            component=tool_input.component,
+            fix_version=tool_input.fix_version,
```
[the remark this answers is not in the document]
DaliborKr: The pass through is intentional. When `SearchJiraIssuesTool` fails, its own `ToolErrorWithContext` already carries the root cause and its context. The outer `component` and `fix_version` are caller context, not diagnostic info about what went wrong. They are only reached on the generic Exception path, which is correct.

ymir/tools/privileged/copr.py
```diff
@@ -155,28 +154,43 @@ async def _run(
             "description": f"Test builds for {jira_issue}",
             "delete_after_days": COPR_PROJECT_LIFETIME,
         }
-        try:
+        with tool_error_context(
+            "Failed to create or update Copr project",
+            copr_user=copr_user,
+            project=jira_issue,
```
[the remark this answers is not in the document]
DaliborKr: Valid point, I have passed the Copr error details (`_copr_error_detail`) to the observability context. The LLM-facing message stays generic.

ymir/tools/privileged/gitlab.py
```diff
@@ -570,20 +586,22 @@ async def _run(
...
             if not project:
                 raise ToolError(f"Failed to get repository for package: {package}")
 
             branches = await asyncio.to_thread(project.get_branches)
-            logger.info(f"Found {len(branches)} branches for package {package}: {branches}")
-            return JSONToolOutput(result=branches)
 
-        except OgrException as ex:
-            logger.warning(f"Failed to get branches for package {package}: {ex}")
-            raise ToolError(f"Failed to get branches for package {package}: {ex}") from ex
```
jpodivin, line 587: Why are we dropping handling of `OrgException` without including it in `include_exception_message_for`?

ymir/tools/privileged/copr.py
```diff
@@ -124,15 +125,11 @@ async def _run(
         srpm_path = tool_input.srpm_path
         dist_git_branch = tool_input.dist_git_branch
         jira_issue = tool_input.jira_issue
-        try:
+        with tool_error_context("Failed to initialize Kerberos ticket"):
```
jpodivin, line 121: Shouldn't `KerberosError` messages be included as well?

ymir/tools/privileged/testing_farm.py
```diff
@@ -923,43 +941,49 @@ async def _run(
...
+                    *ssh_opts,
+                    "-r",
+                    *local_paths,
+                    f"{ssh_host}:{remote_dir}",
+                    stdout=asyncio.subprocess.PIPE,
+                    stderr=asyncio.subprocess.PIPE,
+                )
+                _, stderr = await asyncio.wait_for(active_proc.communicate(), timeout=timeout)
+                if active_proc.returncode != 0:
+                    raise RuntimeError(f"SCP failed: {stderr.decode().strip()}")
```
TomasKorbar: Should we not propagate these RuntimeErrors as well? Perhaps agent can benefit from them with faster recovery if it mistakenly attempts to create directory structure on path where file is located.

TomasKorbar (changes requested):

TomasKorbar:
I submitted one comment to thing that stood out a bit to me. Otherwise good.

opohorel:
overall I like the idea. there is one finding from Codex, which pointed out an issue we already have in the repo:

Codex identified a potential information leak in tool_error_context while validating this PR’s selective-forwarding contract. The underlying behavior predates this PR, but this change expands its use across additional privileged tools and now documents that unexpected exception details remain hidden from the LLM.

Passing the original exception as cause=e causes BeeAI 0.1.83’s error.explain() to render the raw cause chain. Although ReasoningAgent removes the redacted additional_context before formatting the tool error, it does not remove the cause. The resulting text is then sent to the LLM in a ToolMessage.

This was reproduced with a credential that appeared as [REDACTED] in additional_context but remained visible in the rendered RuntimeError cause. Consequently, the distinction introduced by include_exception_message_for is ineffective: exceptions not selected for forwarding can still expose their complete messages through the cause chain.

Please provide an LLM-safe error formatter that excludes both observability context and the raw cause, and add a regression test covering the actual ReasoningAgent formatting path.

DaliborKr:
@opohorel 

> Please provide an LLM-safe error formatter that excludes both observability context and the raw cause

Thanks, this was the core issue, and I reworked the whole approach around it.

- Added an LLM-safe formatter `explain_to_llm()`. It renders only the `FrameworkError` messages in the chain + the `name` context key (the allowed context key list can be extended in the future via `_LLM_SAFE_CONTEXT_KEYS`), and drops the raw cause, stack trace, and additional_context. So the leak through the cause chain is gone.
- `ReasoningAgent` now uses `explain_to_llm()` instead of the original `error.explain()` when building the tool message for the LLM on error.
- Removed `include_exception_message_for` completely, since, as you pointed out, the distinction was ineffective. Now the rule is simple: only messages explicitly raised as `FrameworkError` instances (`ToolError`, `ToolErrorWithContext`) reach the LLM. These errors are raised exclusively in our code (and potentially in BeeAI's code). Any other exception is kept as the cause for observability only.

This also resolves the conversation I had with @nforro regarding the `include_exception_message_for` parameter earlier in this thread, as there is no need to select specific exception types to be provided to the LLM anymore, and it is enough just to raise a `FrameworkError` instance within a tool.

There is a new separate commit with all of these changes so you can check it out.

ymir/tools/privileged/gitlab.py
DaliborKr: You are right, I was too strict here. The `OgrException` message is beneficial for LLM. Since, I dropped `include_exception_message_for` completely and now the new `explain_to_llm()` only renders `FrameworkError` messages, the raw `OgrException` would lose its message. So I re-raise it as a `ToolError` (redacted) so it reaches the LLM. 

https://github.com/packit/ai-workflows/pull/775/changes/f4910b14131f5757bb6fa4351c42b1f4f8fddd94#diff-2692556a0912053b5dffc50c955e51f1893b9bb944968e563f433fb1be7c3e10R601-R604

ymir/tools/privileged/copr.py
DaliborKr: I thought, that at this point, when the tool reaches the `KerberosError`, there is nothing the agent can act on, since it has nothing to do about wrong Kerberos configuration. Therefore I kept only the generic message for the LLM. Full exception is still in the `cause` and `additional_context`, available in logs and observability.

Correct me if, I am wrong.

ymir/tools/privileged/testing_farm.py
DaliborKr: Agreed, this one is useful for the agent. Because of my latest changes, a plain `RuntimeError` would not be forwarded to LLM as is (see my response to @opohorel's review), so I changed both to `ToolError` (redacted) so the stderr reaches the LLM.

ymir/tools/privileged/distgit.py
```diff
@@ -210,7 +208,11 @@ async def _run(
                     result=f"Z-Stream branch {branch} already exists, no need to create it"
                 )
         with (
-            tool_error_context("Failed to create Z-Stream branch", package=package, branch=branch),
+            tool_error_context(
+                "Failed to create Z-Stream branch",
+                package=package,
+                branch=branch,
+            ),
```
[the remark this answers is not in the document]
DaliborKr: Right, I can add `branch` information to the inner context as well.

nforro (commented):
Nice! There is one issue though:

**ymir/tools/base.py:39** catches BeeAI's `AbortError` and converts it to a retryable `ToolErrorWithContext`. `AbortError` inherits both `FrameworkError` and `asyncio.CancelledError`, but its MRO places `Exception` first, so `except Exception` catches it. This affects all newly wrapped operations and turns cancellation/timeouts into normal tool failures, allowing the agent to continue or retry after cancellation. Add `except asyncio.CancelledError: raise` before the broad handler and test with BeeAI `AbortError`.

DaliborKr:
@nforro 
> This affects all newly wrapped operations and turns cancellation/timeouts into normal tool failures, allowing the agent to continue or retry after cancellation

I tested this case and you were right. The solution you proposed solves the problem, so it has been fixed.

nforro (approved):

TomasKorbar (approved):

PR #785, opened 2026-08-26
Fix: Include all terminal labels when checking sibling completion status

majamassarini:
## Problem

Primary issues stuck waiting for siblings that finished with `ymir_backported`, `ymir_backport_errored`, etc. These weren't recognized as terminal states.

**Example:** RHEL-248139 - all 7 siblings finished but primary remained stuck with `ymir_rebase_waiting_for_siblings`.

## Root Causes Fixed

 **Missing terminal labels** - Only checked `ymir_triaged_*`, not completion/error labels

Fixes: https://redhat.atlassian.net/browse/RHEL-248139


lbarcziova (approved):
LGTM

PR #787, opened 2026-08-27
Fetch issues by queue state rather than in daily batches

nforro:
Maximum triage queue depth is controlled by `QUEUE_DEPTH_THRESHOLD` (currently 15). If the current queue depth is lower than the threshold, a cron job running every 15 minutes pushes issues from the `CVE work to do` filter up until the threshold. `ymir_todo` issues are unaffected by this and the `triage_queue_todo` queue is limited only by `MAX_ISSUES` (now lowered to 20).

Fixes https://redhat.atlassian.net/browse/PACKIT-5273.

### DIFF
--- a/openshift/configmap-jira-issue-fetcher-env.yml
+++ b/openshift/configmap-jira-issue-fetcher-env.yml
@@ -1,11 +1,16 @@
 apiVersion: v1
 data:
-  MAX_ISSUES: "100"
+  # Only caps jira-issue-fetcher-todo's per-cycle push now — it ignores
+  # QUEUE_DEPTH_THRESHOLD (set in jira-issue-fetcher-filter-env), so this is
+  # its only throttle. jira-issue-fetcher instead derives its per-cycle cap
+  # from QUEUE_DEPTH_THRESHOLD minus current queue depth.
+  MAX_ISSUES: "20"
   LOGLEVEL: "INFO"
   # Conservative default pending real task duration data from Phoenix traces;
   # see the comment next to stale_label_threshold_hours in
-  # jira_issue_fetcher.py. Shared by both cronjobs (daily + todo) since it's a
-  # property of agent task duration, not of which query triggered the sweep.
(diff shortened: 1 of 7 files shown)


openshift/configmap-jira-issue-fetcher-env.yml
```diff
@@ -1,11 +1,12 @@
 apiVersion: v1
 data:
-  MAX_ISSUES: "100"
+  MAX_ISSUES: "20"
```
lbarcziova, line 7: I am a bit puzzled on this, could you add a short comment explaining how does it go together with QUEUE_DEPTH_THRESHOLD?

openshift/cronjob-jira-issue-fetcher.yml
```diff
@@ -6,7 +6,7 @@ metadata:
     app: jira-issue-fetcher
     component: scheduler
 spec:
-  schedule: "0 8 * * *"  # Daily at 8am UTC — automated CVE batch
+  schedule: "*/5 * * * *"  # Every 5 minutes — tops up the triage queue toward QUEUE_DEPTH_THRESHOLD
```
lbarcziova: what about starting with e.g. 15 min here and adjusting later if needed? It reads the whole filter every run, and I'm not sure processing is fast enough yet to need 5 min freshness.

lbarcziova (commented):
just 2 notes, otherwise looks great!

lbarcziova (dismissed):
this LGTM, thanks! 

I am just thinking how this will work together with #746 , which pushes directly to triage queue, cc @jpodivin

nforro:
> I am just thinking how this will work together with #746 , which pushes directly to triage queue, cc @jpodivin

It should work fine, just the issues pushed from sweeps will have priority over regular issues.

nforro:
Forgot to update the OpenShift README for 15 minutes interval.

majamassarini (approved):

lbarcziova (approved):

PR #792, opened 2026-09-01
Fix: Triage Y-stream CVEs when Z-stream clones are NOT_AFFECTED

majamassarini:
## Summary

Fixes [PACKIT-5281](https://redhat.atlassian.net/browse/PACKIT-5281): Y-stream CVEs were incorrectly skipped or postponed when Z-stream clones were NOT_AFFECTED.

## Problem

When a Y-stream CVE (e.g., rhel-9.9) is checked for eligibility:
- **Low/Moderate severity + CentOS Stream first approach** → told "fix is handled via Z-stream CentOS path"
- **Important/Critical severity** → told "waiting for Z-stream to ship"

Both messages were incorrect when the CVE was actually **NOT AFFECTED** in the component. The Y-stream should have been triaged to confirm it's also not affected, instead of being skipped or postponed.

**Impact:** Maintainers had to manually close Y-stream issues that should have been automatically triaged.

## Solution

Check Z-stream triage status before skipping or postponing Y-stream CVEs.

### Changes

1. **Added `_check_zstream_not_affected()`**
   - Searches for Z-stream clones with `ymir_triaged_not_affected` label
   - Includes fixVersion filter to prevent pagination issues
   
2. **Added `_check_zstream_pending_triage()`**
   - Searches for Z-stream clones without terminal success labels (backported/rebased/rebuilt)
   - Treats failed/errored actions as still pending (Z-stream path is blocked)
   - Includes fixVersion filter to prevent pagination issues
   
3. **Modified `_check_lowmod_ystream_eligibility()`**
   - For Low/Moderate Y-stream CVEs with CS_FIRST approach:
     - ✅ If Z-stream is NOT_AFFECTED → return `IMMEDIATELY` (triage Y-stream)
     - ⏳ If Z-stream pending triage → return `PENDING_DEPENDENCIES` (wait for results)
     - ❌ Otherwise → return `NEVER` (skip Y-stream - existing behavior)
   
4. **Modified `_check_for_dependency_blocker()`**
   - For Important/Critical Y-stream CVEs:
     - ✅ If Z-stream is NOT_AFFECTED → return `None` with specific reason (proceed with triage)
     - ⏳ Otherwise → return `PENDING_DEPENDENCIES` (postpone - existing behavior)

## Code Review Fixes

After code review by Claude, fixed seven additional issues:

1. **Pending-triage logic**: Only exclude SUCCESS labels (ymir_backported/rebased/rebuilt) from pending search. Failed/errored labels are NOT excluded because the Z-stream path is blocked and Y-stream might be needed as fallback.

2. **Shell redirection fix**: Quoted pip requirements with `>=` operators in three Containerfiles (supervisor, c10s, c9s) to prevent shell redirection interpretation. Unquoted `pkg>=1.0` is parsed as "run pkg, redirect output to file =1.0".

3. **NOT_AFFECTED-specific reason**: Return explicit message "Z-stream clone RHEL-XXX was NOT_AFFECTED, checking if Y-stream is also not affected" instead of generic message.

4. **Pagination limit fix**: Added fixVersion filter directly to JQL queries. SearchJiraIssuesTool returns only first 50 results with no pagination. Without version filtering, broad CVE/component queries could hit the limit and miss applicable Z-stream clones.

5. **Duplicate preservation fix**: The NOT_AFFECTED result path was dropping the `duplicate_of` field even though duplicate detection already ran. The triage workflow needs this field to notify maintainers about older rejected trackers. Fixed by passing `duplicate_of` through `_check_for_dependency_blocker()` and including it in the NOT_AFFECTED result. Added regression test combining rejected duplicate with NOT_AFFECTED Z-stream.

6. **Postponed labels fix**: `_check_zstream_pending_triage()` excluded deprecated `ymir_triaged_postponed` but not the actual labels applied by current code (`ymir_postponed_dependency`, `ymir_postponed_no_patch`, `ymir_postponed_pr_pending`). These are terminal triage decisions (Z-stream was triaged and postponed with a reason), but without exclusion they match "pending triage" JQL. Y-stream then waits indefinitely for Z-stream results that already exist. Fixed by excluding the three active postponed labels. Added parameterized test verifying all three are excluded.

7. **Makefile consistency fix**: The `triage-issue` and `process` targets used `$(ISSUE)` instead of `$(JIRA_ISSUE)`, inconsistent with all other targets (process-issue, trigger-pipeline, trigger-reproducer). Users running `make triage-issue JIRA_ISSUE=RHEL-123` got silent failures inside the container. Fixed by changing `$(ISSUE)` to `$(JIRA_ISSUE)` and adding guards to fail fast with usage message (like trigger-pipeline has).

## Related

- Implements the strategy suggested in PACKIT-5281: "triage all Z streams first. The Y-streams should then pick up findings from latest Z"

### DIFF
--- a/Containerfile.c10s
+++ b/Containerfile.c10s
@@ -84,6 +84,7 @@ RUN pip3 install --no-cache-dir \
       unidiff \
       PyYAML>=5.1 \
       sentry-sdk>=2.13.0 \
+      typer \
     && cd /usr/local/lib/python3.12/site-packages \
     && patch -p5 -i /tmp/openinference-reasoning.patch \
     && patch -p5 -i /tmp/openinference-streaming.patch
--- a/Containerfile.c10s
+++ b/Containerfile.c10s
@@ -80,10 +80,10 @@ RUN pip3 install --no-cache-dir \
       specfile \
       pytest \
       pytest-asyncio \
-      GitPython>=3.1.0 \
+      "GitPython>=3.1.0" \
       unidiff \
-      PyYAML>=5.1 \
-      sentry-sdk>=2.13.0 \
+      "PyYAML>=5.1" \
+      "sentry-sdk>=2.13.0" \
       typer \
     && cd /usr/local/lib/python3.12/site-packages \
     && patch -p5 -i /tmp/openinference-reasoning.patch \
--- a/Containerfile.c9s
+++ b/Containerfile.c9s
@@ -83,10 +83,10 @@ RUN python3.11 -m venv --system-site-packages /opt/beeai-venv \
          redis \
          specfile \
          koji \
-         GitPython>=3.1.0 \
+         "GitPython>=3.1.0" \
          unidiff \
-         PyYAML>=5.1 \
-         sentry-sdk>=2.13.0 \
+         "PyYAML>=5.1" \
+         "sentry-sdk>=2.13.0" \
          typer \
     && cd /opt/beeai-venv/lib/python3.11/site-packages \
     && patch -p5 -i /tmp/openinference-reasoning.patch \
--- a/Containerfile.c9s
+++ b/Containerfile.c9s
@@ -87,6 +87,7 @@ RUN python3.11 -m venv --system-site-packages /opt/beeai-venv \
          unidiff \
          PyYAML>=5.1 \
          sentry-sdk>=2.13.0 \
+         typer \
     && cd /opt/beeai-venv/lib/python3.11/site-packages \
     && patch -p5 -i /tmp/openinference-reasoning.patch \
     && patch -p5 -i /tmp/openinference-streaming.patch
--- a/Containerfile.supervisor
+++ b/Containerfile.supervisor
@@ -34,8 +34,8 @@ RUN dnf -y install --allowerasing \
       arize-phoenix-otel \
       redis \
       specfile \
-      sentry-sdk>=2.13.0 \
-      GitPython>=3.1.0 \
+      "sentry-sdk>=2.13.0" \
+      "GitPython>=3.1.0" \
     && dnf -y remove gcc gcc-c++ python3-devel \
     && dnf clean all
 
--- a/Containerfile.supervisor
+++ b/Containerfile.supervisor
@@ -10,6 +10,7 @@ RUN update-ca-trust
 
 RUN dnf -y install --allowerasing \
       # Run time dependencies
+      git \
       krb5-workstation \
       python3 \
       python3-backoff \
@@ -33,6 +34,8 @@ RUN dnf -y install --allowerasing \
       arize-phoenix-otel \
       redis \
       specfile \
+      sentry-sdk>=2.13.0 \
+      GitPython>=3.1.0 \
     && dnf -y remove gcc gcc-c++ python3-devel \
     && dnf clean all
 
--- a/Makefile
+++ b/Makefile
@@ -430,6 +430,22 @@ supervisor-collect:
 	$(COMPOSE_SUPERVISOR) run --rm \
 		supervisor python -m ymir.supervisor.main $(DEBUG_FLAG) collect --no-repeat
 
+.PHONY: triage-issue
+triage-issue:
+	$(COMPOSE_AGENTS) run --rm \
+		-e JIRA_ISSUE=$(ISSUE) \
+		-e DRY_RUN=$(DRY_RUN) \
+		-e AUTO_CHAIN=false \
+		triage-agent
+
+.PHONY: process
+process:
+	$(COMPOSE_AGENTS) run --rm \
+		-e JIRA_ISSUE=$(ISSUE) \
+		-e DRY_RUN=$(DRY_RUN) \
+		-e AUTO_CHAIN=true \
+		triage-agent
+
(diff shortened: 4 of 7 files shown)


nforro:
One general note that's been bothering me for some time. Could you please follow the [contributing guidelines](https://github.com/packit/contributing#git-workflow) and, to quote, use common sense when creating commits, not too big, not too small? I think in this case 2 or 3 commits would be enough. It would make the PR(s) easier to review and the git history cleaner.

nforro (commented):
A couple of things Claude flagged:

## Finding 1 (high confidence, real bug): `_check_zstream_pending_triage` misses the *current* postponed-triage labels

`ymir/tools/privileged/jira.py`, new function `_check_zstream_pending_triage`, excludes these labels as "terminal/already-triaged":

```
ymir_triaged_backport, ymir_triaged_rebase, ymir_triaged_rebuild,
ymir_triaged_postponed, ymir_triaged_not_affected, ymir_triaged,
ymir_backported, ymir_rebased, ymir_rebuilt,
ymir_needs_attention, ymir_triage_errored
```

`ymir_triaged_postponed` is the label the comment relies on for "postponed" — but per `ymir/common/constants.py:202-204`, that label is **deprecated** ("replaced by labels with reason for postponement") and is only ever referenced by `label_postponed_issues` in `triage_agent.py` for *removal* of legacy labels, never applied by current code. The actual labels a postponed triage decision gets today come from `_RESOLUTION_TO_LABEL` (`triage_agent.py:141-152`):

```
Resolution.POSTPONED_DEPENDENCY → ymir_postponed_dependency
Resolution.POSTPONED_NO_PATCH   → ymir_postponed_no_patch
Resolution.POSTPONED_PR_PENDING → ymir_postponed_pr_pending
```

None of these are excluded. So a Z-stream clone that **has already been triaged** and landed on "postponed — waiting on a dependency/patch/PR" (a real, terminal triage decision, just like backport/rebase/rebuild) still matches the "pending triage" JQL. The Y-stream CVE then gets `PENDING_DEPENDENCIES`, "waiting for Z-stream clone triage results" — results that already exist — and will wait indefinitely, since nothing in the normal sweep flow adds one of the excluded labels to resolve that state. This reproduces the same class of bug PACKIT-5281 was filed to fix, just via a different path.

Fix: also exclude `ymir_postponed_dependency`, `ymir_postponed_no_patch`, `ymir_postponed_pr_pending` (no need for `ymir_postponed_y_stream` — that's Y-stream-only).

## Finding 2 (medium confidence): new Makefile targets break the `JIRA_ISSUE` convention

`Makefile`'s new `triage-issue` and `process` targets read `$(ISSUE)`:

```makefile
triage-issue:
	$(COMPOSE_AGENTS) run --rm -e JIRA_ISSUE=$(ISSUE) ...
process:
	$(COMPOSE_AGENTS) run --rm -e JIRA_ISSUE=$(ISSUE) ...
```

Every other target in the file (8+ call sites, including the pre-existing `process-issue`, `trigger-pipeline`, `trigger-reproducer`) takes `JIRA_ISSUE=` from the invoker. A user running `make triage-issue JIRA_ISSUE=RHEL-123` out of habit gets an empty `$(ISSUE)` with no error — `trigger-pipeline`/`trigger-reproducer` both guard with `@if [ -z "$(JIRA_ISSUE)" ]; then ... exit 1; fi`, but the two new targets have no such guard, so this fails silently inside the container instead of at the `make` invocation.

## Finding 3 (lower confidence, pre-existing but adjacent): unquoted `>=` in Containerfile.c10s / Containerfile.c9s

This PR's own fix commit quotes `sentry-sdk>=2.13.0` and `GitPython>=3.1.0` in `Containerfile.supervisor` specifically because unquoted `>=` is parsed by the shell as output redirection (`pkg>=1.0` → runs `pkg`, writes a file named `=1.0`). Verified empirically. The exact same pattern already exists two lines above the `typer` line this PR adds in `Containerfile.c10s`/`Containerfile.c9s`:

```
PyYAML>=5.1 \
sentry-sdk>=2.13.0 \
+typer \
```

These are pre-existing (not introduced by this PR), so lower priority, but since the PR is touching this exact list and just fixed the identical bug in a sibling file, it's worth a follow-up: both version pins are silently being dropped and a stray `=5.1`/`=2.13.0` file is left in the build context.

majamassarini:
> not too big, not too small? I think in this case 2 or 3 commits would be enough.

I am always in doubt here, I used to squash commits in the past, because describing them "manually" was too much work and my english is not so good, however now that AI writes commit messages for me, I don't squash them any longer, because I think it is easier, in this way, to review things and decide if something completely wrong has been done. However I squashed them here.

nforro:
> However I squashed them here.

Thanks. Though the last two fixup commits should be squashed as well. I'm completely fine with doing that just before merging the PR though, if it makes reviews easier. Although that probably requires re-approvals.

nforro (approved):
LGTM

majamassarini:
> Thanks. Though the last two fixup commits should be squashed as well. 

They can, however I will keep them separate. One is related with the rebase and the changes driven by the work done on the main branch. The other one addresses the reviews. So I think they can stay separate.

PR #794, opened 2026-09-02
Revamp error_list

nforro:
When a task ends up in error_list, additional metadata are now stored alongside the error details and Jira key - unique ID, timestamp and the original trigger, so it is possible to re-enqueue a task to its original queue in case the error has been analyzed and fixed (or it was transient).

`make show-error-list` now shows error IDs and timestamps and `make requeue-error ERROR_ID=N` moves an errored task from error_list to its original queue and resets the retry counter.

### DIFF
--- a/openshift/Makefile
+++ b/openshift/Makefile
@@ -80,6 +80,16 @@ show-reproducer-queue:
 show-error-list:
 	@oc exec deployment/valkey -- valkey-cli --no-raw LRANGE error_list 0 -1 | python3 scripts/error_list.py --file -
 
+# Requeue a specific error_list entry (by the [id=...] shown in show-error-list)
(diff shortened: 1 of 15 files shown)


opohorel (dismissed):
Apart from the merge conflict, it looks good to me. Not sure that this approve is going to hold after the conflict is resolved, if not just ping me about it

nforro:
> Not sure that this approve is going to hold after the conflict is resolved, if not just ping me about it

Thanks, I'm going to address all Qodo's remarks first.

lbarcziova (approved):

opohorel (approved):

majamassarini (approved):
Nice!

TomasTomecek (approved):

PR #799, opened 2026-09-04
Capture exceptions raised by tool calls in y_stream sweep

jpodivin:


lbarcziova (approved):

opohorel:
From Codex:

• Review verdict: PR #799 is valid as a diagnostic patch, but it will not fix the sweep.
  - P1 root cause: the Y-stream sweep calls load_rhel_config(), but the CronJob manifest (openshift/cronjob-sweep-y-stream.yml:27) and
    Compose service (compose.yaml:459) never mount rhel-config.json. The image does not copy it either. Every Y-stream check therefore ends
    with FileNotFoundError; #799 merely sends that error to Sentry.
  - Tests miss this because CheckCveTriageEligibilityTool is fully mocked.
  - Minor inherited issue: quote "sentry-sdk>=2.13.0" in Containerfile.sweep:29. Unquoted > is interpreted as shell redirection.

PR #802, opened 2026-09-07
sweep: increase memory limits and correct the mount path

jpodivin:
With these changes the sweep job can finally run successfully.


opohorel (changes requested):
I believe you should also mount the rhel-config in `cronjob-sweep-dependency.yml` as you are loading the rhel-config there too

opohorel (approved):
LGTM, thanks!

lbarcziova (approved):

PR #806, opened 2026-09-08
New API

TomasKorbar:
- Add HTTP API server for submitting consolidation jobs
- Add Jira webhook command endpoint for bot-mention dispatching
- Setup OIDC authentication support for our api server
- Extend race prevention mechanism in consolidation job submission


nforro (commented):
It seems there are some missing pieces, here are some LLM findings:

1. **P1: Public unauthenticated job submission**

   `ymir/api/consolidation.py:80-97,146-148` exposes `/api/consolidation` without authentication through `openshift/route-api.yml`. Anyone reaching the route can enqueue costly arbitrary workflows. Additionally, webhook authentication fails open when `JIRA_WEBHOOK_SECRET` is unset (`ymir/api/jira_webhook.py:65-69`). Authentication should cover every mutating endpoint and fail closed.

2. **P1: API image is never published**

   `openshift/imagestream-api.yml:10` imports `quay.io/jotnar/ymir-api:latest`, but `.github/workflows/build-and-push.yml` has no job publishing `Containerfile.api`. After merge, deployment will receive no new API image.

3. **P1: Deployment requires an undocumented secret**

   `openshift/deployment-api.yml:50-51` references mandatory secret `api-env`, but `openshift/README.md:11-46` neither documents nor creates it. Unless provisioned out-of-band, the pod remains in `CreateContainerConfigError`.

4. **P2: None of the 30 API tests run in CI**

   `Makefile.tests:29-52` enumerates existing package test directories but omits `ymir/api/tests`. Current green checks therefore provide no test coverage for the new server.

5. **P2: `source_issues` accepts invalid cardinalities**

   `ymir/api/consolidation.py:31,109` accepts zero, one, or more than two issues, although consolidation operates on a pair. One issue queues a successful no-op; more than two populates all branches while `ymir/agents/prompts/mr_consolidation/prompt.j2:7-13` describes only the first two. Constrain this to exactly two entries when supplied.

TomasKorbar:
> It seems there are some missing pieces, here are some LLM findings:
> 
>     1. **P1: Public unauthenticated job submission**
>        `ymir/api/consolidation.py:80-97,146-148` exposes `/api/consolidation` without authentication through `openshift/route-api.yml`. Anyone reaching the route can enqueue costly arbitrary workflows. Additionally, webhook authentication fails open when `JIRA_WEBHOOK_SECRET` is unset (`ymir/api/jira_webhook.py:65-69`). Authentication should cover every mutating endpoint and fail closed.
> 
Yeah there is no authentication, but the service is accessible only from internal network and ymir_todo can be placed by essentially anyone, so i am not sure whether implementing authentication is necessary/useful here.
The fail on missing JIRA_WEBHOOK_SECRET makes sense.
>     2. **P1: API image is never published**
>        `openshift/imagestream-api.yml:10` imports `quay.io/jotnar/ymir-api:latest`, but `.github/workflows/build-and-push.yml` has no job publishing `Containerfile.api`. After merge, deployment will receive no new API image.
> 
ah missed this. Did not know where to put it. Will fix.
>     3. **P1: Deployment requires an undocumented secret**
>        `openshift/deployment-api.yml:50-51` references mandatory secret `api-env`, but `openshift/README.md:11-46` neither documents nor creates it. Unless provisioned out-of-band, the pod remains in `CreateContainerConfigError`.
> 
Ah will add info to documentation.
>     4. **P2: None of the 30 API tests run in CI**
>        `Makefile.tests:29-52` enumerates existing package test directories but omits `ymir/api/tests`. Current green checks therefore provide no test coverage for the new server.
> 
Dammit.
>     5. **P2: `source_issues` accepts invalid cardinalities**
>        `ymir/api/consolidation.py:31,109` accepts zero, one, or more than two issues, although consolidation operates on a pair. One issue queues a successful no-op; more than two populates all branches while `ymir/agents/prompts/mr_consolidation/prompt.j2:7-13` describes only the first two. Constrain this to exactly two entries when supplied.
Right. Makes sense.

nforro:
> Did not know where to put it. Will fix.

You also need to create the repo on quay.io.

lbarcziova:
thanks a lot for kickstarting this!

As for the current code, I have few findings with the help of Codex, worth double-checking:

  - /api/consolidation is exposed through a public OpenShift Route without authentication - the point already brought up by Nikola, I think adding the authentication doesn't hurt and if we want this to be right now triggered only via Jira, it seems like a more controlled approach (I think we do check for the group of the user who currently triggers via `ymir_todo`, for this we should also do similar user/group validation)
 - Jira webhook authentication uses the wrong protocol. Native Jira Cloud webhooks send an HMAC in X-Hub-Signature, calculated over the raw request body. The PR instead compares a raw X-Webhook-Secret value. Standard Jira Cloud deliveries will therefore not authenticate. See ymir/api/jira_webhook.py:26,64-75 and Atlassian’s webhook documentation (https://developer.atlassian.com/cloud/jira/platform/webhooks/#validating-webhook-deliveries).
  - Queue submission is still vulnerable to a race. The API checks for existing pending/active jobs and then performs a separate Redis submission. Two concurrent requests can both pass the check, with the later HSET overwriting the first source-specific job. See ymir/api/consolidation.py:53-70 and ymir/common/merge_queue.py:47-66.
  - Redis failures in the pre-check path are handled inconsistently. The source-specific HGETs happen outside the try block, so Redis failures bypass the intended JSON error response and propagate through the webhook handler differently from failures during submit_merge_job.
  - Readiness does not reflect Redis availability. /healthz always returns 200, and the same endpoint is used for readiness. If Redis becomes unavailable after startup, Kubernetes will continue routing
    requests to an API that cannot submit jobs.

---

  I also wanted to bring up the Ymir triggering rearchitecture research here too (https://github.com/packit/research/blob/00a812858e100da0d7fb124d3b83220de990af19/research/ymir-triggering-rearchitecture/index.md) - the approach here uses webhooks, but I remember we were seriously considering also Automation + event router, with manual trigger forms. This choice affects authentication, payload format, authorization, deduplication, and UX. This would be worth re-descussing on arch. Considering that, we could split this PR into the API foundation and then the followup integration, wdyt? With that, the API foundation could still be tested standalone.

The initial research also showed the current OpenShift route may not be reachable from Jira directly, we would need to consider alternatives for that case: simple Ymir dashboard, polling, ..?

openshift/README.md
```diff
@@ -138,6 +143,88 @@ Agents are deployed in the `jotnar-ymir--jotnar-ymir` project.
...
+  |  config.json            |    |  JWKS (jwcrypto)       |
+  +--------------------------+   +-------------------------+
+```
+
+### Browser usage (trace server UI)
+
+1. Visit the trace server at `https://ymir.redhat.com` (or `https://trace-server-jotnar-ymir--...`)
+2. Click "login" in the header — redirects to Red Hat SSO
+3. After login, navigate to "submit" to submit consolidation jobs
+4. The SPA manages tokens automatically (PKCE flow)
```
nforro, line 255: How can I query traces in JSON with curl?

nforro (commented):
Two points from GPT-5.6 Luna:

**P1: JWT issuer and audience are not validated**

   `ymir/api/auth.py:89-95` constructs `jwcrypto.jwt.JWT(key=keyset, jwt=token)` and returns its claims, but never checks `iss` or `aud` against the configured OIDC provider/API. This accepts a correctly signed, unexpired token with arbitrary issuer and audience claims. I reproduced this with a token containing `iss="wrong-issuer"` and `aud="other-client"`; `_validate_token()` accepted it. Validate the expected issuer and API audience before authorizing the request, and add rejection tests.

**P2: API tests are still absent from PR CI**

   `Makefile.tests:18-25,92-93` now defines `check-api-in-container`, but `.github/workflows/check-in-container.yml:31-39` has no `check-api-in-container` matrix entry. The current green PR checks therefore still do not execute `ymir/api/tests`; add the target to the workflow matrix.

openshift/README.md
TomasKorbar: To get traces you do not need to authenticate. But the consolidation job would be submitted as
```
curl -X POST https://api-jotnar-ymir--jotnar-ymir.apps.gpc.ocp-hub.prod.psi.redhat.com/api/consolidation \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"package": "bind", "target_branch": "rhel-9.6.0"}'
```

nforro: Sorry, I didn't realize the login in trace-server web UI is not for traces but for consolidation submission.

.github/workflows/build-and-push.yml
```diff
@@ -132,6 +132,20 @@ jobs:
...
+      - name: Build and push ymir-api to quay.io registry
+        uses: sclorg/build-and-push-action@adf1dee2b786ccbb2e2a708c3510a90e2ab3392d # v4.1.5
+        with:
+          registry: "quay.io"
+          registry_namespace: "jotnar"
+          registry_username: ${{ secrets.REGISTRY_LOGIN }}
+          registry_token: ${{ secrets.REGISTRY_TOKEN }}
+          dockerfile: "Containerfile.api"
+          docker_context: "."
+          image_name: "ymir-api"
```
lbarcziova, line 147: please create the repo in quay for this, the auto building won't work without it

ymir/api/jira_webhook.py
```diff
@@ -0,0 +1,141 @@
...
+    comment_body = (body.get("comment") or {}).get("body")
+    if not comment_body:
+        return web.json_response({"ignored": True, "reason": "empty comment body"})
+
+    bot_account_id = _get_bot_account_id()
+    if bot_account_id is None:
+        logger.warning("JIRA_BOT_ACCOUNT_ID not configured, ignoring webhook")
+        return web.json_response({"ignored": True, "reason": "bot account not configured"})
+
+    command_text = _extract_command(comment_body, bot_account_id)
```
lbarcziova, line 167: didn't we want to have this restricted based on account role?

ymir/api/consolidation.py
```diff
@@ -0,0 +1,152 @@
...
+    """Payload for the POST /api/consolidation endpoint."""
+
+    package: str
+    target_branch: str
+    source_issues: list[str] | None = None
+    release_strategy: Literal["merged", "per_commit"] | None = None
+
+    @field_validator("source_issues")
+    @classmethod
+    def validate_source_issues_length(cls, v: list[str] | None) -> list[str] | None:
```
lbarcziova, line 32: I thought we want to support by this also multiple issues to consolidate (>2), is that not the case?

lbarcziova (commented):
just few questions, otherwise LGTM

.github/workflows/build-and-push.yml
TomasKorbar: i did https://quay.io/repository/jotnar/ymir-api

ymir/api/jira_webhook.py
TomasKorbar: Right. Added

ymir/api/consolidation.py
TomasKorbar: Backend does not support that yet and i do not think that trying to mock that would be good solution. I will make submitting more issues possible with the consolidation rewrite.

lbarcziova (approved):
lgtm, thank you!

PR #808, opened 2026-09-08
Run Copr builds directly and use LLMs only for diagnosis

opohorel:
Call the existing build tool deterministically so successful builds no longer require model orchestration. Share the structured Copr result schema and invoke a diagnosis-only agent for failed builds with logs.

Migrate backport, inheritance, rebase, MR updates, and consolidation while preserving their timeout and retry policies. Keep rebuild and the incremental backport repair loop unchanged.

Add regression tests for gateway decoding, zero-LLM success handling, error classification, cancellation, and diagnosis without resubmission. Document the shared build flow and its dry-run behavior.

Validation: 64 focused container tests passed; Ruff lint and format checks passed. Broader container testing found two gateway startup test failures and the same missing-MCP-settings error in the install smoke test. The gateway test failures also reproduce on unchanged commit 00d2ba6. Other component suites passed.

Assisted-by: Codex

I've ran e2e test for build succeeding on first try and build failing on the first time with the need to fix it. No regressions were found. GPT-6 Astra estimates around 8% of token cost savings per backport task.

### DIFF
--- a/ymir/agents/build_agent.py
+++ b/ymir/agents/build_agent.py
@@ -1,32 +1,146 @@
+import asyncio
+import logging
 from typing import Any
+from urllib.parse import urlsplit
 
 from beeai_framework.agents.requirement.requirements.conditional import (
     ConditionalRequirement,
 )
 from beeai_framework.memory import UnconstrainedMemory
 from beeai_framework.middleware.trajectory import GlobalTrajectoryMiddleware
-from beeai_framework.tools import Tool
+from beeai_framework.tools import Tool, ToolError
 from beeai_framework.tools.think import ThinkTool
+from beeai_framework.utils.cancellation import AbortController
+from beeai_framework.workflows import Workflow
+from pydantic import BaseModel
 
 from ymir.agents.reasoning_agent import ReasoningAgent
 from ymir.agents.utils import (
+    get_agent_execution_config,
     get_chat_model,
     get_tool_call_checker_config,
     is_reasoning_enabled,
     render_template,
+    run_tool,
 )
 from ymir.common.logging_setup import get_trajectory_writeable
-from ymir.common.models import BuildInstructionsInput
+from ymir.common.models import (
+    BuildFailureAnalysisInput,
+    BuildFailureAnalysisOutput,
+    BuildInputSchema,
+    BuildInstructionsInput,
+    BuildOutputSchema,
+    BuildResult,
+)
 from ymir.tools.unprivileged.commands import RunShellCommandTool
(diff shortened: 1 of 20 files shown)


ymir/agents/build_agent.py
```diff
@@ -1,32 +1,108 @@
...
+        logger.warning("Copr build tool failed for %s: %s", build_input.jira_issue, e)
+        return BuildOutputSchema(success=False, error=str(e), is_infra_error=True)
+
+    # MCP normally returns a dictionary; also support text-only JSON results.
+    # Malformed responses must raise rather than being interpreted as success.
+    result = (
+        BuildResult.model_validate_json(raw_result)
+        if isinstance(raw_result, str)
+        else BuildResult.model_validate(raw_result)
+    )
```
nforro: Shouldn't this be rather handled in `run_tool()`?

ymir/agents/build_agent.py
```diff
@@ -1,32 +1,108 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def run_build(
+    *,
+    build_input: BuildInputSchema,
+    available_tools: list[Tool],
+    local_tool_options: dict[str, Any],
+) -> BuildOutputSchema:
```
nforro, line 53: I realize it's an overkill, but wouldn't it be better to use a `Workflow` here? For consistency and future-proofing.

nforro (commented):
Just curious, what was the motivation for this change?

opohorel:
> Just curious, what was the motivation for this change?

mainly to reduce token spend on a workflow, where LLM isn't that needed. thanks for the comments, I'll make the changes.

nforro (commented):
LGTM, a couple of LLM-identified issues:


1. **High** `ymir/agents/mr_consolidation_agent.py:786`: The fallback Copr project name uses the package verbatim. Valid RPM names such as `compat-libstdc++-33` contain `+`, which Copr rejects. Sanitize and length-limit the generated identifier.

2. **Medium** `ymir/tools/privileged/copr.py:26`: `ymir-tools` now imports the newly added `BuildResult`, but still declares `ymir-common>=0.1.0`. Upgrading tools with an older common package causes an import failure. Bump the common version and minimum dependency.

3. **Medium** `ymir/agents/build_agent.py:110`: `except Exception` catches BeeAI `AbortError`, which inherits both `FrameworkError` and `asyncio.CancelledError`. Framework-originated cancellation can therefore be treated as an ordinary diagnosis failure. Add `except asyncio.CancelledError: raise` first and test `AbortError`.

nforro (approved):

PR #812, opened 2026-09-09
sweep: expand list of allowed gitlab hosts

jpodivin:
List of allowed gitlab hosts misses some that we already allow on egress. This inconsistency impacts operations with gitlab API conducted by sweep and supervisor.

### DIFF
--- a/ymir/supervisor/gitlab_utils.py
+++ b/ymir/supervisor/gitlab_utils.py
@@ -17,7 +17,14 @@ GITLAB_URL = "https://gitlab.com"
 # be constrained to trusted instances -- otherwise an attacker-controlled host
 # (e.g. supplied through an LLM-authored ``blocker_reference``) would receive
 # the token.  This is the single source of truth for that allowlist.
-ALLOWED_GITLAB_HOSTS = frozenset({"gitlab.com", "gitlab.cee.redhat.com"})
+ALLOWED_GITLAB_HOSTS = frozenset(
+    {
+        "gitlab.com",
+        "gitlab.cee.redhat.com",
+        "gitlab.gnome.org",
+        "gitlab.freedesktop.org",
+    }
+)
 
 
 @cache


DaliborKr:
Looks good.

I just wonder. I haven't found `gitlab.cee.redhat.com` on the egress list. Is it covered by the line `cidrSelector: 172.0.0.0/8`? (This is not an issue since it obviously works, but I'm just curious.)

jpodivin:
> `gitlab.cee.redhat.com`

This is a good point. We are using it quite a lot, I'd say it is covered by the range. But maybe we could add it explicitly?

DaliborKr:
> But maybe we could add it explicitly?

I guess since it should be harmless, the explicit record in the list would make the usage of `gitlab.cee.redhat.com` clearer, if nothing else.

jpodivin:
> > But maybe we could add it explicitly?
> 
> I guess since it should be harmless, the explicit record in the list would make the usage of `gitlab.cee.redhat.com` clearer, if nothing else.

The exact URL used in tools, that's why it was included with original PR. If we are going to add it to egress too, it should be a separate PR.

lbarcziova (commented):
I am a bit confused, the commit/MR title says sweep, but this is adjusting the supervisor code?

jpodivin:
> I am a bit confused, the commit/MR title says sweep, but this is adjusting the supervisor code?

I know for sure that the issue affects sweep workflows, those that use this utility at least. But you could argue it's a more general thing.

edit: there ... commit message changed.

lbarcziova:
> I know for sure that the issue affects sweep workflows, those that use this utility at least. But you could argue it's a more general thing.

as far as I know the supervisor code shouldn't be used/deployed and is to be removed, are you sure about this? Or are the saying the sweep workflows use these?

jpodivin:
> > I know for sure that the issue affects sweep workflows, those that use this utility at least. But you could argue it's a more general thing.
> 
> as far as I know the supervisor code shouldn't be used/deployed and is to be removed, are you sure about this? Or are the saying the sweep workflows use these?

That is exactly what I'm saying. That's how it was built. https://github.com/packit/ai-workflows/blob/b4f80ba3f748e0926d26d57c2eeb5fc537babe1b/ymir/sweep/pr_pending.py#L29

DaliborKr (approved):

lbarcziova:
> That is exactly what I'm saying. That's how it was built. 

could we then rather extract this into some common constants? (I missed this during the original PR review) It would be good to not depend on the unmaintained supervisor code.

jpodivin:
> > That is exactly what I'm saying. That's how it was built.
> 
> could we then rather extract this into some common constants? (I missed this during the original PR review) It would be good to not depend on the unmaintained supervisor code.

That thought did cross my mind, yes. It should be a general "sweep", if you pardon me, not just a one off for supervisor. I believe that there could be a considerable amount of redundant code across the repo. Different implementations of the same functionality.

lbarcziova:
from a quick look, it seems sweep is the only deployed consumer of the supervisor. I agree it's outside of the scope of this PR, but could you please address this as a follow-up since it was missed during the review of #746?

PR #814, opened 2026-09-09
Add GitHub read-only token authentication to resolve rate limiting (PACKIT-5214)

majamassarini:
## Summary

Implements authenticated GitHub API access to increase rate limits from 60 to 5,000 requests/hour, resolving throttling issues that prevented agents from fetching upstream repository information.

**Issue**: https://redhat.atlassian.net/browse/PACKIT-5214  
**Trace**: RHEL-191367

## Problem

- Agents were hitting GitHub unauthenticated rate limit (60 req/hr per IP)
- Multiple agents share the same IP address  
- Critical failures when extracting upstream fix information from GitHub PRs

## Solution

- Add privileged MCP tools for authenticated GitHub:
  - PR metadata retrieval
  - Compare-range commit retrieval
  - PR, commit, `.patch`, `.diff`, and other GitHub-hosted patch-source retrieval
 - Route GitHub upstream metadata and patch requests through those gateway tools.
 - Use the GitHub REST API media types for recognized PR/commit patch and diff URLs.
 - Retain unauthenticated public-GitHub fallback only for standalone unprivileged gateway use when neither injected GitHub tools nor `MCP_GATEWAY_URL` is available.
 - Add validation for direct backport `UPSTREAM_PATCHES` input so empty comma-separated entries fail before repository operations begin

## Security

- `GITHUB_READONLY_TOKEN` is available only to the privileged MCP gateway and the pull-request-pending sweep CronJob.
- Agent pods with shell execution and LLM access do not receive the token.
- GitHub tools are read-only: they cannot push code, create pull requests, or modify GitHub resources.
- Use an expiring, fine-grained token restricted to needed repositories with read-only permissions. Do not use the classic `public_repo` scope because it permits writes.


nforro:
> **Token Requirements:**
> 
> * Scope: `public_repo` (read-only access to public repositories only)
> * Purpose: Rate limiting only - no write operations
> * Cannot push code, create PRs, or modify any GitHub resources

While I understand the motivation, and while this seems like an elegant solution, it still violates the rule that a LLM mustn't be allowed to access any secrets, and opens the possibility of human error - e.g. wrong scope selected when renewing the token. Yes, the risk can be minimized, but never completely mitigated.

lbarcziova:
@majamassarini which GH account will be used for access? Asking if it actually will have access to something private and what's the risk.

I also thought agents access GitHub stuff via RunShellCommandTool (curling API endpoints), could you double check it, if it should be covered too? Here, for the problem of not storing the secret in the pod, we could run a [GitHub MCP server](https://github.com/github/github-mcp-server) behind the privileged gateway instead (same pattern as logdetective-mcp), and instruct agent for other searching in github to use those tools instead of curl github API.

cgwalters:
This came up in a chat I had with @TomasTomecek what I think would make sense here is to reuse the Packit Github App. 

It's technically trickier in workflows that happen outside of GH - need a flow that is continually minting new tokens (due to the 1h expiry) and re-injecting them into active environments.

> We could run a [GitHub MCP server](https://github.com/github/github-mcp-server) behind the privileged gateway instead

Yes to combine these two threads, it looks like https://github.com/github/github-mcp-server/blob/main/docs/github-app-auth.md is documented.

cgwalters:
> and opens the possibility of human error - e.g. wrong scope selected when renewing the token.

Yes, though the GH MCP server also offers a blanket `--read-only` commandline argument https://github.com/github/github-mcp-server#read-only-mode so it'd be belt-and-suspenders on top of the token.

Though what would be quite important of course is to ensure the GH app private key is secured, because having that compromised would be quite severe.

majamassarini:
@nforro @lbarcziova @cgwalters sorry, I was working on it and didn't moved it to draft mode.

@nforro yes, I agree, and this is why I have reworked it in my last push
@lbarcziova I was thinking of creating a fine-grained token for read-only public-repository access. Ideally, we could create a new Ymir user for Github as we have it for Gitlab. The PR/compare lookup now goes through the gateway ( after my latest push). The generic shell tool can still run unauthenticated curl, but has no credential to disclose or reuse. I would use a Github MCP server if we will add more Github operations.
@cgwalters I agree the minting solution would be ideal. However, I think it will require much more effort, and I am not sure if we want to commit to that work to avoid the leakage of a read-only token for public repos. But If we want I can open a followup for it.

ymir/agents/backport_agent.py
```diff
@@ -715,7 +719,7 @@ async def prepare_normal_backport(state):
             for idx, upstream_patch in enumerate(state.upstream_patches):
                 patch_name = f"{state.jira_issue}-{idx}.patch"
                 content = await run_tool(
-                    "get_patch_from_url",
+                    _patch_fetch_tool_name(upstream_patch),
```
lbarcziova: iirc this tool is quite used in the triage agent too, should we change it there too?

lbarcziova (commented):
one note, otherwise this redesign makes sense to me, thanks!

openshift/README.md
```diff
@@ -15,6 +15,25 @@ Agents are deployed in the `jotnar-ymir--jotnar-ymir` project.
   GITLAB_TOKEN
   ```
 
+  `github-readonly-env`:
+  ```
+  GITHUB_READONLY_TOKEN
+  ```
+
+  **Important**: Use a fine-grained, expiring token restricted to the required
+  repositories with only read permissions. Do not use the classic `public_repo`
```
TomasTomecek: I'm confused by the "required repositories" part. We need to obtain info about *any* public Github repo, right? For our agents to do their job

majamassarini: yes this is a public repo only scope token. I'll adjust the docs 🫣

lbarcziova (approved):
LGTM!

PR #817, opened 2026-09-09
Add explicit instructions to not mess up RPM specs

antbob:
This patch adds explicit instructions to all spec-editing prompts: don't modify RPM macro definitions or conditional macro constructs.

We ran into issues with consolidation agent that just edited the spec like this for no apparent reason:

-%{?!apply_patch:%define apply_patch(qp:m:) {%__apply_patch %**}}
+%define apply_patch(qp:m:) {%__apply_patch %**}

This can theoretically happen with other spec modifying agents. While its not clear why that happens, perhaps it has tried to "simplify" things or just hallucinated an edit, its best to put some guardrails against it at least at the prompts level for now (it doesnt seem to happen often, could be a one off / rare condition) as atm i cant think of any legit reasons for agents to modify those things.


nforro (approved):

lbarcziova (approved):
🙏

PR #823, opened 2026-09-11
Comment on partial Jira closures

lbarcziova:
~~Change MR cleanup to close an open bot MR when any referenced Jira issue is already closed, even if other referenced issues remain open. Such an MR is no longer usable as-is, so the cleanup comment identifies the closed and remaining issues and explains how to reopen and retrigger Ymir if needed.~~
Based on the Slack discussion within team, changing this to comment only on these occurences; the MRs might be genuinely useful in some scenarios, requiring just small changes.

Reported in the forum chat.

Assisted-by: Codex (GPT-5)

TODO:

- [ ] agree with team whether this change of behaviour is ok
- [ ] check docs

RELEASE NOTES BEGIN

Our automatic cleanup now also comments on MRs that reference closed Jiras, to inform about the current state.

RELEASE NOTES END


nforro (dismissed):
LGTM, thanks. Should we perhaps run `make check-mr-cleanup-in-container` as a check on PRs?

majamassarini (dismissed):
LGTM

lbarcziova:
@nforro good point, addded, PTAL 🙏

nforro (approved):

PR #825, opened 2026-09-14
Fix rebase consolidated Jira metadata

majamassarini:
Ensure consolidated rebase MRs include every Jira issue in commit Resolves footers and the newest changelog entry.


ymir/agents/rebase_consolidation.py
```diff
@@ -23,6 +26,68 @@
 
 logger = logging.getLogger(__name__)
 
+_JIRA_REFERENCE_RE = re.compile(
+    r"^(?P<prefix>\s*(?:[-*]\s*)?)(?P<tag>Resolves|Related):\s*(?P<issues>.*)$",
+    re.IGNORECASE,
+)
```
nforro: We should make this consistent across workflows. There are at least two other places where these are parsed, but both recognizing only `Resolves` AFAICT. Is `Related` really relevant only here? What about other keywords that check_tickets [recognizes](https://gitlab.com/redhat/centos-stream/ci-cd/dist-git-gating-tests/-/blob/065379a57be1dab433b5c0843a89a142eb427eb4/bureaucrabot/__init__.py#L86-89)?

majamassarini: codex writes:

```
• The key distinction is between commit footers and changelog formatting.

  For the commit message, we should always write:

  Resolves: RHEL-100
  Resolves: RHEL-200

  That is the explicit requirement, and it is what Errata automation relies on.

  For the changelog, the log agent is instructed to follow the package’s existing convention. Some packages use:

  - Resolves: RHEL-100

  Others use:

  - Related: RHEL-100

  The new helper does not choose Related: itself. It only sees that the freshly generated changelog entry already uses Related: and appends missing tickets in that same format. Otherwise the changelog entry would
  mix styles:

  - Related: RHEL-100
  - Resolves: RHEL-200

  That is why supporting Related: is intentional.

  The reviewer is right that Ymir has several parsers with different keyword sets. But they serve different purposes:

  - Resolution-specific code should parse only Resolves:.
  - Code that discovers Jira references generally may parse both Resolves: and Related:.
  - This helper is a formatter/preserver, so it must recognize both formats that the log agent is allowed to generate.

  I would not add Fixes: or Reverts: to this helper: those express different relationships and should not cause us to add more Jira metadata as though they were resolution references.
```

nforro: Thanks, but the generated text is mostly nonsense - `Resolves` vs. `Related` have nothing to do with any convention, they mean different things, and can be (and AFAIK almost always are) combined in a single changelog. Also, how is `Fixes` a different relationship? `Reverts`, sure, that makes sense, we can most likely just ignore it.

My points were:
- We should avoid parsing the same thing in multiple places
- Is it correct that `Related` is ignored in MR consolidation and Y-Stream inheritance (probably yes, but we should make sure)?
- Should we recognize `Fixes` (and treat it the same way as `Resolves`)?

majamassarini: Codex says that `Fixes` is meant to reference CVEs and not Jiras (I don't know if this is true).
Also, now I understand what `Related` means here and why I shouldn't use it in this code.

I have extracted the regexp, without `Related`, in constants and reused it.

nforro: Wait, I see now where the confusion comes from: https://github.com/packit/ai-workflows/blob/1e1a9b34c82d2e0ca9cf2228fc7c00fd8f8ea47d/ymir/agents/prompts/log/instructions.j2#L16-L24
To me this seems completely wrong. There should always be `Resolves` in the new changelog entry, unless it's a fixup (but that's never the case with Ymir fixes). @TomasKorbar you authored that prompt, could you please explain the logic behind it to me?

TomasKorbar: Yep, this is related to the inconsistency i have been trying to solve in Bind package. The situation was as follows: Maintainer did not put Resolves into changelog, because that way he could cherry pick the change to Ystream commit and did not have to change the changelog, only the commit message. LLM was at the time confused by the absence and sometimes put the resolves there and sometimes omitted it.

nforro: Thanks, so the prompt should be fixed like e.g. this, right?
```diff
- include one in your entry using the same tag (Resolves or Related)
- and the same formatting (indentation, bullet prefix)
+ include one in your entry using the same formatting (indentation,
+ bullet prefix), but always use "Resolves"
```
Let me open a follow-up PR for that.

nforro (approved):
LGTM. I've opened https://github.com/packit/ai-workflows/pull/829 for the related log agent prompt issue.

PR #830, opened 2026-09-15
chore(sentry): adjust trace sampling

mfocko:
Lower down the sampling of transaction to save some quota. Especially since a lot of the caught events come from failed tool calls.

By default sample 50% of the transactions, also allow for customization via environment variable, if needed.

For more info see: https://docs.sentry.io/platforms/python/sampling/#sampling-transaction-events

Fixes PACKIT-5116


ymir/common/utils.py
```diff
@@ -459,6 +459,23 @@ def extract_text_from_adf(adf_body) -> str:
...
+
+    Args:
+        sampling_context: context data
+
+    Returns: traces sample rate (between 0 and 1)
+    """
+    if rate := os.getenv("SENTRY_TRACES_SAMPLE_RATE"):
+        return float(rate)
+    # TODO: Take sampling_context into account
+    return 0.5
```
mfocko, line 476: For Packit we do 25% for staging and 10% for production. I guess we can start with 50% here and adjust on the go.

majamassarini (approved):
Let's try, thanks.

nforro (approved):

PR #833, opened 2026-09-17
Fix PACKIT-5421: Check if package already built with fixed dependency

majamassarini:
## Summary

Fixes PACKIT-5421: Ymir was incorrectly recommending rebuild when a package was already built with the fixed dependency version.

## Changes

Added check_package_built_with_fixed_dependency() that verifies if a package's current build already used the dependency fix using two methods:

1. **root.log inspection (primary)**: Downloads build logs from Brew to see which exact dependency NVR was installed during the build. Most reliable since it shows what actually happened. Handles epochs, subpackages, and concurrent architecture probes.

2. **Timestamp comparison (fallback)**: When root.log unavailable (logs expire after some time), compares package completion time vs dependency fix time. If package built after fix, assumes it has the fix.

root.log is preferred because it's precise and handles edge cases (cached packages, epoch overrides). Timestamp fallback is needed because build logs are not preserved indefinitely.

## Performance

- Concurrent Koji lookups
- HEAD checks before downloads  
- 30s overall timeout across all architectures

## Security

- Validates package names match expected component
- Escapes JQL queries
- Validates NVR format

## Testing

Added 36+ unit tests covering all scenarios including gzipped logs, epochs, subpackages, dotted names, and fallback behavior.

Related: PACKIT-5289


opohorel:
From Codex:
```
  - [P1] Only the fastest architecture is checked. utils.py#L572-L583 cancels all remaining log requests after the first success. Different architectures can use different buildroots; one fixed result can incorrectly mark the whole build NOT_AFFECTED. All available architecture logs must agree, with
    missing/conflicting logs producing clarification.

  - [P1] Active builds are excluded. utils.py#L425-L430 requires Closed/Done plus a shipping resolution. Since Jira status transitions are disabled by default, a valid Koji build with Fixed in Build may remain New/In Progress and will be missed, causing the unnecessary rebuild this PR aims to
    prevent. Include active issues with valid builds while excluding rejected closed issues.

  - [P2] Metadata failure is treated as proof a rebuild is needed. utils.py#L782-L807 returns False for missing Koji builds, validation failures, and absent timestamps; the caller then declares the package definitely needs rebuilding. These cases should return an indeterminate result and request
    clarification.
```

majamassarini:
## Code Review Findings

Review focus: _"pay attention that we don't raise unnecessary errors, or we don't bother users with unnecessary requests. We ask users only when we are sure there is no data. If we can decide for a rebuild reading only one log is better. And for infra issues we don't bother users, we retry it later."_

### Bug 1 (High severity): `completion_time` comparison uses incompatible string formats

**File:** `ymir/common/utils.py` line 1066 (commit `1b3bb552`, not yet pushed)

```python
cutoff = datetime.now(UTC) - timedelta(days=90)
cutoff_str = cutoff.strftime("%Y-%m-%d %H:%M:%S")
if completion_time > cutoff_str:
    raise
```

`completion_time` from Koji is an `xmlrpc.client.DateTime` with format `"20260825T100000"`, while `cutoff_str` is `"2026-06-24 13:36:49"`. String comparison gives wrong results because at position 4, `"0"` (ASCII 48) is always greater than `"-"` (ASCII 45). **Any build from the same year always compares as "recent"**, defeating the 90-day cutoff entirely. Old builds with missing root.logs will always re-raise `TransientInfrastructureError` instead of falling back to timestamp comparison.

**Fix:** Convert both to `datetime` objects before comparing, or use Koji's `completion_ts` (Unix timestamp) instead.

---

### Bug 2 (High severity): Missing `TRIAGED_ALREADY_FIXED` from terminal label sets

The new `JiraLabels.TRIAGED_ALREADY_FIXED` label was added to `build_rebase_siblings_jql` and `build_rebuild_siblings_jql` exclusions, but is missing from **three other locations** that enumerate terminal labels:

1. **`ymir/agents/rebase_consolidation.py` line 286-302** -- `terminal_labels` set in sibling processing. Already-fixed siblings will be re-processed during rebase consolidation.

2. **`ymir/agents/rebase_consolidation.py` line 509-529** -- `terminal_labels` list for pending-sibling check. Already-fixed siblings will appear as "pending," blocking the primary issue.

3. **`ymir/tools/privileged/jira.py` line 964-981** -- JQL exclusion list for pending-triage Z-stream clones. Already-fixed Z-stream clones will appear as still pending triage, potentially blocking Y-stream postponement decisions.

---

### Bug 3 (Medium severity): `_fetch_root_log` returns `[]` for invalid NVR, but test expects `TransientInfrastructureError`

**File:** `ymir/common/utils.py` line 607-611

```python
if not nvr_match:
    logger.warning(f"Could not parse package NVR: {package_nvr}")
    return []
```

The test `test_fetch_root_log_invalid_nvr` (test_utils.py:1624) expects `TransientInfrastructureError` to be raised. The function's docstring also states it raises this exception. The code should raise instead of returning `[]` to match the documented contract.

---

### Bug 4 (Medium severity): Test mocks missing `build_id` for no-rootlog tests

**File:** `ymir/common/tests/unit/test_utils.py` lines 839 and 915

Both `test_check_package_built_with_fixed_dependency_no_rootlog_built_after_fix` and `test_check_package_built_with_fixed_dependency_no_rootlog_built_before_fix` define `candidate_build` without a `build_id` field. The production code at `utils.py:1022-1024` checks `build_id = package_build.get("build_id")` and raises `TransientInfrastructureError` if missing. Both tests will fail with `TransientInfrastructureError` before reaching the intended timestamp-comparison fallback.

**Fix:** Add `"build_id": 123456` to the candidate_build dicts in both tests.

---

### Bug 5 (Low-Medium): Noarch builds skip root.log verification entirely

**File:** `ymir/common/utils.py` lines 1039-1041

When a build produces only noarch RPMs, `built_archs` is set to `None` and root.log is never fetched. Noarch builds DO have root.logs stored under the actual build architecture (typically x86_64). This means noarch packages always take the weaker timestamp-comparison path, which asks users for clarification when root.log could give a definitive answer. Per the review goal of not bothering users unnecessarily, this could be improved by checking Koji task info for the actual build architecture.

---

### Cleanup 1: Dead `cve_id` variable in `already_fixed is None` branch

**File:** `ymir/agents/triage_agent.py` line 1152

```python
cve_id = getattr(data, "cve_id", None) or ""
```

This variable is assigned but never referenced by any of the four `reason` handlers below (lines 1154-1261). It appears to be left over from the `already_fixed is True` branch above, where `cve_id` IS used. Safe to remove.

---

### Efficiency 1: Sequential Koji fetches in `_compare_build_timestamps`

**File:** `ymir/common/utils.py` lines 875-876

Two independent `asyncio.to_thread(_get_koji_build, ...)` calls are awaited sequentially. The sibling function `_compare_dependency_evrs` (line 810) uses `asyncio.gather` for the same pattern. Using `gather` here would halve the latency for every timestamp-comparison fallback.

---

### Efficiency 2: Redundant `_get_koji_build` call for `package_nvr`

**File:** `ymir/common/utils.py` line 1017

`_select_highest_evr_build` already fetches build metadata for every candidate NVR (including the winning one) via `_get_koji_build`. The same NVR is fetched again at line 1017. Consider returning the build dict from `_select_highest_evr_build` to avoid this extra Koji round-trip.

ymir/common/utils.py
```diff
@@ -398,6 +407,835 @@ async def check_build_in_buildroot(
...
+            try:
+                root_logs = await _fetch_root_log(package_nvr, built_archs, require_all=not noarch_build)
+            except TransientInfrastructureError:
+                import time
+
+                completion_ts = package_build.get("completion_ts")
+                if completion_ts:
+                    cutoff_ts = time.time() - (90 * 24 * 60 * 60)
+                    if completion_ts > cutoff_ts:
+                        raise
```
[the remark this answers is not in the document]
majamassarini: If completion_ts is missing, the build metadata is already incomplete — and the timestamp fallback (_compare_build_timestamps) will return None for missing metadata anyway, which leads to a clarification request rather than a wrong rebuild decision. The existing flow is safe.

ymir/common/utils.py
```diff
@@ -398,6 +407,851 @@ async def check_build_in_buildroot(
...
+
+        # Handle case where active query failed
+        if active_candidates is None:
+            logger.warning(f"Active builds query failed for {package} in {fix_version}")
+            # If we have closed candidates, we can still check them
+            # But if we don't, this is a transient infrastructure failure
+            if not closed_candidates:
+                raise TransientInfrastructureError(
+                    f"Jira query failed for {package} in {fix_version} - "
+                    f"no closed builds found and active builds query failed"
```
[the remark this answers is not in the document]
majamassarini: Not worth it. If the closed build proves the package is already fixed (already_fixed=True), the active query failure is irrelevant — the work is done. If the closed build shows an older dependency, we already return False (needs rebuild), which is the correct conservative choice — rebuilding when an active build might exist is wasted work but not harmful, while waiting on a failed query would block progress unnecessarily. The current behavior is safe.

nforro:
I wonder if there isn't a more robust way to get dependencies used for a build, we could check e.g. how Koschei does it, perhaps there is some library for `root.log` processing.

docs/rebuild-decision-logic.md
```diff
@@ -0,0 +1,50 @@
+# Rebuild Decision Logic
+
+How Ymir determines whether a package needs to be rebuilt for a CVE dependency fix.
+
+## Decision Flow
+
+1. **Search Jira** for builds with "Fixed in Build" set (closed + active)
```
nforro: Is this necessary? Why not check root-log of the latest `-z-pending` build directly?

ymir/common/utils.py
```diff
@@ -398,6 +407,869 @@ async def check_build_in_buildroot(
...
+        package: Package name
+        fix_version: RHEL fix version (e.g., "rhel-10.2.z")
+        available_tools: List of available tools for Jira queries
+
+    Returns:
+        Tuple of (closed_builds, active_builds) where:
+        - closed_builds: List of (issue_key, nvr) tuples for closed builds
+        - active_builds: List of (issue_key, nvr) tuples for active builds, or None if query failed
+    """
+    from ymir.common.version_utils import get_fix_version_variants
```
jpodivin: Does it have to be imported here? Is there some odd circular dependency going on?

ymir/common/utils.py
```diff
@@ -398,6 +407,869 @@ async def check_build_in_buildroot(
...
+            logger.debug(f"Build {package_nvr} produced architectures: {list(built_archs)}")
+
+        # Step 3b: Fetch root.log from Brew
+        # For arch builds: require ALL arches. For noarch: accept any log found.
+        root_logs = []
+        if built_archs:
+            try:
+                root_logs = await _fetch_root_log(package_nvr, built_archs, require_all=not noarch_build)
+            except TransientInfrastructureError:
+                import time
```
jpodivin: Why not import this at the top?

ymir/common/utils.py
```diff
@@ -398,6 +407,869 @@ async def check_build_in_buildroot(
...
+        and epoch is the epoch number if present in root.log, else None
+    """
+    # Pattern: "Installing: [epoch:]<known_name>-<version>-<release>.<arch>"
+    # Sort by length descending to match longer names first (e.g., golang-bin before golang)
+    for pkg_name in sorted(known_names, key=len, reverse=True):
+        # Escape package name for regex (handles dots, etc.)
+        escaped_name = re.escape(pkg_name)
+        # Pattern: optional epoch, exact package name, then version-release.arch
+        # Use word boundary or hyphen after name to prevent matching subpackages as parent
+        pattern = rf"Installing:\s+(?:(\d+):)?{escaped_name}-([^\s]+)"
```
lbarcziova: I agree with @jpodivin that this approach might be fragile, did you consider utilising LLM in this step (parsing the root.log)?

docs/rebuild-decision-logic.md
majamassarini: We discussed it on DM, it is more straightforward to look up the jira issues directly, since we may need to instruct the user to update some of them.

ymir/common/utils.py
majamassarini: No, I moved it on the top now.

ymir/common/utils.py
majamassarini: Oh no really. This should be deterministic enough, I thought. But I can absolutely inspect and analyze the regexp more.

majamassarini:
> I wonder if there isn't a more robust way to get dependencies used for a build, we could check e.g. how Koschei does it, perhaps there is some library for `root.log` processing.
> 
> Ah, it should be possible to get Koji/Brew repo used for a specific build and then parse the dependency NVR from its metadata and avoid fetching and parsing `root.log` altogether:

We discussed that checking root.log is more reliable. Since it checks what was installed and not only available. However we are investigating if there are other solutions to understand what was installed without parsing root.log.

PR #834, opened 2026-09-17
Add protected context compaction for backports

opohorel:
Context compaction is opt-in and disabled by default. It is only enabled when BACKPORT_CONTEXT_MANAGEMENT=true is explicitly configured. I was trying to find optimal backport where the compaction would happen, but our agents are pretty good with token consumption already. If merged we can turn this feature on only when we are sure that it works as expected.


ymir/agents/reasoning_agent/context_management.py
```diff
@@ -188,17 +209,305 @@ def strip_manage_context_from_exchange(exchange: list[AnyMessage]) -> list[AnyMe
...
+            _message_snapshot(message)
+            for message in new_messages
+            if message.meta.get(YMIR_PROTECTED_META_KEY)
+        ]
+        if original_tasks != candidate_tasks:
+            raise ValueError("protected task changed")
+
+        candidate_memory = UnconstrainedMemory()
+        expected_candidate = [_message_snapshot(message) for message in new_messages]
+        await candidate_memory.add_many(new_messages)
```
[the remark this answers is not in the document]
opohorel: we always use UnconstrainedMemory

README-agents.md
```diff
@@ -81,6 +81,29 @@ repair prompt provides the file path for reading previous attempts and
 appending the outcome. The directory is removed with the issue's working
 directory on a fresh run.
 
+Backport context management is initially disabled. Set
+`BACKPORT_CONTEXT_MANAGEMENT=true` to let the model compact older investigation
```
lbarcziova, line 85: 👍 

any easy way to detect an issue suitable for testing this, when running in prod?

lbarcziova (approved):
LGTM

README-agents.md
opohorel: pretty much anything that is not a simple backport. sadly as the LLMs got better, it is harder to get them to grow the context over 50k or so. they don't get confused like they used to

lbarcziova: and could we detect this even programatically to flag this?

opohorel: I'm not sure. I'll probably just look through the trace server occasionally and try to see which BackportWorkflow had big cost and run it locally. also the next thing would be to find components which have rules, to see if they were kept in place after compaction

lbarcziova: sounds good

PR #835, opened 2026-09-17
Add trace links to Jira failure comments

lbarcziova:
Append a configurable trace-server issue link to error comments while preserving the existing Jira footer format. Fall back when TRACE_VIEWER_URL is empty and cover both error and non-error comments.

Assisted-by: Codex (GPT-5)


ymir/agents/constants.py
```diff
@@ -1,12 +1,17 @@
...
 
 from ymir.common.constants import YMIR_COMMENT_MARKER
 
 BRANCH_PREFIX = "automated-package-update"
 ZSTREAM_TARGET_LABEL = "target::zstream"
 
 JIRA_BROWSE_URL = "https://redhat.atlassian.net/browse/{issue}"
+DEFAULT_TRACE_VIEWER_URL = (
+    "https://trace-server-jotnar-ymir--jotnar-ymir.apps.gpc.ocp-hub.prod.psi.redhat.com"
+)
```
nforro: Do we really want to have this hardcoded here? I think omitting the appendix in case `TRACE_VIEWER_URL` is missing is fine.

nforro (dismissed):
LGTM, thanks!

ymir/agents/constants.py
lbarcziova: good point, addressed

opohorel (approved):

